Issue #39, opened 2026-07-24
Benchmark open-weight detectors and pointing models (OWLv2, Grounding DINO, AI2 Molmo) against RampNet


jonfroehlich:
## Motivation

The model-comparison harness (#34, #36) has now scored two proprietary VLMs and one open one against RampNet on the richmond split. Every general-purpose VLM has lost badly, and the newest entrant lost worst of all:

| model | P | R | F1 | tp/fp/fn |
|---|---|---|---|---|
| **rampnet** | **0.964** | **0.768** | **0.855** | 238/9/72 |
| gemini-3.1-pro-preview | 0.631 | 0.700 | 0.664 | 217/127/93 |
| gemini-3.6-flash | 0.626 | 0.642 | 0.634 | 199/119/111 |
| Qwen/Qwen3-VL-8B-Instruct | 0.323 | 0.452 | 0.377 | 140/**293**/170 |

(richmond, 124 reviewed panos, perspective tiling, match radius 0.022. Bend and Qwen3-VL-32B are running.)

The pattern worth noticing: **these are all chat models doing localization as a side skill.** They are FP-heavy — 119–293 false positives against RampNet's 9 — which is what you get when a conversational model is asked to enumerate small objects. Before concluding "general models can't do this," we should test the model classes that are actually *built* for it. All of the following are open weights, i.e. free to run on Hyak.

### 1. Open-vocabulary detectors: OWLv2 + Grounding DINO

- `google/owlv2-large-patch14-ensemble`
- `IDEA-Research/grounding-dino-base`

Text-prompted *detectors*, not chat models: give them "curb ramp", get boxes back. These are the standard strong open baseline for exactly this task, and they're small — minutes to run, no multi-GPU sharding.

**They also unlock a capability the harness currently lacks.** `docs/model_comparison.md` states "No AP for VLMs" because Gemini/Qwen boxes carry no calibrated score, so we only ever report a single operating point. OWLv2 and Grounding DINO emit **real per-box confidences**, which would give us PR curves, AP, and a threshold sweep for a non-RampNet model for the first time. That matters directly for the recall-first direction: a detector that can be *tuned* toward recall is worth far more than a chat VLM pinned at one operating point.

### 2. AI2 Molmo — points, not boxes

- `allenai/MolmoPoint-8B` — a dedicated pointing model
- `allenai/Molmo2-8B` (Qwen3-based) and `allenai/Molmo2-O-7B` (Olmo-based, fully open incl. data)

[...]

Issue #20, opened 2026-07-21
Set up a harness for experimenting with newer detection models/backbones


jonfroehlich:
## Motivation

RampNet's paper state is preserved at `v1.0-iccv2025`, and the deployment side is now exercising the model on imagery beyond GSV — notably Mapillary 360 panos (Richmond, VA), where image quality is generally lower (consumer rigs, motion blur, re-compression). Newer backbones and foundation models have landed since the paper was trained, and it would be good to know systematically whether any of them improve curb-ramp detection — especially on lower-quality imagery.

## Proposal

Set up a repeatable experiment path so trying a newer model is a config change plus a training/eval run, not a fork:

- **Benchmark of record**: the corrected gold-set evaluation (#9) as the fixed comparison target; report P/R at the operating threshold plus AP, alongside the committed `v1.0-iccv2025` numbers.
- **Model swapping**: make the backbone/architecture configurable in `stage_two/train.py` (builds on warm-start support, #5), keeping the keypoint-head + heatmap contract stable so the evaluators and export path keep working.
- **Candidates to seed the queue**: newer ViT/DINO-family backbones with the existing head, and (separately scoped) a look at whether open-vocabulary or VLM-based detectors are competitive zero/few-shot.
- **Results tracking**: one results table (repo doc or issue thread) with env spec per run (#6), so experiments accumulate instead of scattering.
- **Cross-domain eval**: include a lower-quality-imagery slice (e.g. a validated Mapillary sample from the Richmond deployment) so improvements are measured where deployments actually hurt.

Happy to discuss scope — filing so the idea has a home while the Mapillary deployment work is fresh.

## Follow-ups from the model-comparison harness (PR #34)

The VLM-comparison harness (RampNet vs Gemini/Qwen) landed via PR #34; these are the deferred increments it leaves for this issue:

[...]

Issue #35, opened 2026-07-23
Experiment: hybrid RampNet + open VLM (Qwen3-VL) — ensemble / cascade / distillation


jonfroehlich:
## Motivation

RampNet is strong but not perfect — on the deployment benchmark it sits around P ~0.95 / R ~0.80, so ~1 in 5 real curb ramps is still missed. General VLMs now do bounding-box detection, and the model-comparison harness (#20, PR #34) lets us score them on the same ground truth. The question this issue tracks: **can a *hybrid* of RampNet + an open VLM beat RampNet alone — at a cost an open-source, academically-hosted project can actually afford?**

Project Sidewalk is open source, runs on academic servers, and has very little money, so "just call a paid VLM per pano" is off the table at deployment scale. That constraint isn't a blocker — it *shapes* the design.

## The cost reframe: self-host the VLM

- **Gemini (paid API)** → useful as a *reference baseline*, but never a deployment component (per-call cost × millions of panos).
- **Qwen3-VL (open weights)** → self-hostable on **makelab2 (A40, 48 GB, 24 TB)** or **Hyak**; inference cost is GPU-hours, not dollars. This makes Qwen — not Gemini — the real hybrid partner.
- **Tiling wrinkle worth testing:** the harness feeds the VLM 1024×1024 perspective tiles, not the full 4096+ pano. The "A40 OOMs Qwen at native res" concern (why Qwen was earmarked for Hyak) may not apply to *tiled* inference — so makelab2 could be self-sufficient here. Verify empirically.

## Approaches (with the standard names)

| Approach | What it is | Fit for us |
|---|---|---|
| **Ensemble / late (decision-level) fusion** — Weighted Boxes Fusion (Solovyev et al. 2019), NMS-fusion | Run both, pool + merge detections weighted by confidence | Best accuracy; pays VLM cost on *every* pano forever |
| **Confidence-gated cascade** (propose→verify) | RampNet everywhere; VLM invoked only on RampNet's uncertain/empty panos to recover misses or prune FPs | Bounds VLM calls to a fraction — budget-friendly |
| **Offline distillation / pseudo-labeling** (teacher→student, weak supervision) | VLM improves the *auto-generated training set* once; deploy RampNet alone | **Free at inference**; most on-brand for an auto-labeling project; most publishable |

## Decision gate — do this first (cheap, no training)

[...]

PR #40, opened 2026-07-24
Benchmark open detectors and pointing models: OWLv2, Grounding DINO, Molmo (#39)

jonfroehlich:
Closes part of #39 — the wiring **and the runs**: OWLv2, Grounding DINO, and Molmo2-8B are scored on both cities (Hyak, L40S), and the results tables live in `docs/model_comparison.md`. Still open from #39: `MolmoPoint-8B` (wired, never met real weights), `Molmo2-O-7B`, and the 32B-FP8 nice-to-have.

Every model scored before this PR was a chat VLM doing localization as a side skill, and they lose the same way: 119–293 false positives against RampNet's 9. This adds the two model classes built to *detect* (open-vocab) or *point* (Molmo) rather than chat, so "the weakness is specifically a *chat* problem" gets tested rather than assumed. **The runs say it isn't:** the open-vocab detectors do far *worse* than the chat VLMs (OWLv2's best sweep F1 0.184, Grounding DINO 0.073, vs Gemini-3.6-flash 0.634), while Molmo2-8B lands as the best open-weight challenger (F1 0.457 richmond / 0.449 bend) — balanced P≈R, no FP flood — yet still ~0.4 F1 behind RampNet. OWLv2's consolation prize is a 0.971 recall ceiling that recovers 69 of RampNet's 72 richmond misses, but at ~128 FP per recovered ramp it is a recall oracle, not a usable candidate generator. (These stay general, off-the-shelf models given a short untuned query — not detectors trained for curb ramps; the only purpose-built model here is RampNet. See "Scope of the claim" in the doc.)

## Open-vocabulary detectors — the substantive part

`OwlV2Detector` (`google/owlv2-large-patch14-ensemble`) and `GroundingDinoDetector` (`IDEA-Research/grounding-dino-base`) emit a **calibrated per-box score**, and that score is now carried end to end — `pixel_boxes_to_points` → `dedup_points` (keeps the higher-scoring copy of a cross-view duplicate) → `score_pano`'s confidence-ordered greedy match. That gives the harness, for the first time on a non-RampNet model:

- **AP** — new table column; `aggregate()` computes it via `rampnet.metrics` when every prediction carries a score. Chat VLMs read `-`.
- **PR curves** — `--pr-out DIR` writes JSON per model plus a combined PNG.
- **A threshold sweep** (`--sweep`) and a re-scored operating point (`--op-threshold`).

Both are free: detections cache down to a score **floor** (`--score-threshold`, default 0.05, part of the cache key), and every higher operating point re-scores that cache with no second model run. This is what the recall-first direction needs — a detector you can tune beats a chat model pinned at one point.

## Molmo — points, not boxes

`MolmoDetector` removes the box→center reduction every other challenger needs. `MolmoPoint`'s special-token path (`extract_image_points` + constrained decoding) is wired separately, selected by model id.

## Three findings

1. **Molmo's coordinate scale is not uniformly 0–100**, as the issue assumed. Molmo 1 emits percentages in `<point x= y=>`; **Molmo 2 emits 0–1000** triplets in `<points coords="…"/>`, per its own model card. The two differ *syntactically*, so the scale is inferred per tag (unlike Qwen's case, where both conventions looked identical) — `--molmo-coord-scale` overrides. Out-of-frame points are dropped, so a wrong scale fails loudly ("detected nothing") instead of biasing quietly.
2. **OWLv2's padded-square box frame is already handled by transformers.** Its processor pads to `max(h, w)` before resize, but current `_scale_boxes` scales by `max(h, w)` internally — verified on a 2:1 crop, where both target-size conventions put the top box at y 0.815 against a true 0.817. `owlv2_target_size` keeps the square anyway: it is also correct under the older per-axis scaling, and it states the frame the caller normalizes against rather than relying on a library internal.
3. **RampNet's AP is truncated and not comparable to the detectors'.** Its bundle detections were extracted at a 0.5 peak threshold, so its curve has no low-confidence tail (every sweep row below 0.5 is identical) and AP 0.763 richmond / 0.754 bend is a **lower bound**. Documented as a caveat: compare operating points against RampNet, or re-extract at a lower peak threshold.

## Also

- `dump_detections.py` draws all three prediction shapes: boxes, **scored** boxes (score printed — it's the number the sweep tunes), and points.
- The curb-ramp definition is factored into `CURB_RAMP_DEFINITION`, shared by the box and point prompts. `DETECTION_PROMPT` is **byte-identical**, so `test_gemini_cache_key_is_frozen` still pins `b4401afc…` and the thousands of already-paid Gemini detections still hit the cache.
- `compare.py` now skips a model that fails to load for *any* reason (printing the exception type) rather than only `ImportError`/`RuntimeError` — a checkpoint whose remote code won't load must not cost a multi-model cluster run the models that already ran.
- `run_open_models.slurm`: Hyak launcher for the new leg.

## Verification status

- **OWLv2 / Grounding DINO:** 2-pano desktop wiring probe first (scores carried through, boxes on ramps, cache reuse and model-load skip confirmed), then the full Hyak runs. Desktop (RTX 3070) and cluster (L40S) produced *identical* numbers on the probe, so a desktop rehearsal is faithful.
- **Molmo2-8B: verified, then run.** The overlay-first discipline paid for itself — the first real run put every point on the left edge, because Molmo 2's `coords` list opens with an **image index** the parser mistook for a point id. Fixed (a verbatim-output regression test pins it), overlay re-confirmed crosshairs on ramps, then both cities run and scored. It needs a **dedicated env** at the model cards' `transformers==4.57.1` pin — the pin is real, not cautionary; 5.x fails at import inside Molmo's own Hub code (runbook has the recipe).
- **MolmoPoint-8B remains wired but unverified** — it has never met real weights.

RampNet's own rows reproduce exactly: 0.964/0.768 richmond, 0.961/0.761 bend.

## Review follow-ups

A post-run review pass (`f1e437d`) tightened three things: the `--sweep` table drops thresholds below the detector's cache floor (rows down there silently repeated the floor row while reading as real measurements); the Molmo triplet parser decides its alignment by the point-id column (1, 2, 3, …) rather than token count alone, so a generation truncated mid-triplet can no longer misalign every pair into in-frame garbage near x≈0; and the scorer's `prediction_confidence` helper is shared with the CLI instead of duplicated. Deliberately *not* done: pinning HF revisions into detector signatures — it would invalidate every already-paid cached detection, so it needs its own design pass (follow-up issue material).

## Tests

37 new tests; full suite **137 pass** (rebased onto #33's matcher refactor, numbers unchanged).

The first commit is unrelated pending work from the in-flight Qwen run (`sbatch -A`, `PYTHON` override, one-node GPU request), committed separately so it isn't tangled in this.

Refs #20, #36.

### DIFF
--- a/docs/model_comparison.md
+++ b/docs/model_comparison.md
@@ -86,6 +86,15 @@ radius of a real GT point, which the per-detection human verdict scored differen
   same set. Note also that `--op-threshold` truncates the curve it is computed from: the AP
   printed alongside a thresholded row is the AP *of that row's operating range*, so quote
   full-range AP from a run without `--op-threshold`.
+- **AP is not comparable across models at different floors — and RampNet's is truncated.**
+  Every model's curve stops where its detections stop. RampNet's bundle detections were
+  extracted at a **0.5** peak threshold, so its curve has no low-confidence tail at all
+  (visible in `--sweep`: every row below 0.5 is identical) and its AP — 0.763 richmond /
+  0.754 bend — is a **lower bound**, close to its recall ceiling of 0.768 / 0.761 times a
+  near-1.0 precision envelope. The open detectors are cached down to 0.05, so their curves
+  extend into a region RampNet's simply doesn't cover. Compare AP between OWLv2 and
+  Grounding DINO freely; against RampNet, compare operating points, or re-extract RampNet's
+  detections at a lower peak threshold first.
 - **A swept threshold is tuned on the test set.** The `--sweep` table's best-F1 row is
   chosen on the benchmark itself. There is no separate val split, so quote it as an
   optimistic upper bound on what threshold tuning buys, not as a held-out result.
--- a/docs/model_comparison.md
+++ b/docs/model_comparison.md
@@ -15,8 +15,75 @@ Three classes of challenger, which fail differently and are worth keeping distin
 
 The chat VLMs are all doing localization as a side skill, and they lose the same way: they
 are false-positive-heavy (119–293 FP against RampNet's 9). The other two classes exist in
-this harness to test whether that is a property of *general models* or of *chat models* —
-see "What each model class buys you" below.
+this harness to test whether that is a property of *general models* or of *chat models*.
+
+**It is not.** See the results below — the purpose-built open-vocabulary detectors do far
+*worse* than the chat VLMs on this task, not better.
+
+## Results
+
+Perspective tiling, match radius 0.022, all models scored against the same derived GT.
+Open detectors are shown at their 0.05 cache floor; their tuned operating points are in the
+sweep below. Run on Hyak (L40S); RampNet and Gemini rows are cache-scored.
+
+**richmond** (124 reviewed panos, 310 GT ramps)
+
+| model | P | R | F1 | AP | tp/fp/fn |
+|---|---|---|---|---|---|
+| **rampnet** | **0.964** | 0.768 | **0.855** | 0.763 | 238/9/72 |
+| gemini-3.1-pro-preview | 0.631 | 0.700 | 0.664 | – | 217/127/93 |
+| gemini-3.6-flash | 0.626 | 0.642 | 0.634 | – | 199/119/111 |
+| Qwen3-VL-32B-Instruct | 0.760 | 0.297 | 0.427 | – | 92/29/218 |
+| Qwen3-VL-8B-Instruct | 0.323 | 0.452 | 0.377 | – | 140/293/170 |
+| owlv2-large-patch14-ensemble | 0.033 | **0.971** | 0.064 | 0.104 | 301/8799/9 |
+| grounding-dino-base | 0.028 | 0.852 | 0.053 | 0.032 | 264/9321/46 |
+
+**bend** (110 reviewed panos, 327 GT ramps)
+
+| model | P | R | F1 | AP | tp/fp/fn |
+|---|---|---|---|---|---|
+| **rampnet** | **0.961** | 0.761 | **0.850** | 0.754 | 249/10/78 |
+| gemini-3.1-pro-preview | 0.706 | 0.581 | 0.638 | – | 190/79/137 |
+| gemini-3.6-flash | 0.608 | 0.587 | 0.597 | – | 192/124/135 |
+| Qwen3-VL-32B-Instruct | 0.706 | 0.294 | 0.415 | – | 96/40/231 |
+| Qwen3-VL-8B-Instruct | 0.379 | 0.336 | 0.357 | – | 110/180/217 |
+| owlv2-large-patch14-ensemble | 0.037 | 0.951 | 0.070 | 0.093 | 311/8187/16 |
+| grounding-dino-base | 0.038 | 0.850 | 0.073 | 0.049 | 278/6969/49 |
+
+### What the numbers say
+
+1. **RampNet still wins by a wide margin**, and nothing tested comes close on F1.
+2. **Purpose-built detectors did worse than chat models, not better.** OWLv2's best F1 over
+   the whole threshold sweep is **0.184** (thr 0.25: P 0.130 / R 0.310); Grounding DINO's is
+   **0.073**. Both are far below Gemini-3.6-flash's 0.634. The issue-#39 hypothesis — that
+   the chat VLMs' weakness was a *chat* problem — is refuted. Open-vocabulary detection with
+   a text query is simply not selective enough for an object that looks like a slightly
+   different patch of concrete.
+3. **Capacity isn't the chat VLMs' problem either.** Qwen-32B moved to the *precise* end
+   (P 0.760 / R 0.297) versus 8B's FP flood (P 0.323 / R 0.452) — the operating point moved,
+   F1 barely did (0.427 vs 0.377).
+4. **But OWLv2 has an extraordinary recall ceiling.** At its floor it finds **97.1%** of
+   richmond's ramps, against RampNet's 76.8%.
+
+### The recall-first angle: OWLv2 as a candidate generator
+
+Recall matters more than precision here (a false negative is a permanently missing ramp; a
+false positive is a cheap review). So the question is not OWLv2's F1 but whether it *sees*
+what RampNet misses. It does — nearly all of it:
+
+| OWLv2 thr | RampNet FN | recovered by OWLv2 | union recall | OWLv2 FP | FP per recovered ramp |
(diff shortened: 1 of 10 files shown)


jonfroehlich:
## Results are in — the hypothesis is refuted

Ran on Hyak (L40S, `gpu-l40s-makelab`). Full tables are now in `docs/model_comparison.md`.

**richmond** (124 panos, 310 GT ramps) — open detectors shown at their 0.05 cache floor:

| model | P | R | F1 | AP | tp/fp/fn |
|---|---|---|---|---|---|
| **rampnet** | **0.964** | 0.768 | **0.855** | 0.763 | 238/9/72 |
| gemini-3.1-pro-preview | 0.631 | 0.700 | 0.664 | – | 217/127/93 |
| gemini-3.6-flash | 0.626 | 0.642 | 0.634 | – | 199/119/111 |
| Qwen3-VL-32B | 0.760 | 0.297 | 0.427 | – | 92/29/218 |
| Qwen3-VL-8B | 0.323 | 0.452 | 0.377 | – | 140/293/170 |
| owlv2-large | 0.033 | **0.971** | 0.064 | 0.104 | 301/8799/9 |
| grounding-dino-base | 0.028 | 0.852 | 0.053 | 0.032 | 264/9321/46 |

bend is the same story (OWLv2 R 0.951 / P 0.037; GDINO R 0.850 / P 0.038).

**The premise of this issue does not hold.** The purpose-built open-vocabulary detectors do
*worse* than the chat VLMs, not better. OWLv2's best F1 across the whole sweep is **0.184**
(thr 0.25: P 0.130 / R 0.310) and Grounding DINO's is **0.073**, against Gemini-3.6-flash's
0.634. Text-prompted open-vocabulary detection isn't selective enough for an object that
looks like a slightly different patch of concrete.

Capacity isn't the chat VLMs' problem either: Qwen-32B moved to the precise end
(P 0.760 / R 0.297) versus 8B's FP flood — the operating point moved, F1 barely did.

**What is genuinely interesting: OWLv2 as a recall oracle.** It finds 97.1% of richmond's
ramps, and recovers **69 of RampNet's 72 misses** — a union recall of **0.990**:

| OWLv2 thr | recovered of 72 | union recall | OWLv2 FP | FP per recovered ramp |
|---|---|---|---|---|
| 0.05 | **69** | **0.990** | 8799 | 128 |
| 0.15 | 46 | 0.916 | 2923 | 64 |
| 0.25 | 18 | 0.826 | 640 | 36 |

But at 36–128 FP per recovered ramp versus **~6** for Gemini-3.6-flash (issue #35), it's a
6–20× less efficient complement. Recall oracle, not a usable candidate generator.

## Molmo: a real bug, caught by the overlay

The `dump_detections` gate earned its keep. First real Molmo2-8B run put **every point on the
left edge** (x≈0) while the GT sat mid-frame. Raw output:

```
<points coords="1 1 308 305 2 752 377">curb ramp</points>
```

That leading `1` is an **image index**, not a point id — 7 tokens for 2 points, 13 for 4. The
parser consumed it as an id and shifted every coordinate one slot left. Fixed in 61c52d0 by
chunking tokens after dropping the index, rather than using the model card's own
`([0-9]+) ([0-9]{3,4}) ([0-9]{3,4})` regex — that pattern only resynchronizes past the index
*by accident* (it demands 3–4 digits), and the same constraint silently drops any point in
the leftmost/topmost 10% of a view. Regression tests use the verbatim model output.

Re-verified: points now land on the ramps. Full Molmo runs are queued.

Also confirmed the `transformers==4.57.1` pin is real, not cautionary — on 5.14.1 Molmo's Hub
code raises `TypeError: Unexpected keyword argument image_use_col_tokens`. It needs `einops`
**and `requests`**. It gets its own env at the pin rather than a downgrade of the env the
other models use; the harness imports fine on both.

## Process note

Desktop and cluster produced **identical** numbers on the 2-pano smoke (OWLv2 18/156/1/5,
AP 0.356; GDINO 18/160/1/3, AP 0.247) and the same 94 OWLv2 boxes across the same six views —
so a local probe is a faithful rehearsal of the cluster job.

One thing worth fixing separately: the detection cache key covers the model, rig and prompt
but **not the parser version**, so a parser bug can persist in cached detections after the
code is fixed. The Molmo bug was caught before any full run cached anything, but that was
luck rather than design.

jonfroehlich:
## Molmo results are in — both cities done, zero failures

The coord-parser fix held. Full unified table (Hyak L40S, perspective tiling, radius 0.022):

**richmond** (124 panos, 310 GT ramps)

| model | P | R | F1 | AP |
|---|---|---|---|---|
| **rampnet** | **0.964** | 0.768 | **0.855** | 0.763 |
| gemini-3.1-pro-preview | 0.631 | 0.700 | 0.664 | – |
| gemini-3.6-flash | 0.626 | 0.642 | 0.634 | – |
| **molmo2-8B** (points) | 0.410 | 0.516 | **0.457** | – |
| Qwen3-VL-32B | 0.760 | 0.297 | 0.427 | – |
| Qwen3-VL-8B | 0.323 | 0.452 | 0.377 | – |
| owlv2-large | 0.033 | 0.971 | 0.064 | 0.104 |
| grounding-dino-base | 0.028 | 0.852 | 0.053 | 0.032 |

**bend:** Molmo P 0.510 / R 0.401 / F1 0.449 — same relative placement.

**Molmo is the best open-weight model** and the only challenger with a *balanced* profile
(P≈R) instead of an FP flood or extreme caution — and it does it natively in points, no
box→center reduction. But it still trails RampNet by ~0.40 F1 and both Geminis, and it's
**sparse**: on the verified overlay pano it proposed 2 points where 4 ramps were visible,
which is where its 150/196 false negatives come from.

### Final verdict on #39

All three proposed candidates are now benchmarked. None beats RampNet, and the issue's core
hypothesis — that purpose-built models would beat the chat VLMs — is **refuted**: the
open-vocabulary detectors are the *worst* models tested, and the pointing model lands between
the two Geminis and the Qwens. Ranked by F1: RampNet ≫ Gemini-pro > Gemini-flash > **Molmo** >
Qwen-32B > Qwen-8B ≫ OWLv2 > Grounding DINO.

The lasting deliverables regardless of the verdict: AP / PR curves / threshold sweep for any
scored detector, a points-native path, the OWLv2 recall-oracle finding (union recall 0.990),
and the measurement that hood/nadir pixels actively hurt AP.

This branch is ready for review — I don't think there's more to run here unless you want the
OWLv2 query sweep (item 4 in Next increments) or MolmoPoint-8B's special-token path as a
second data point.

Issue #54, opened 2026-07-24
Lower the deployment operating point (peak threshold 0.55 → ~0.25–0.35): a free, recall-first lever


jonfroehlich:
## Motivation

RampNet's deployment operating point — peak extraction at `threshold_abs = 0.55`, `min_distance = 10` — has never been characterized below 0.55. The recall error analysis (#25) took a first look and found a **free recall lever**: dropping the threshold buys ~+7–10 recall points for one changed constant (no retraining), and **0.55 is not even F1-optimal** (F1 peaks nearer 0.25–0.35). Under recall-first (a false negative is a permanent loss from the inventory; a false positive is a cheap review), that's the trade we want. But #25 only sampled a handful of discrete thresholds — this issue does it the standard way and corrects a ground-truth bias that a raw curve hides.

## What to produce

**1. A proper continuous precision–recall curve + AP** — the standard artifact, not a few hand-picked points.

Do it with a **single low-floor extraction**, not per-threshold re-inference: run RampNet once per pano, extract *all* peaks at a low floor (`threshold_abs ≈ 0.05`, keep `min_distance = 10`), and carry **each peak's height as its confidence score**. Then sweep the score threshold post-hoc in fine steps (0.05) to get:

- the full PR curve (precision & recall vs. threshold),
- **Average Precision** integrated over the whole curve,
- F1-vs-threshold, with the deployed **0.55** point marked.

This also fixes an AP problem: RampNet's *committed* benchmark detections were extracted at a 0.5 peak threshold, so the AP we can currently quote is **truncated** (integrates only the high-confidence tail) and isn't comparable to the open detectors cached at a 0.05 floor. The low-floor extraction yields RampNet's **first honest, comparable AP**.

`min_distance` stays 10 (#25 found 10→3 is a dud: ~0.5 pt, costs precision).

**2. A ground-truth-completeness correction on the low-confidence end.** A PR curve is only as trustworthy as the completeness of its GT. Our benchmark GT was labeled/reviewed largely in the ~0.55 detection regime (plus a missed-ramp pass), so when the threshold drops we surface faint detections in a confidence band the GT never fully audited — and a **real ramp nobody labeled scores as a false positive**. That means the **left half of the PR curve systematically understates precision** by an unknown amount (#25's own caveat: "low-threshold precision deserves a fresh spot-check before being quoted").

[...]

PR #69, opened 2026-07-28
Annapolis GT review: P 0.964 / R 0.728, and misses are a distance problem

jonfroehlich:
Ground truth for the Annapolis, MD split (#68), and the finding that came out of it.

## Scores

| | Precision | Recall |
|---|---|---|
| All 125 | **0.964** (CI 0.931–0.982) | **0.728** (CI 0.674–0.776) |
| Unbiased (120) | 0.961 (CI 0.922–0.981) | 0.692 (CI 0.632–0.746) |

125/125 panos fully judged, 227 detections reviewed, 80 confident missed marks
(+34 unsure). With `--lenient-duplicates`, precision is 0.977 / 0.972.

## The density question is settled

The split existed to decide whether Annapolis's 27.96% detection rate — against
Morgantown's 11.2% — was real infrastructure or a false-positive artifact. **It is
real**: only 8 of 222 judged detections were wrong. A compact colonial grid plus the
Naval Academy genuinely carries more curb ramps per pano.

## Precision fits the camera curve; the camera is a new kind

Annapolis is the first split shot on a **survey-grade vehicle rig** rather than a
consumer action cam — a Trimble MX7, uniformly 8000×4000, one contributor across 101
sequences. Precision 0.964 lands between richmond (0.960) and morgantown (0.975),
consistent with the existing story, though the CI is wide enough not to be strong
evidence alone.

It is the most legible imagery in the benchmark: detection abstention was **2.2%
(5/227)**, the lowest of any split, beating morgantown's 4.3%.

## The finding: misses are a distance problem

| | n | median est. distance |
|---|---|---|
| True detections | 214 | **11.5 m** |
| Missed (confident) | 80 | **20.4 m** |
| Missed (unsure) | 34 | 24.3 m |

**80% of confidently-missed ramps lie beyond the median distance of a successful
detection**; 97% of the unsure ones do. The ratio is independent of the assumed 2.5 m
camera height — changing it rescales all distances equally, so only the metre labels
move.

Restricting to ramps within ~12.5 m, **recall rises from 0.733 to 0.870 at essentially
unchanged precision (0.958)**.

That reframes the ~0.70 recall every city split reports as substantially a *viewpoint*
limitation rather than a detection-quality one: a ramp 20–40 m down the street is
missed, and the same ramp is found once the vehicle drives closer.

## What it implies for deployment

Multi-view aggregation over the un-thinned pano stream should recover most of this —
the city run has 53,232 panos, and this benchmark deliberately thins to 30 m spacing so
the same ramp never appears twice. That work belongs in `sidewalk-auto-labeler`
(it answers "what are the curb ramps in this city?"), and the records already carry the
`lat`/`lng`/`camera_heading` needed for pixel → ground-ray association.

The gain lives on the **distance** axis, not the confidence axis: in the near field,
raising the threshold to 0.90 gives precision 1.000 but recall 0.282. "Seen close in
some view" has headroom that "seen confidently in some view" does not.

Measuring the real gain needs a split this benchmark does not have yet — a dense,
un-thinned corridor with ground truth at the **ramp** level in world coordinates rather
than per pano. Worth filing separately; it also reopens the deferred operating-point
question (#54/#55), since redundancy changes where the threshold should sit.

## Smaller notes

- **Negative check: 21 of 25 `empty`-group panos clean.** The other 4 held 10 real
  ramps — but **zero false detections**. Never wrong when it fired, only silent, which
  is the same far-field story. (Morgantown was 25/25.)
- **Vintage is a non-finding.** 2020-05 gives P 0.987 / R 0.704, 2023-10 gives
  P 0.952 / R 0.742; small counts, overlapping intervals. Detection density is
  near-identical (1.86 vs 1.80 per pano), so vintage is not confounded with sampling.
- **No `review_notes` on this split.** Nothing about the rubric fought back the way it
  did in Budapest, but the reviewer's own confidence rating is not on the record —
  worth adding on a re-review.
- `camera_model` casing differs by vintage (`Trimble mx7` vs `Trimble MX7`). No code
  reads the field; only matters for hand-grouping.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -188,6 +188,26 @@ Smaller notes:
   across the two (1.86 vs 1.80 per pano), so vintage is not confounded with sampling.
 - **3 duplicate marks** (2 in the unbiased subset). Scored as false positives by default; with
   `--lenient-duplicates` precision is 0.977 all-panos / 0.972 unbiased.
+- **How the distances above were computed — no depth estimation is involved.** An equirectangular
+  pano maps the vertical axis linearly to elevation, so a point at `y` sits at depression
+  `(y - 0.5)·180°` below the horizon, and on flat ground a camera at height `H` sees it at
+  `H / tan(θ)`. The whole model is that one line, with `H` assumed to be 2.5 m.
+  - `H` **cancels out of every comparison here** — it scales all distances linearly, so the
+    hit/miss ratio is 1.79 whether the mount is 1.8 m or 3.5 m. Only the metre labels move.
+  - **The metre labels are soft near the horizon, which is exactly where the misses sit.**
+    `H/tan(θ)` is steep there: at `y = 0.51` a shift of 0.01 (≈20 px in the 2048-tall render)
+    takes the estimate from 79.6 m to 39.7 m. Read "20.4 m" as *far*, not as a measurement.
+  - **The conclusion does not depend on any of that**, because `H/tan((y-0.5)π)` is strictly
+    monotonic in `y` — so the claims are rank statements that survive any monotonic distance
+    model, including a correct one. Testing raw `y` with no distance model at all: Mann-Whitney
+    **z = -5.69** (p < 1e-7), and **P(a random missed ramp sits closer to the horizon than a
+    random detected one) = 0.716**. The 80% / 97% figures above are likewise rank statements.
+    Only the cutoff table depends on the metres, and only for where the cutoffs sit.
+  - **Unverified assumptions**: `camera_pitch` and `camera_roll` are `null` on all 125 panos, so
+    the horizon is assumed to be exactly at `y = 0.5` (plausible — Mapillary equirects are
+    gravity-aligned at stitch time and the MX7 is a fixed mount — but unchecked), and the ground
+    is assumed flat. Annapolis is coastal and fairly flat, which is why this is defensible here;
+    **do not reuse this method on morgantown's hillsides** without accounting for grade.
 - **This split carries no `review_notes`.** Unlike budapest, nothing about the rubric fought back
   here, but the reviewer's own confidence rating is not on the record — worth adding on a
   re-review.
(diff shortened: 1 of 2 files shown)


jonfroehlich:
Note on where the distance figures come from, since it's a fair thing to ask before
quoting them: **there is no depth estimation involved.** They are flat-ground
trigonometry on the equirectangular elevation axis —

```
dep = (y_normalized - 0.5) * pi     # depression below horizon
d   = H / tan(dep),  H = 2.5 m      # assumed camera mount height
```

Pushed a note to the split's section in `benchmark/README.md` recording the method and
its limits. The short version:

- **Camera height cancels from every comparison.** It scales all distances linearly, so
  the hit/miss ratio is 1.79 whether the mount is 1.8 m or 3.5 m — only the metre labels
  move.
- **The metre labels are soft near the horizon, which is where the misses sit.** At
  `y = 0.51`, a 0.01 shift (≈20 px in the 2048-tall render) takes the estimate from
  79.6 m to 39.7 m. "20.4 m" means *far*, not a measurement.
- **The conclusion doesn't rest on the metres.** `H/tan((y-0.5)π)` is strictly monotonic
  in `y`, so the claims are rank statements that survive any monotonic distance model.
  Dropping the distance model entirely and testing raw `y`: Mann-Whitney **z = -5.69**
  (p < 1e-7), and P(a random missed ramp sits closer to the horizon than a random
  detected one) = **0.716**. The 80% / 97% figures are rank statements too; only the
  cutoff table depends on the metres.
- **Two unverified assumptions**: `camera_pitch`/`camera_roll` are `null` on all 125
  panos, so the horizon is assumed to be exactly at `y = 0.5`; and the ground is assumed
  flat. Annapolis is coastal and flat enough for that to hold — morgantown's hillsides
  would need grade accounted for, so the note warns against reusing the method there.

Trustworthy metres would need Mapillary's SfM camera poses or a monocular depth model,
but that would only sharpen the labels, not change the finding.

Issue #90, opened 2026-07-30
Tune the YOLO baseline's peak LR / warmup before reporting it — the collapse is an untuned default (#51)


jonfroehlich:
Parent: #51 · Related: #70 (stabilized rerun), #71 (pre-registered protocol)

## Why this is a fairness issue, not a performance one

The supervised-YOLO baseline currently runs **Ultralytics defaults**. Per the provenance
block in `scripts/model_comparison/yolo_baseline/README.md`, `optimizer=auto` resolves to
**`MuSGD(lr=0.01, momentum=0.9)`** in all six jobs — overriding the passed `lr0`/`momentum` —
and the realized schedule peaks at **`lr/pg0 = 0.029`** at the end of `warmup_epochs=3.0`.

**Every config's validation collapse begins at exactly that peak.** That is documented with
figures (`figures/fig1`–`fig4`) and the record rules out the alternatives: not a crash (no
NaN/Inf, train loss falls straight through), not preemption (requeues don't line up), not
small batch (the `y11l_pano` / `y26_pano` batch=4 control), not architecture (both families
collapse and both recover), not a data fault.

The record's own conclusion is that the peak LR is too high for this task — 150k
single-class, small-object images fine-tuned from COCO weights — and it flags that
hypothesis as **untested**.

That leaves one hole in an otherwise airtight protocol. #71 exists to defend the baseline
against "you sandbagged it," and every selection decision is already made on validation
data under rules fixed in advance. But **an untuned default that visibly destabilizes
training is still a sandbagging exposure**: a reviewer can say we reported YOLO at a
learning rate its own curves show to be wrong, and they would be right. Until the
hypothesis is tested we cannot distinguish "YOLO is unstable on this task" from "we ran
YOLO at a bad LR."

## Scope — deliberately bounded in both directions

Fairness runs both ways. RampNet's published recipe is itself lightly tuned (**one epoch,
~9.4k steps, constant LR, no scheduler** — see #84), so an exhaustive HPO on YOLO would
overcorrect and produce a differently-unfair comparison. The goal is **parity of tuning
effort**, not a maximally-tuned YOLO.

So: a small, principled grid on **one** config, decided here before it runs.

- **Config:** `y11l_pano` — it collapsed *and* recovered, so the effect is visible in both
  directions, and its ~2.6–2.8 h epoch is among the cheapest.
[...]

Issue #84, opened 2026-07-29
Stage 2 trains for 1 epoch at constant LR (~9.4k steps): measure the epoch curve, and with it the auto-label ceiling


jonfroehlich:
## What the recipe actually is

Stage 2 training, as committed and as used for the paper:

| | value | where |
|---|---|---|
| epochs | **1** ("default: 1, as in the paper") | `stage_two/train.py:35` |
| batch size | **1 per rank** | `train.py:261` |
| world size | 4 nodes × 4 GPUs = **16** | `run_train.slurm:4-6` (partition `ckpt-all`, scavenger) |
| ⇒ global batch | 16 | |
| train split | 150,063 panos | of 214,376 total |
| ⇒ **total optimizer steps** | **9,379** | 150,063 ÷ 16 |
| optimizer | Adam, lr **1e-5, constant** | `PRESET_LR['scratch']`, `train.py:269` |
| **LR schedule** | **none — no decay, no warmup** | no scheduler anywhere in the file |
| AMP | on (GradScaler + autocast) | `train.py:270,357` |
| selection | lowest mean **val heatmap loss** → `best_model.pth` | `train.py:449-452` |
| per-epoch checkpoints | **already saved** as `epoch_{N}_step_{S}.pth` | `train.py:435-446` |

The last row matters for cost: the whole epoch curve can be read off **one** run with no code change.

## Why this is worth measuring

Epochs is the wrong unit for cross-model comparison; steps is the right one:

| | steps/epoch | configured | total steps |
|---|---|---|---|
| **RampNet Stage 2** | 9,379 | 1 epoch | **9.4k** |
| COCO detection "1x" — the field's *shortest* standard schedule | ~7.4k | 12 epochs | ~88k |
| YOLO baseline, pano (#51) | 37,516 | 60 epochs | ~2.25M |

So Stage 2 trained for roughly **one-tenth of the shortest standard detection schedule**, and the supervised baseline it is being compared against gets ~60× more exposure to the same data.

**But it is genuinely not obvious that 1 is wrong**, which is exactly why it needs measuring rather than arguing:

- It is a *fine-tune* of a pretrained ConvNeXtV2-base (`convnextv2_base.fcmae_ft_in22k_in1k_384`) at lr 1e-5, not from-scratch training. The 12–36 epoch conventions were set for models starting much colder.
[...]

PR #91, opened 2026-07-30
Onboard Tillicum: cluster doc, Slurm port, and first-login recon (#51/#70)

jonfroehlich:
Onboards the lab onto **Tillicum**, UW-IT's usage-billed H200 cluster (makelab provisioned
2026-07-29), and prepares the YOLO baseline to move there. Nothing in flight is touched.
(Additive-only until `3dbbf46`, which adds an optional, off-by-default worker
pass-through to `hyak_yolo_runbook.sh` — see **Review fixes** below.)

## Why

klone's `ckpt` scavenger partition has stopped converting compute into results for #51.
Measured 2026-07-30:

- **496.5 GPU-hours** consumed on the baseline since 2026-07-24 (`sacct`, all arms)
- **170 job restarts in one night** across five jobs (39 / 37 / 33 / 33 / 28)
- **zero completed epochs** on five of six arms across that night; four still hold an ep1
  `best.pt`

The binding constraint is not the nominal 8.24 h checkpoint slice but the **effective**
contiguous run the partition hands out, which has collapsed to minutes. Any arm whose epoch
exceeds that can never complete one. **Tillicum does not preempt** — that is the entire
reason to pay for it.

This matters beyond convenience. The #71 protocol already concedes that every reportable
checkpoint is **undertrained**, so any benchmark number is a *lower bound* on supervised
YOLO. "We ran out of patience on a scavenger partition" is not a defensible reason for a
baseline to underperform, and "you undertrained YOLO" is the first objection any reviewer
raises against the paper's central comparison.

## What's here

**`docs/tillicum.md`** — the first general cluster doc in the repo (we have task runbooks
for klone, but nothing describing a cluster itself). QoS table, cost model, storage terms,
software story, and a directive-by-directive Slurm migration diff.

The headline for anyone submitting: **our `run_yolo_train.slurm` is not submittable on
Tillicum as written.** No partitions exist (QoS only), the default QoS caps at 24 h against
our `--time=72:00:00`, and `--cpus-per-task=12` with one GPU violates a hard 8 CPU : 1 GPU
ratio and is rejected outright.

**`scripts/model_comparison/run_yolo_train_tillicum.slurm`** — a port kept as a separate
file rather than a branch inside the klone script, since nearly every directive differs and
the klone version is the preserved record of the #51 runs and should not churn.

The non-obvious part is that it separates the GPU **allocation** from the ultralytics
**device**. Tillicum fixes CPUs at 8 per GPU, so 16 dataloader workers for the I/O-bound
tiles arm means allocating 2 GPUs — but letting ultralytics use both puts it in DDP, which
changes effective batch and LR dynamics and breaks comparability with the klone single-GPU
runs. So we allocate 2 and pin `DEVICE=0`, paying for an idle GPU on purpose. That isolates
dataloader throughput as the only changed variable, which is exactly the question: does the
2× bill buy more than a 2× speedup?

It also sets `save_period=5`. Ultralytics defaults to keeping only `last.pt`/`best.pt`,
which permanently forecloses a **compute-matched** comparison — RampNet's published model
trained ~1 epoch / ~9.4k steps at constant LR (#84) while this baseline gets 60 epochs with
a schedule, and reporting both the converged *and* the matched point is much stronger than
the converged one alone. `results.csv` has per-epoch metrics but no weights, so it cannot be
reconstructed later.

**`scripts/tillicum_recon.sh`** — read-only, ~10 s, submits nothing, costs nothing. Settles
in one pass every item the doc currently marks UNVERIFIED.

## Measured, not assumed

**The data that has to cross: 286 GB in ~911,000 files** (tiles 210 GB / 718,415; pano
76 GB / 192,938).

- 286 GB **fits the 1 TB allocation** comfortably — storage quota is not a blocker.
- ~911k files **is** the blocker. `du -sh` over the tiles tree beat three separate timeouts
  before completing on a 550 s budget, and even `ls -lU | head -201` on `tiles/images/train`
  timed out. That is metadata throughput, not bandwidth.

Consequences recorded in the doc: transfer archives rather than the tree (a per-file rsync
pays ~911k round trips), run the `tar` inside a Slurm job rather than on a login node, stage
`pano` first so the smoke test is unblocked while `tiles` moves, and **time the untar** —
it is our first direct measurement of whether Tillicum's flash fixes the bottleneck the
tiles training arm is also probing.

## Status

*This section originally read "nothing here has run", which was true when the PR was
opened on 2026-07-30 and has not been true since 2026-07-31. Rewritten 2026-08-17.*

Three jobs have run and are recorded in the doc:

| job | what | result |
|---|---|---|
| `198638` | `debug` smoke test | environment works on H200 / Rocky 9; also surfaced the `hyakusage` vs `sacctmgr` billing disagreement |
| `198910` | data regeneration from HF, 4 h 40 m, $4.20 | dataset verified equivalent to klone — identical counts *and* identical md5 of the sorted filename list across all eight directories |
| `217298` | one `y11x_tiles` epoch, pre-registered | **2.38×** klone's L40S (5.0 vs 2.1 it/s), and it is the filesystem rather than the GPU |

The HostName is settled: `tillicum.hyak.uw.edu` is correct and the get-started page's
`tillicum.hyak.edu` is wrong. **The per-user home path is settled** (2026-08-18, `1b008e3`):
`cd ~ && pwd` on a login node returns `/gpfs/home/jfroehli`, so the as-run `~/RampNet` and
`run_yolo_data_prep_tillicum.slurm`'s `/gpfs/home/$USER/RampNet` are the same directory and
no script needed changing.

Companion change outside this repo: `dotfiles` gains a `tillicum` target for `wsl-ssh.ps1`
plus its operating rules, which are auto-imported into every Claude session.

## Review fixes (3dbbf46)

A deep review of the branch found 15 issues; all are fixed in `3dbbf46`. Two mattered:

- **The 2-GPU dataloader trick did nothing.** `run_yolo_train_tillicum.slurm` never passed
  `workers` to `train()`, and ultralytics defaults to `workers: 8` capped at
  `min(os.cpu_count() // nd, workers)` — so the pool never grew with the allocation. A
  `--gres=gpu:2 --cpus-per-task=16` run would have paid 2× for 8 workers and reported "no
  faster", answering the question above for the wrong reason. `WORKERS` now binds to
  `SLURM_CPUS_PER_TASK`, and it has to be set on the first link of a chain because
  `resume=True` inherits the loader count from `last.pt`.
- **`tillicum_setup_env.sh` could not run the data stage that depends on it** — no
  `datasets`, no `huggingface_hub`, but `download_dataset.py` imports `datasets` at module
  scope. `198910` completed, so that package was hand-installed and lived in no committed
  script. Now installed and gated, along with `torchvision` from the cu126 index.

The rest: a `mksquashfs` append hazard on requeue, the equivalence check made a gate rather
than a report (it promised "stop and diff" while exiting 0), the two prep guards that were
dropped by routing through the runbook, `CHAIN=<n>` to replace the continuation that
dropping `--requeue` gave up, a usage example that trained yolo11l under a y26 name, an
unreachable `|| echo` in the recon, unpinned installs in the smoke test, and the superseded
"keep prep on klone" directive marked as such where it sits.

**Note this PR is no longer additive-only.** `hyak_yolo_runbook.sh` gains an optional
`--workers` / `OPENBLAS_NUM_THREADS` pass-through in its `prep` stage. Behaviour outside a
Slurm allocation is unchanged, so the #51 klone runs are unaffected.

## Related

- #51 — supervised YOLO baseline (the runs that motivated this)
- #70 — stabilized rerun; this is its unblocking dependency
- #90 — LR/warmup sweep; also gated on non-preemptable capacity

### DIFF
--- a/docs/tillicum.md
+++ b/docs/tillicum.md
@@ -153,15 +153,27 @@ Two conclusions:
   `ls -lU | head -201` on `tiles/images/train` timed out. That is metadata throughput,
   not bandwidth.
 
-**So transfer an archive, never the tree.** `tar` per split on klone → move a handful of
-large files → untar on Tillicum. A per-file `rsync`/`scp` would pay ~911,000 round trips.
-Run the `tar` inside a Slurm job, **not on a klone login node** — that is exactly the
-heavy-login-process reap that kills the SSH master. Note the untar on `/gpfs` is itself
-metadata-heavy, so time it: it is our first real measurement of whether Tillicum's flash
-actually fixes this, which is the same question the tiles training arm is asking.
+**So pack into SquashFS, never copy the tree** — see
+`scripts/model_comparison/pack_yolo_dataset.slurm`, and UW-IT's own recommendation in the
+answers below. A per-file `rsync`/`scp` pays ~911,000 round trips; `tar` fixes the
+transfer but not the destination, since untarring recreates all 911k files on `/gpfs` and
+pays the metadata cost permanently. A SquashFS image is mounted read-only, so the
+destination only ever holds **one file**.
+
+Run the pack inside a Slurm job, **not on a klone login node** — that is exactly the
+heavy-login-process reap that kills the SSH master.
+
+**The subtle part is the Ultralytics label cache.** Ultralytics writes
+`labels/<split>.cache` and validates it against a hash of the label+image **absolute
+paths**. So (a) the `.cache` files must be *inside* the image — they exist on klone
+already and a read-only mount cannot create them — and (b) mounting at a different path
+than they were built under silently invalidates them and forces a full ~911k-file
+rescan. Budget one slow first epoch on Tillicum to regenerate them, and do not mistake
+it for the steady-state cost. Getting this wrong is invisible: training still works, it
+is just permanently slow.
 
 Staging order: **`pano` first (76 GB, 193k files)** — it unblocks the smoke test and the
-pano epoch-time measurement while the much larger `tiles` archive moves behind it.
+pano epoch-time measurement while the much larger `tiles` image builds behind it.
 
 > The docs describe demo accounts as receiving "100 GB of dedicated project storage,"
 > while our provisioning email says 1 TB. Assume 1 TB (the email is specific to us) but
@@ -233,17 +245,51 @@ Record the measured epoch times back into this file and into #70.
 - **Budget changes:** contact UW-IT; enforcement semantics (blocks new submissions vs.
   cancels running jobs) are **UNVERIFIED** and were asked about on 2026-07-30.
 
-## Open questions
+## Answers from UW-IT (Sumaiya Sathar, 2026-07-30)
 
-Sent to UW-IT 2026-07-30, unanswered at time of writing:
+All six questions answered. Several change the plan, so they are recorded here rather
+than left in a mailbox.
 
-1. Does budget enforcement block *new submissions* or *cancel running jobs*?
-2. Is there any way to get more CPUs per job without the GPU billing multiplier?
-3. Guidance for many-small-file datasets on `/gpfs`; is packing into an archive or
-   container format recommended?
-4. Storage: quota-increase path, and what "purged at the end of the project" means.
-5. Do the 100 free demo hours expire?
-6. How to request `long` QoS access.
+1. **Budget enforcement is safe.** Set to **$1,500/month, enforcement active**. Enforced
+   budgets **do not cancel running jobs** — a job already running continues to
+   completion; enforcement only **blocks new submissions** until the cap is raised or
+   the period rolls over. So the cap cannot destroy an in-flight experiment, which was
+   the only reason to have preferred warn-only.
+2. **No CPUs without GPUs — confirmed, and it is a rate-model constraint, not a
+   scheduling one.** There is no QoS granting extra cores without the matching GPU
+   allocation; UW-IT is considering it, but the UW-approved rate model cannot charge for
+   CPUs independently, and unbalanced nodes are undesirable. **So the 2-GPU dataloader
+   trick is the only lever**, and its 2× billing is unavoidable — see the trap above.
+3. **Many small files → `/gpfs/scrubbed`, and pack into SquashFS.** This is the big one.
+   Active many-small-file datasets belong on **`/gpfs/scrubbed`** (larger capacity, has
+   an automatic cleanup policy) rather than the 1 TB project quota, and UW-IT explicitly
+   recommends **SquashFS** for this shape. Data Commons is an option if the dataset can
+   be public.
+4. **Storage grows in 1 TB increments** on request to `help@uw.edu` (subject "Tillicum")
+   with a workflow justification — it is shared active-compute storage, not archival.
+5. **No project end date.** "Purged at the end of the project" simply means that if we
+   stop using Tillicum we must copy data off so the space can be reclaimed. Nothing
+   expires on a timer.
+6. **1 TB is correct for us** — the 100 GB figure in the public docs applies to *demo*
+   accounts. Providing a worktag makes it a regular account, which is why we got 1 TB.
+7. **The 100 free GPU hours do not expire.** They remain until used.
+8. **`long` QoS requires the Special QoS Access Request Form** — still to be submitted,
+   then reviewed. Not a blocker for the measurement phase (everything fits in 24 h), but
+   it gates comfortable production runs.
+
+### What this changes
+
+- **Dataset goes to `/gpfs/scrubbed`, not `/gpfs/projects/makelab`.** The 1 TB project
+  quota is for artifacts we want backed up (checkpoints, results), not the 286 GB of
+  training data. This also removes storage pressure from the plan entirely.
+- **Pack with SquashFS, not tar.** See
+  `scripts/model_comparison/pack_yolo_dataset.slurm`. tar fixes the *transfer* but not
+  the *destination* — untarring recreates all ~911k files on `/gpfs` and pays the
+  metadata cost permanently. A SquashFS image is mounted read-only and the destination
+  only ever holds one file.
+- **Enforced budget is safe to leave on**, so the runaway-spend guard costs us nothing.
+- **The `DEVICE=0` / 2-GPU design in the Tillicum launcher is now confirmed necessary**
+  rather than a workaround pending a better answer.
 
 ## Related
 
--- a/docs/tillicum.md
+++ b/docs/tillicum.md
@@ -109,6 +109,33 @@ At that ceiling:
 
 Monitor with `hyakusage`.
 
+### `debug` is free — but the two cost tools disagree (measured 2026-07-31)
+
+The full QoS table, from `sacctmgr show qos`:
+
+| QoS | Priority | UsageFactor | MaxWall | per-user cap |
+|---|---:|---:|---|---|
+| `debug` | 50 | **0.000000** | 1 h | 1 GPU, 8 CPU, 200 G, 1 node |
+| `normal` | 25 | 1.000000 | 1 day | 48 GPU |
+| `interactive` | 35 | 1.000000 | 8 h | — |
+| `urgent` | 200 | 1.000000 | 3 days | 64 GPU |
+| `long` | 25 | 1.000000 | 7 days | — |
+| `wide` | 25 | 1.000000 | 1 day | 96 GPU |
+
+`debug` bills at **UsageFactor 0** and carries *higher* priority than `normal` (50 vs
+25), so every environment and throughput probe belongs there. That is the basis for the
+"costs nothing" claim in `scripts/tillicum_smoke.slurm`.
+
+**Caveat, unresolved:** `hyakusage` does not agree. The smoke job (`198638`, 2 min on
+`debug`) shows up in its QoS breakdown as **0.03 GPU-hours, $0.03** — i.e. raw
+wall-clock × $0.90 with the 0.0 UsageFactor apparently *not* applied, even though
+`hyakusage`'s own header says "billable GPU hours = raw GPU hours × QOS multiplier."
+Slurm's accounting config and the reporting tool are stating different things, and we do
+not know which one ITBill actually follows. Worth asking UW-IT, but not urgent: `debug`
+is capped at 1 h × 1 GPU, so the exposure is **at most $0.90 per job** even if
+`hyakusage` turns out to be the honest one. Do not quote a "free" figure from this
+without saying which tool it came from.
+
 ### The 2-GPU trap for I/O-bound arms
 
 This is the cost decision specific to our workload. Our tiles arm is **I/O-bound, not
@@ -249,8 +276,38 @@ are DockerHub and the NVIDIA NGC catalog.
 **Our `environment.yml` should not be assumed to transfer.** It pins linux-64 packages
 against **CUDA 11.8**, and Tillicum is H200 (sm_90) on Rocky 9. CUDA 11.8 nominally
 covers sm_90, but a stack built for klone's OS and driver is not a safe bet on a
-different distro and a newer card. **UNVERIFIED — do not plan around either outcome
-until someone tries it.** The low-risk path is an NGC PyTorch container.
+different distro and a newer card. **Still UNVERIFIED** — nobody has tried to solve
+`environment.yml` here, because the YOLO baseline does not need it. The low-risk path
+remains an NGC PyTorch container.
+
+### The YOLO stack, however, is VERIFIED (2026-07-31)
+
+`scripts/tillicum_setup_env.sh` builds it, and it reproduces the klone `#51` toolchain
+**exactly** — confirmed by running the interpreter, not by reading a lockfile:
+
+```
+3.11.15   torch 2.13.0+cu126   ultralytics 8.4.105
+```
+
+which is character-for-character what klone's training logs report
+(`Ultralytics 8.4.105 · Python-3.11.15 · torch-2.13.0+cu126`). It lives at
+`/gpfs/projects/makelab/$USER/envs/rampnet-yolo` — the backed-up 1 TB allocation rather
+than `scrubbed`, because the environment is small and annoying to rebuild while the
+dataset is huge and reproducible. It has since driven a full 4 h 40 m production job
+(`198910`) without incident.
+
+Two details in that script worth keeping if it is ever edited:
+
+- **Conda is not optional here.** Tillicum's *system* python is 3.9.25, and the cu126
+  wheel index tops out at torch 2.8.0 for 3.9 — so a naive `pip install torch
+  ultralytics` silently yields 2.8.0 + 8.4.113 and a baseline nobody can publish. The
+  conda module is the only way to get 3.11 on this cluster.
+- **The version check is a gate, not a report.** The script refuses to install a
+  substitute torch and exits non-zero on any mismatch. That matters because installing
+  `ultralytics` last can pull its own torch over the pinned one; the post-install assert
+  is what catches it. A mismatched toolchain produces plausible numbers and an
+  unpublishable comparison, which is the failure mode `#71`'s protocol exists to
+  prevent.
 
 ## Migrating our Slurm scripts
 
--- a/docs/tillicum.md
+++ b/docs/tillicum.md
@@ -1,0 +1,227 @@
+# Tillicum — access, migration, and what it costs
+
+Working notes for running RampNet jobs on **Tillicum**, UW-IT's usage-billed GPU
+cluster. The makelab group was provisioned on **2026-07-29**.
+
+This doc exists because Tillicum is **not** klone-with-different-hostnames — the
+scheduler model, the walltime ceiling, the CPU:GPU ratio, and the cost model all
+differ in ways that make our existing `.slurm` scripts **non-submittable as written**.
+See [Migrating our Slurm scripts](#migrating-our-slurm-scripts) for the diff.
+
+> **Sourcing.** Everything in the tables below is from the [Tillicum
+> docs](https://hyak.uw.edu/docs/systems/tillicum/scheduling-jobs) and the provisioning
+> email (2026-07-29), read on 2026-07-30. Sections marked **UNVERIFIED** are our
+> inference or are undocumented — we have not run a single job on Tillicum yet, so
+> nothing here is confirmed by experience. Correct this file from measurement, not
+> from re-reading the docs.
+
+## Why we care: the klone problem this solves
+
+The supervised-YOLO baseline (#51) ran on klone's `ckpt` scavenger partition, which is
+free but **preemptable**. As of 2026-07-30 that had stopped working as a compute
+strategy:
+
+- **496.5 GPU-hours consumed** on the baseline since 2026-07-24 (`sacct`, all arms).
+- **170 job restarts in one night** across five jobs (39 / 37 / 33 / 33 / 28).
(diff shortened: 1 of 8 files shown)


scripts/model_comparison/run_yolo_train_tillicum.slurm
```diff
@@ -0,0 +1,161 @@
...
+# roughly RampNet's ~1-epoch budget) is exactly the ep1 best.pt that four of the six
+# klone arms are still holding. The "one-epoch models" are the matched comparison.
+if os.path.exists(last):
+    print(f"[resume] {last} exists -> resuming", flush=True)
+    YOLO(last).train(resume=True)
+else:
+    bf = float(batch)
+    batch_arg = int(bf) if bf.is_integer() else bf
+    dev = [int(d) for d in device.split(",")] if "," in device else int(device)
+    kw = dict(
```
jonfroehlich, line 206: `kw` never passes `workers`, so the 2-GPU trick this script is built around cannot do anything.

Ultralytics' default is `workers: 8` (`cfg/default.yaml`), and `build_dataloader` caps at `min(os.cpu_count() // nd, workers)`. Allocating 16 CPUs still gives 8 loader workers. Our own record confirms it: job `217298` echoed `workers=8`.

So `sbatch --gres=gpu:2 --cpus-per-task=16` doubles the bill (~$43/day) and changes nothing, and the pre-registered question -- does the 2x bill buy more than a 2x speedup -- would come back "no" for the wrong reason.

One more thing for whoever fixes this: `resume=True` on line 137 reuses the args saved in the checkpoint, so `workers` is unreachable on a resumed run even once it is added here.

scripts/tillicum_setup_env.sh
```diff
@@ -0,0 +1,91 @@
...
+fi
+
+echo "=== installing pinned torch ==="
+"$PY" -m pip install --upgrade pip
+"$PY" -m pip install "torch==$TORCH" --index-url "$TORCH_INDEX"
+
+echo "=== installing pinned ultralytics ==="
+# --no-deps would risk a broken install; instead pin ultralytics and let it resolve its
+# own deps, then assert torch was not silently upgraded underneath us.
+"$PY" -m pip install "ultralytics==$ULTRA"
```
jonfroehlich: This env cannot run the data stage that depends on it.

`run_yolo_data_prep_tillicum.slurm` sets `PYBIN` to exactly this env and then calls `hyak_yolo_runbook.sh data`, which runs `download_dataset.py` -- whose line 3 is `from datasets import load_dataset`. ultralytics does not pull `datasets` in.

The klone runbook has it right: `pip install ultralytics datasets huggingface_hub` (`hyak_yolo_runbook.sh:89`), with the comment "datasets+hf_hub for download_dataset.py".

Job `198910` did complete this stage, which means `datasets` was installed by hand during the session and exists in no committed script. That is the clean-clone replication failure CLAUDE.md calls out.

scripts/model_comparison/pack_yolo_dataset.slurm
```diff
@@ -0,0 +1,123 @@
...
+# ---------------------------------------------------------------------------------
+echo "[check] label caches that will be packed:"
+ls -la "$SRC/$DS/labels/"*.cache 2>/dev/null || echo "  NONE -- expect a full rescan on first use"
+
+# -noD: do NOT compress data blocks. The payload is JPEG, already compressed; deflating
+# it would burn hours of CPU for ~nothing and make reads slower. Metadata/inodes are
+# still compressed, which is where the win actually is.
+# -no-xattrs: nothing here uses them, and they bloat the index.
+echo "[mksquashfs] starting $(date)"
+mksquashfs "$SRC/$DS" "$IMG" \
```
jonfroehlich, line 113: `mksquashfs` appends to an existing image by default, and the idempotency guard above has a hole for the partial case.

This job is `--requeue` on `ckpt`. A preemption mid-pack leaves `${DS}.sqfs` present but the `.sha256` missing, so `if [ -f "$IMG" ] && [ -f "$SUM" ]` on line 59 is false and the `rm -f` inside it never runs. Line 95 then packs into the leftover file.

Best case it dies on the partial superblock. Worst case it produces a duplicated image and line 104 writes a valid checksum over it, so the checksum discipline certifies a bad image.

Fix is `rm -f "$IMG"` unconditionally before packing, plus `-noappend`.

scripts/model_comparison/pack_yolo_dataset.slurm
```diff
@@ -0,0 +1,123 @@
...
+cat <<EOF
+
+[ok] $DS packed.
+
+Next, on Tillicum -- note /gpfs/scrubbed, not the 1 TB /gpfs/projects/makelab quota
+(UW-IT recommends scrubbed for active many-small-file datasets):
+
+  # mount read-only, no untar, no 911k files ever created on /gpfs
+  apptainer exec --nv \\
+      --bind /gpfs/scrubbed/\$USER/${DS}.sqfs:/data/${DS}:image-src=/ \\
```
jonfroehlich, line 147: The label cache is not the only absolute path inside the image.

`prepare_yolo_dataset.py` writes `path: {cfg.out}` into `data.yaml`, so the packed copy carries e.g. `/gscratch/scrubbed/jfroehli/yolo/tiles`, which does not exist on Tillicum. Ultralytics resolves `train: images/train` against that absolute `path:`, so the dataset load fails outright no matter where the image is mounted.

The comment block above concludes that mounting "at a stable path" is enough. That holds for the cache hash but not for `data.yaml`, and a read-only mount cannot rewrite it -- this needs a writable overlay or a bind-mounted `data.yaml`.

It fails loudly rather than silently, but it fails on the day the transfer route finally exists.

scripts/model_comparison/run_yolo_data_prep_tillicum.slurm
```diff
@@ -0,0 +1,92 @@
...
+
+echo
+echo "=== STAGE: prep (tiles + pano) ==="
+date
+bash hyak_yolo_runbook.sh prep
+
+echo
+echo "=== VERIFICATION — counts must match the klone record ==="
+date
+for ds in tiles pano; do
```
jonfroehlich, line 95: The header says a mismatch means "stop and diff", but this block only prints. There is no comparison and no non-zero exit, so a run that produces 557,398 tiles instead of 557,413 exits 0 and reads as green in the log.

`tillicum_setup_env.sh:85` is the right shape for this -- it makes its equivalent check a hard gate with `sys.exit(0 if ok else 1)`.

Also, the four `labels/` directories the doc verifies are not checked here at all.

scripts/model_comparison/run_yolo_data_prep_tillicum.slurm
```diff
@@ -0,0 +1,92 @@
...
+
+echo
+echo "=== STAGE: data (HF download, ~462 GB) ==="
+date
+bash hyak_yolo_runbook.sh data
+
+echo
+echo "=== STAGE: prep (tiles + pano) ==="
+date
+bash hyak_yolo_runbook.sh prep
```
jonfroehlich, line 73: Going through the runbook drops both guards `run_yolo_prep.slurm` adds for this exact call: `--workers $SLURM_CPUS_PER_TASK` and `OPENBLAS_NUM_THREADS=1` ("one BLAS thread per worker (no RLIMIT_NPROC storm)").

Without them `prepare_yolo_dataset.py` falls back to `--workers max(1, (os.cpu_count() or 4) - 1)`. `os.cpu_count()` reports the whole node and ignores the cgroup, so on an XE9680 this forks tens of workers onto the 8 CPUs we allocated, each with its own BLAS threads and each holding a decoded pano against `--mem=200G`.

`198910` survived it, so this is latent rather than a failure that happened. But it is a silent divergence from klone on a job we are asking to treat as equivalent, and the margin depends on which node it lands on.

scripts/model_comparison/run_yolo_train_tillicum.slurm
```diff
@@ -0,0 +1,161 @@
...
+#   --time=72:00:00           ->  24:00:00       normal QoS caps at 24 h. 7-day jobs
+#                                                need `long` QoS (pre-approval).
+#   --gpus-per-node=1         ->  --gres=gpu:N   Tillicum's documented spelling.
+#   --cpus-per-task=12        ->  8 x N GPUs     HARD 8:1 ratio; 12-on-1 is REJECTED.
+#   (no mail directive)       ->  --mail-type=NONE   Be explicit so the site default
+#                                                cannot decide for us (on klone an
+#                                                inherited END,FAIL,TIME_LIMIT produced
+#                                                ~1 email per minute under preemption).
+#
+#   YOLO_DATA=$DATA/pano/data.yaml YOLO_IMGSZ=1280 BATCH=4 NAME=y26_pano_till \
```
jonfroehlich: This example trains the wrong model. `YOLO_CKPT` defaults to `yolo11l.pt` (line 61), so copy-pasting this gives a YOLO11l run stored under a y26 name. The only signal is `base: yolo11l.pt` in the log header, which is easy to miss, and the weights would then be compared against the #51 y26 arms as if they were the same architecture.

Every example in the klone script sets `YOLO_CKPT` explicitly.

docs/tillicum.md
```diff
@@ -0,0 +1,585 @@
...
+cancel or preempt jobs that are already running." That is the entire reason to pay.
+
+## Access
+
+```bash
+ssh <UWNetID>@tillicum.hyak.uw.edu
+```
+
+Duo 2FA, same as klone — **Claude cannot authenticate this.** The `wsl-ssh.ps1` helper
+in `dotfiles` has no `tillicum` target yet; adding one (with a `ControlPersist` master,
```
jonfroehlich: This section is stale, and so is the PR description.

`tillicum_recon.sh:9` and `tillicum_setup_env.sh:25` both document themselves as run via `wsl-ssh.ps1 tillicum script ...`, and the PR body still says "nothing here has run / nobody has logged in" while this doc reports jobs `198638`, `198910` and `217298` with timings, costs and a 2.38x throughput ratio.

The home path is the one that will bite someone. Recon item 1 calls it "the top unknown", the recon evidently ran (`tillicum_smoke.slurm` cites its conda/apptainer findings), but the answer was never written back: the as-run commands here use `~/RampNet` while `run_yolo_data_prep_tillicum.slurm:42` still hardcodes `/gpfs/home/$USER/RampNet` as a guess.

As it stands nobody can tell which UNVERIFIED markers are still open.

jonfroehlich (commented):
Deep review of the whole branch. 15 findings; I checked each against the code at `3831513` rather than reasoning from the diff alone.

Two are worth fixing before anything else runs on Tillicum:

- **The 2-GPU dataloader trick does nothing.** `run_yolo_train_tillicum.slurm` never passes `workers` to `train()`, so ultralytics stays at its default of 8 no matter how many CPUs we allocate. The script header, the doc and the PR body all rest on that lever, and job `217298` already echoed `workers=8`. As written, `--gres=gpu:2 --cpus-per-task=16` doubles the bill for nothing.
- **`tillicum_setup_env.sh` cannot run the data stage that depends on it** -- no `datasets`, no `huggingface_hub`, but `download_dataset.py` imports `datasets` at module scope. Job `198910` completed, so that package was installed by hand and is in no committed script.

The rest are smaller: a `mksquashfs` append hazard on requeue, an equivalence check that prints instead of gating, two prep guards dropped by routing through the runbook, a usage example that trains yolo11l under a y26 name, and a set of stale UNVERIFIED markers -- the PR description still says nothing has run.

Details inline.

docs/tillicum.md
```diff
@@ -0,0 +1,585 @@
...
+
+192 × NVIDIA **H200** (141 GB HBM3e, NVLink 4.0), 8 GPUs per node, 24 × Dell XE9680,
+1,536 Intel Emerald Rapids cores, 400 Gbps NDR InfiniBand, ~3 PB flash, Rocky 9.
+
+Per-GPU binding: **200 GB system RAM and 8 CPUs**. This ratio is not negotiable —
+the docs are explicit that exceeding 8 CPUs per GPU requires requesting *additional
+GPUs*. **CPU-only jobs are prohibited**; every job must request ≥1 GPU.
+
+> Operational consequence: our CPU-side data prep (`run_yolo_data.slurm`,
+> `run_yolo_prep.slurm`) has no business on Tillicum. Keep prep on klone, train on
```
jonfroehlich, line 108: This directive is contradicted by a script in the same PR. `run_yolo_data_prep_tillicum.slurm` runs prep on Tillicum, and the "regenerate, don't transfer" section 120 lines down explains why it had to. A reader who stops here concludes the new script is a mistake.

Needs an inline correction pointing forward -- caveats travel with the numbers.

scripts/model_comparison/run_yolo_train_tillicum.slurm
```diff
@@ -0,0 +1,161 @@
...
+#       sbatch scripts/model_comparison/run_yolo_train_tillicum.slurm
+#
+# COST. Tillicum BILLS: $0.90/GPU-hour where GPU-hour = elapsed x N GPUs. A 24 h
+# 2-GPU job is 48 GPU-hours (~$43). Do not leave allocations idle. Check `hyakusage`.
+#SBATCH --qos=normal
+#SBATCH --job-name=yolo_curb_ramp_train
+#SBATCH --time=24:00:00
```
jonfroehlich, line 36: Dropping `--requeue` was right (nothing preempts here), but nothing replaces the unattended continuation it bought. `--time=24:00:00` is a hard stop, and the comment at 124-126 says to "chain jobs and resume" -- there is no chaining in this script: no `--dependency=afterany` resubmit, no wrapper.

The doc frames this as the fallback behind requesting `long` QoS, so it is known rather than an oversight. But at ~3.0 h/epoch for tiles it means roughly five manual resubmissions over five days, each of which has to be remembered, which is the same silent-stall failure mode this PR exists to fix.

scripts/model_comparison/run_yolo_train_tillicum.slurm
```diff
@@ -0,0 +1,161 @@
...
+#SBATCH --nodes=1
+#SBATCH --ntasks=1
+#SBATCH --mail-type=NONE
+#SBATCH --output=logs/yolo_train_till_%j.out
+#SBATCH --error=logs/yolo_train_till_%j.err
+# gres/cpus/mem are NOT hardcoded -- see the GPUS block below, which must be kept in
```
jonfroehlich: Two problems in two lines.

The comment says gres/cpus/mem are "NOT hardcoded" immediately above three `#SBATCH` lines that hardcode them, and it points at a "GPUS block below" that does not exist -- the only block below is DEVICE (line 59), which is about which GPU ultralytics uses, not the allocation.

The suggested override also omits `--mem`, so a 2-GPU job inherits 200G instead of the 400G our own 200 GB-per-GPU binding entitles it to. That halves RAM in exactly the configuration meant to raise dataloader throughput.

scripts/model_comparison/run_yolo_data_prep_tillicum.slurm
```diff
@@ -0,0 +1,92 @@
...
+    tiles  train  557413
+    tiles  val    161002
+    pano   train  150063
+    pano   val    42875
+
+  A mismatch means the regenerated dataset is NOT the one the #51 arms trained on.
+  Do not train against it until the difference is understood.
+EOF
+date
+echo "Cost so far: run `hyakusage`"
```
jonfroehlich: Backticks inside double quotes are command substitution, so this runs `hyakusage` and splices its billing table into the echo instead of printing the instruction. If it is not on PATH the line degrades to "Cost so far: run " with a "command not found" on stderr. Single quotes fix it.

scripts/tillicum_recon.sh
```diff
@@ -0,0 +1,83 @@
...
+
+echo
+echo "## 3. SCHEDULER  (expect: no partitions, QoS-driven)"
+echo "-- sinfo --"
+sinfo -o "%.14P %.6a %.10l %.6D %.6t %N" 2>&1 | head -10 | sed 's/^/  /'
+echo "-- QoS available to us --"
+sacctmgr -nP show assoc user="$USER" format=Account,QOS 2>&1 | sed 's/^/  /'
+echo "-- do we already have long/wide/urgent? --"
+sacctmgr -nP show assoc user="$USER" format=QOS 2>/dev/null | tr ',' '\n' \
+    | grep -iE "long|wide|urgent" | sed 's/^/  /' || echo "  (none -- long QoS needs the request form)"
```
jonfroehlich: The `|| echo` never fires. It is attached to a pipeline ending in `sed`, which exits 0 on empty input, so when grep finds nothing the pipeline still succeeds and the explanation is never printed. The operator gets a blank section and cannot tell "no special QoS" from "sacctmgr failed".

Line 23 and line 55 both get this right by putting the `||` inside the subshell before the sed.

scripts/tillicum_setup_env.sh
```diff
@@ -0,0 +1,91 @@
...
+if ! "$PY" -m pip index versions torch --index-url "$TORCH_INDEX" 2>&1 | grep -q "${TORCH}"; then
+    echo "!! torch $TORCH NOT available for $("$PY" -V). Available:"
+    "$PY" -m pip index versions torch --index-url "$TORCH_INDEX" 2>&1 | head -3
+    echo "!! Refusing to install a substitute -- that would break comparability with #51."
+    exit 1
+fi
+
+echo "=== installing pinned torch ==="
+"$PY" -m pip install --upgrade pip
+"$PY" -m pip install "torch==$TORCH" --index-url "$TORCH_INDEX"
```
jonfroehlich: torchvision is not installed from the cu126 index here, and the gate below does not check it, so the "matches klone exactly" claim has a hole. The klone runbook installs `torch torchvision` together from cu126 (`hyak_yolo_runbook.sh:87`).

As written, torchvision arrives transitively from PyPI via ultralytics with its own hard `torch==` pin. Usually benign -- a PyPI torchvision pinning `torch==2.13.0` is satisfied by `2.13.0+cu126` -- but the bad cases are pip replacing the pinned torch, or landing a torchvision built against a different CUDA, which lines 74-86 never inspect.

The comment at 69-70 anticipates exactly this. The assert just has a torchvision-shaped hole.

scripts/tillicum_smoke.slurm
```diff
@@ -0,0 +1,120 @@
...
+say "BUILD A MINIMAL ULTRALYTICS ENV"
+# A venv on top of the module python is far lighter than solving environment.yml, and
+# the only question that matters here is whether a CUDA-12 torch wheel drives an H200.
+if [ ! -x "$VENV/bin/python" ]; then
+    python3 -m venv "$VENV" 2>&1 | tail -2
+fi
+"$VENV/bin/python" -m pip install -q --upgrade pip 2>&1 | tail -2
+# cu126 wheels: Tillicum is sm_90, and our klone stack is torch 2.13.0+cu126 (see the
+# yolo_baseline record), so matching it keeps the comparison honest.
+"$VENV/bin/python" -m pip install -q torch --index-url https://download.pytorch.org/whl/cu126 2>&1 | tail -3
```
jonfroehlich: The comment above says matching the klone stack "keeps the comparison honest", but neither install is pinned. If `module load conda` fails, `python3` is the system 3.9.25 and the cu126 index tops out at torch 2.8.0 there -- which is the exact "2.8.0 + 8.4.113" combination the doc names as the trap, reported back as a green probe.

Low impact, since this is deliberately a probe and any recent torch answers "does a CUDA-12 wheel drive an H200". But it also builds into `$ENVROOT/ultra` on scrubbed while `tillicum_setup_env.sh` builds `envs/rampnet-yolo` on projects, so we end up with two divergent environments on the cluster.

PR #87, opened 2026-07-29
Refresh the YOLO baseline record: durable weights, ckpt ceiling, fork (#51)

jonfroehlich:
Refreshes `scripts/model_comparison/yolo_baseline/README.md` against the live runs. Issue #51.
Documentation only — no code, no behaviour change, no new numbers in the reported tables.

## Why

The committed record had drifted from reality in three ways, one of them actively false.

**1. A false statement.** "Where the weights live" still read:

> _TODO — stage to Hugging Face (`projectsidewalk/…`) or lab storage and record the URL here._

The weights are now staged and verified:

```
/gscratch/makelab/jonf/rampnet_yolo_baseline_51/
  <arm>/{best.pt,last.pt,results.csv,args.yaml}   # 5 arms, 1.8 GB
  MANIFEST.md                                     # job IDs, per-arm metrics, sha256 of every weight
```

All **10 weight files sha256-verified against source** (10 ok, 0 mismatched). They had lived
**only** on `/gscratch/scrubbed`, which auto-purges after ~21 idle days, and the run dirs date
from 07-26 — so anyone reading the old text would reasonably conclude the weights were still
one purge from gone, and either redo the staging or worry.

**2. Stale numbers, and one that mattered.** The status table was the 07-28 snapshot. Beyond
the refresh, the record claimed every checkpoint is undertrained and every benchmark number a
lower bound. That is **no longer uniform**:

- `y26_pano`'s `best.pt` is **epoch 18** — proven, not inferred: `best.pt` and `last.pt` are
  byte-identical (both sha256 `aac93e74…50bf`). mAP50-95 0.460 vs 0.413 at ep1.
- The other four arms' `best.pt` are frozen at **epoch 1** (mtimes 07-26 20:22/21:42/23:35/23:58,
  each ~one epoch after the 18:10 start, while `last.pt` is current). For `y26_tiles` this is
  exact: `best_fitness` = 0.35579 = its ep1 mAP50-95.

Reporting all five identically would **understate** `y26_pano` and **overstate** the rest, so the
caveat is now per-arm and says *"one-epoch model"* in those words where that is what it is.

**3. Two findings were missing entirely.**

**The ckpt slice ceiling** — a limitation *distinct* from the LR instability already documented.
The instability says scores got worse; this says an arm cannot advance at all. Across **158
scheduling segments** over 3.5 days, the longest slice ever granted was **8.24 h** and **none
[...]

PR #92, opened 2026-07-30
Record where the YOLO baseline weights actually live (#51)

jonfroehlich:
One-section fix to `scripts/model_comparison/yolo_baseline/README.md`.

## The problem

That record is meticulous about provenance — toolchain versions, resolved hyperparameters
(including `optimizer=auto` silently resolving to `MuSGD(lr=0.01)` and overriding the passed
`lr0`), the Slurm job↔config map, dataset prep decisions, and two documented self-corrections.

Its "Where the weights live" section was still a TODO, for the one artifact in the whole
record that **cannot be regenerated**.

## The fix

They have been durable since 2026-07-29 and nothing pointed at them:

**`/gscratch/makelab/jonf/rampnet_yolo_baseline_51`** — klone lab storage, outside the
~21-day `/gscratch/scrubbed` purge that this record exists to survive. 1.8 GB; each arm
directory holds `best.pt`, `last.pt`, `args.yaml`, `results.csv`. A `MANIFEST.md` there
already carries the snapshot time, the job↔config map, per-arm epoch/mAP state, and
**sha256 for all 10 `.pt` files** verified against source at copy time — it simply was not
referenced from the record.

## Two caveats included rather than left implicit

- **It is a snapshot, not a live mirror**, taken 2026-07-29T20:51Z while all five arms were
  still training. They have advanced since — `y11l_pano` 13 → 15 epochs, `y26_pano` 18 → 20
  as of 2026-07-30 — so it must be re-taken before anything is quoted as final, and
  `/gscratch/scrubbed/jfroehli/yolo_runs/<arm>/results.csv` stays the live source until then.
- **`y26_tiles_l40s` is missing entirely.** That arm postdates the snapshot (Slurm 37889646,
  forked to a dedicated `gpu-l40s` node on 2026-07-29 to escape the ckpt preemption ceiling)
  and needs adding on the next pass.

Staging to Hugging Face is still worth doing for external reproducibility, and the section
says so — it is just no longer the only thing standing between these weights and the purge.

Related: #51 · #91 (Tillicum onboarding, which re-snapshots as part of the migration)

### DIFF
--- a/scripts/model_comparison/yolo_baseline/README.md
+++ b/scripts/model_comparison/yolo_baseline/README.md
@@ -214,9 +214,28 @@ protocol governs the results that will replace that note.
 
 ## Where the weights live
 
-`best.pt` files are **not** in git. Durable homes:
+`best.pt` files are **not** in git. Durable home:
 
-_TODO — stage to Hugging Face (`projectsidewalk/…`) or lab storage and record the URL here._
+**`/gscratch/makelab/jonf/rampnet_yolo_baseline_51`** on Hyak (klone) — lab storage, *not*
+`/gscratch/scrubbed`, so it sits outside the ~21-day auto-purge that motivated this record.
+1.8 GB total; each arm directory holds `best.pt`, `last.pt`, `args.yaml`, `results.csv`.
+
+`MANIFEST.md` in that directory records the snapshot time, the Slurm job↔config map, the
+per-arm epoch/mAP state at snapshot, and a **sha256 for all 10 `.pt` files**, verified
+against source at copy time.
+
+Two caveats that matter before quoting anything off it:
+
+- **It is a point-in-time snapshot taken 2026-07-29T20:51Z, while all five arms were still
+  training** — not a live mirror. The arms have advanced since (`y11l_pano` 13 → 15 epochs,
+  `y26_pano` 18 → 20, checked 2026-07-30). Re-snapshot before treating it as final; until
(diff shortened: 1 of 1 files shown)


jonfroehlich:
Closing as redundant — **PR #87 already fixes this TODO**, and more thoroughly than this branch does. I flagged it as an open gap after reading `main`, without checking that an open PR already closed it. #87 additionally documents the sha256 verification result (10 ok / 0 mismatched), the `/gscratch/scrubbed` purge rationale, and why `last.pt` is kept alongside `best.pt` (optimizer/EMA state for resume — exactly what the `y26_tiles_l40s` fork needed). Both would have edited the same lines and conflicted.

Two small deltas from this branch that #87 does not currently carry, noted here so they are not lost:

1. **The snapshot has measurably drifted.** #87 says it was staged "while the runs were still going"; the concrete drift as of 2026-07-30 is `y11l_pano` 13 → 15 epochs and `y26_pano` 18 → 20. Worth an explicit "re-snapshot before quoting as final," with `/gscratch/scrubbed/jfroehli/yolo_runs/<arm>/results.csv` named as the live source until then.
2. **`y26_tiles_l40s`'s own weights are not in the durable snapshot.** #87 mentions the fork as a *consumer* of `last.pt`, but the fork postdates the 2026-07-29T20:51Z copy, so its weights are not covered by it.

Happy to add both to #87 as a commit if wanted.

Issue #96, opened 2026-07-31
Assess location precision for the unassessed candidate cities — the gate every route to 500k depends on


jonfroehlich:
## Why this is the critical path

`docs/curb_ramp_data_sourcing.md` §6 prices the routes to a 500,000-ramp corpus and lands on
one blocking result:

| Pool | Ramps | Cumulative |
| :--- | ---: | ---: |
| **Good** (NYC + Portland + Bend) — already fully used | 276,615 | 276,615 |
| **+ every OK city** (LA + Austin + DC + Nashville) | 193,898 | **470,513** ❌ |
| + unassessed cities | ~236,000 | ~706,500 |
| + state DOTs | ~198,800 | ~905,300 |

**500,000 is not reachable on assessed data alone.** Good plus *every* OK city reaches 470,513
— ~30k short — and that is already after accepting a quality tier the paper deliberately
rejected. So **every route to 500k depends on cities whose location precision nobody has
checked**, which makes this assessment the gate, not discovery. Discovery is done (§3).

The work is cheap and unblocked: a few days of visual review, **no GPU, no pipeline run, no
sourcing commitment**. It determines whether 500k is a real target or an arithmetic one. If
roughly 60% of the ~236k unassessed pool passes at Good, 500k clears on city-inventory data
alone with no state-DOT tail.

## What "assessment" means here

Paper §3.1 / Table 1 assessed eight cities by **manually overlaying curb-ramp coordinates on
aerial imagery** and judging whether they land on the physical ramp (Fig. 2), bucketing Good /
OK / Poor. No thresholds were published. **The selection rule was "Good only"** — the corpus is
three cities because only three passed, not because only three were available.

`docs/curb_ramp_data_sourcing.md` §5 proposes quantifying it while extending it, so the tiers
stop being a judgment call:

| Check | Why it matters | Test |
| :--- | :--- | :--- |
| **Positional offset** | The Stage 1 mechanism — a misplaced coordinate produces a misplaced label. Fig. 2b is the failure | Sample ~50 points/city, measure metres from the true ramp on aerial imagery; report a **distribution**, not a bucket |
| **Per-ramp vs per-corner** | One point per *corner* collapses paired ramps to a single label — **the supervision gap behind Paterson's failure** | Records per intersection; inspect a paired corner. NYC's ~1.8/intersection implies per-ramp |
[...]

Issue #86, opened 2026-07-29
RampNet 2.0: measurement, condition inference, and tagging — we already collect the supervision and discard it at ingest


jonfroehlich:
## The thesis

RampNet 2.0 is not "find more curb ramps." It is **find + measure + rate + tag** — moving from *"is there a ramp here?"* to *"is this ramp usable, and what is wrong with it?"*

Detection is the part that is close to done. RampNet holds F1 0.908 on un-anchored gold GT and beats all eight challengers on every benchmark split by 0.19–0.34 F1; the remaining levers are incremental (#54's threshold change bought +7.7 recall for free, #78's TTA bought nothing). The marginal value of a better detector is falling, and the north star — an AI labeler that finds, tags **and rates** at least as well as humans — is three-quarters unstarted.

**And this is the part that street-level imagery uniquely owns.** #85 established that aerial orthoimagery already detects curb ramps in production (34k ramps at 1-inch GSD). What aerial structurally *cannot* do is judge slope, surface condition, obstruction, tactile warning, or whether a ramp points into traffic — all of which need a pedestrian-height viewpoint. Condition inference is therefore the durable differentiator, not a nice-to-have.

## The unique data we already have and currently discard

This is the concrete finding. Three separate streams of condition and measurement supervision already flow through this repository, and **all three are truncated to geometry at read time**:

**1. Project Sidewalk severity, tags, and descriptions — dropped on ingest.**
`stage_one/crop_model/ps_model/data/download_data.py` pulls `/v2/rawLabels` CSV exports from **12 Project Sidewalk city deployments** (blackhawk-hills, chicago, cliffside-park, columbus, knox, mendota, newberg, oradell, pittsburgh, sea, st-louis, teaneck) and consumes exactly six columns: `Label Type`, `Agree Count`, `Disagree Count`, `Panorama ID`, `Panorama X/Y`, `Panorama Width/Height` (lines 237–262). Every label's **severity rating (1–5)**, **tags**, and **free-text description** are read and thrown away.

That is the crown-jewel dataset for this problem — human accessibility judgments on curb ramps, at scale, with per-label validation counts — and the pipeline uses it purely as a source of pixel coordinates.

**2. Government attributes — reduced to three fields.**
[...]

Issue #59, opened 2026-07-25
Would 2-3M training labels buy recall? Measure the scaling curve and the label ceiling before a sourcing campaign


jonfroehlich:
## Question

Should RampNet 2.0 chase a much larger Stage 1 training set — **2–3M labels (or 10× the panoramas)** instead of today's **214,376 panos / 849,895 labels** — and would that shift performance?

This issue records the analysis, the reasons to be skeptical of *volume for volume's sake*, and the three cheap experiments that should settle it **before** anyone commits to a multi-month data-sourcing campaign.

## TL;DR

**Volume alone probably won't move much. Volume *plus* new domains *plus* better label recall probably would.** Everything we have measured says the binding constraints are (a) apparent size / input resolution, (b) the operating point, and (c) domain shift — none of which is fixed by more samples drawn from the same three GSV cities. There is also a plausible mechanism by which naive scaling makes the worst failure mode *worse* (see "The label ceiling" below).

But we have never actually plotted a data-scaling curve for this model, so this is inference from adjacent measurements, not a measurement. Experiments E1–E3 are the cheap way to convert it into one.

## What the measurements say the bottleneck is

From `docs/detection_recall_analysis.md` (#25) and the model-comparison runs (`docs/model_comparison.md`):

| Evidence | Number | Implication for "more data" |
| --- | --- | --- |
| Recall by distance | 0–8 m **.842**, 8–12 m .879, 12–18 m .812, 18–25 m **.564**, 25 m+ **.182** | Reliable to ~18 m, effectively blind past 25 m |
| Recall by apparent size | 20–32 px **.189**, 32–50 px .671, 50–80 px .825, 80 px+ **.876** | At 4096 px wide a far ramp is a few dozen pixels. **More examples do not add pixels.** |
| Share of "misses" that were sub-threshold | **~44%** | Much of the gap is *calibration*, not vision (#54: threshold 0.55 → 0.25–0.35 buys +7–10 recall pts, free) |
| In-distribution gold set (1k panos, TTA, conf 0.55) | **P .949 / R .873** | Little in-distribution headroom left to buy |
| OOD degradation (clovis, 2018 GoPro Fusion) | **P .914 / R .713** vs richmond .960/.765, bend .954/.758 | The real loss is *domain*, not sample count |
| Resolution forecast (#25) | **+0.103 recall at 2× linear res**, saturating ~0.875 | Needs a retrain, and needs far-field examples to learn from |

[...]

Issue #46, opened 2026-07-24
Qualitative failure taxonomy: bucket each model's FP/FN from cached detections


jonfroehlich:
## Motivation

The P/R/F1 table tells us *that* the challengers fail; it doesn't tell us *how*. The chat VLMs are "FP-heavy" (119–293 FP vs RampNet's 9) — but we don't know how much of that flood is genuine hallucination vs measurement artifact. Several confounds could be inflating FP (and FN) without the models being as wrong as the numbers imply:

- **Box→center + tight radius** double-penalizes loose localization: a correct-but-loose box whose center lands >0.022 from the GT point is scored as an FP *and* the ramp as an FN — one miss, counted twice.
- **Hood/nadir** (`pitch_deg=-30` → ~⅓ of every view is vehicle hood + black cap; GDINO's top box is the hood — see companion nadir-mask issue).
- **Cross-view dedup** double-counting at tile seams (#43).

## What to build

A script over `.model_cache/` that, for each model + city, classifies every FP and FN against the derived GT (`rampnet/detection_eval.py`):

- **FP buckets:** on-hood/nadir · near-GT-but-outside-match-radius (localization) · cross-view dedup duplicate · on driveway / crosswalk paint / stairs · genuine hallucination (far from anything).
- **FN buckets:** small/distant · occluded · seam-truncated (#43) · inside the hood region.
- **Output:** per-model bucket counts + a worst-cases gallery, reusing the `dump_detections.py` overlay machinery.

## Why it's cheap and high-value

- **No model re-runs** — cached detection points + derived GT already contain everything the geometric buckets need (the gallery also needs the local pano images).
- Turns "119 FPs" into "40 localization / 25 hood / 30 driveway / 24 real," which is what tells us whether the RampNet lead is as clean as it looks — and lets us qualify the paper's numbers precisely.
- **De-risks the cascade (#35):** if OWLv2's ~8,800 FPs are mostly obvious junk (hood/driveway), an arbiter kills them trivially; if they're ambiguous concrete, the arbiter struggles too. The taxonomy sets the cascade's ceiling *before* we build it.

Refs #39, #43, #35.

jonfroehlich:
Measurable half done in #94 (branch `analysis/miss-taxonomy-46`). Summary of what the buckets say,
and what is left.

**FN side.** Every miss bucketed from the committed low-floor caches (0.05 floor, well under the
[...]

PR #97, opened 2026-07-31
Assess Denver, and freeze the Stage 1 inventories for replication (#96, #59)

jonfroehlich:
First cities through the location-precision gate filed as #96, plus the frozen Stage 1 inputs the
ICCV paper should have shipped with. Analysis and tooling only — no pipeline change, no retrain, no
city sourced. Full suite **516 passed** (was 304; +212 tests).

**Denver is assessed and passes. Seattle is partially assessed and produced a methodological
correction instead of a verdict.** Four things previously asserted in this repo are contradicted
below; every correction is recorded in place rather than quietly patched.

---

## 1. Denver is **Good** — median 0.29 m, 92.3% within 1 m

All 59 chips reviewed by Jon in one sitting, against the §5e rubric. Wilson intervals throughout,
because at n≈55 with a rate near 5% the normal approximation runs below zero.

| | value | 95% CI |
| :--- | ---: | :--- |
| **Offset median** | **0.29 m** | — |
| p75 / p90 / max | 0.64 / 0.95 / 2.07 m | — |
| **Within 1 m** ("on the ramp") | **92.3%** (48/52) | 81.8–97.0 |
| Within 0.5 m | 63.5% (33/52) | 49.9–75.2 |
| **Phantom** (readable corner, no ramp) | **5.5%** (3/55) | 1.9–14.9 |
| Unjudgeable | 6.8% (4/59) | 2.7–16.2 |

The threshold is stated rather than assumed, since the paper published none: *Good = ≥90% within
1 m, phantom under 10%*. A ramp is 1.2–1.8 m deep, so "within 1 m of centre" is very nearly "on the
concrete". **This is our threshold, not the paper's** — Denver's Table 1 peers cannot be re-scored
against it without repeating the exercise.

**It also settles §5d for Denver.** Zero chips where the reviewer saw *more* ramps than Denver
publishes; 1.37 ramps seen per corner against 1.21 records per corner. Denver's low pairing ratio is
**ramp design vocabulary, not missing records** — the benign reading, now confirmed on imagery.

**Denver passing moves the usable pool from 276,615 to 349,385** — the single largest available
addition. It does not reach 500,000, and §6 shows nothing available does.

## 2. ⚠️ Denver is **mid-pack, not weak**, on per-ramp vs per-corner — I benchmarked against one city

`inventory_geometry.py`. NYC calibrates it: it publishes `rampid` **and** `cornerid`, so 6 m
single-link clustering can be scored against the publisher's own grouping — **P 0.976 / R 0.973**.
That is what licenses running it on cities with no corner key, and every remaining city gets it free
before any reviewer time.

I first called Denver "weak on pairing" by comparing it to NYC alone. **NYC is the outlier:**

| City | Records | Rec/corner (6 m) | vs published corner key |
| :--- | ---: | ---: | :--- |
| **NYC** — in training | 217,679 | **1.61** | P .976 / R .973 |
| **Portland** — in training | 46,101 | **1.38** | — |
| **Seattle** | 38,364 | **1.34** | — |
| **Bend** — in training | 14,805 | **1.31** | — |
| Sioux Falls | 19,991 | 1.22 | — |
| **Denver** | 72,770 | **1.21** | — |
| Arlington | 10,342 | 1.12 | — |
| Charlotte | 40,600 | 1.09 | P .621 / R .406 ⚠️ |
| Minneapolis | 18,453 | 1.09 | P .924 / R .859 |
| Boston | 24,022 | 1.08 | — |
| San Francisco | 50,096 | **6.64** | ⚠️ broken |

The real result is about the corpus, not the city: **all three training cities sit above every
candidate**, so any addition dilutes pairing density. And the mechanism is mostly benign — Sioux
Falls' `Diagonal`-only corners hold **1.004** records against `Directional`-only **1.365**, so a low
ratio is largely design vocabulary, which is the diversity §8 asks for.

**Never benchmark a candidate against one reference city.** That was the error.

## 3. ⚠️ Seattle's "systematic shift" is not a registration error — and the statistic needed a null

Seattle's first eleven measurable chips read median 2.33 m with a mean offset *vector* of 2.06 m
against a mean *magnitude* of 2.37 m — an **87% systematic share**. I reported that as a registration
error. The direction is real; the conclusion was not.

**The share statistic had no null.** It was documented as "~0 for noise", but under random headings
the mean vector shrinks only as `1/sqrt(n)`, so the expected share is `~0.9/sqrt(n)`:

| city | n | share | null median | p |
| **Seattle** | 11 | **87%** | 39% | **0.0013** |
| Denver | 52 | 24% | 16% | 0.221 |

Seattle's lean survives its null — and **Denver's 24% never meant anything**, being inside its own.

**A direction still doesn't say which side is wrong,** so measure the other two legs of a triangle
that must close:

```
(ramps vs imagery)  =  (ramps vs centrelines)  +  (centrelines vs imagery)
 review sheet, n=11      n≈31,000                   n≈2,500
```

**Leg 1 — `inventory_centerline_offset.py` (new).** A ramp sits half a roadway from the centreline on
one side or the other, so `median(east side) + median(west side)` = twice the shift while the unknown
half-width cancels and returns as a sanity check.

| | resultant shift | half-width E / N | samples |
| :--- | ---: | ---: | ---: |
| **Seattle** | **0.00 m** | 6.24 / 6.26 m | 31,430 / 30,809 |
| **Denver** (control) | **0.12 m** | 7.35 / 7.46 m | 49,594 / 49,404 |

Denver is the calibration and it lands: 0.12 m here against **0.10 m** from its 52 reviewer clicks,
measured independently. Flat across a 10°–30° sweep of the cardinal cutoff. **Seattle's coordinates
are unbiased against Seattle's own street network — tighter than Denver's.**

**Leg 2 — `verify_chip_georeference.py --sites-from-verdicts` (new).** Orthorectification error is
*local*, so a five-neighbourhood average cannot clear the places the verdicts came from. Measured
under the review itself, **all eleven chips clear at ≤0.32 m, mean east +0.02 m** — including the
8.79 m click, at 0.11 m.

**The triangle predicts 0.0–0.3 m and the sheet says 2.06 m. It does not close, and that is the
result.** No city-wide displacement can hide from 31,430 samples or a per-chip imagery check. The
sheet's own quartiles agree once you stop averaging: p25 is **0.47 m** against a 2.33 m median, and a
uniform shift cannot produce a chip that is dead-on. What made 87% was a handful of large
**per-record** errors sharing a heading.

Consequently: the SHIFT framing is **withdrawn**; there is **no constant to subtract**, which was the
optimistic branch; **Seattle's Poor rating is neither confirmed nor overturned**, because eleven
chips on a basemap already declared inadequate for grading settle nothing; and the
**62.9%-installed-after-2019** confound is now the leading candidate.

The durable fix: `inventory_review_summary.py` prints the null and p-value, **refuses to call a shift
on its own**, and names the two instruments that can answer it.

## 4. ⚠️ The obvious basemap is not good enough, and it fails silently

`inventory_review_sheet.py` renders the §5 positional instrument. The first version used Esri World
Imagery, which over Denver is leaf-on, hazy and effectively ~1 m, and at z=21 serves *"Map data not
yet available"* as a flat grey tile **the fetcher pasted in as evidence**. Denver's own
`Aerial2016_tilecache` is leaf-off and 0.057 m/px.

`probe_basemap.py` exists because the lesson kept recurring in new forms — King County *advertises*
maxLOD 23 and **404s above z20**, so the deepest level must be probed, never read. Every King County
year is **leaf-on at 27–41% vegetation cover** against Denver's 7.2%, which is why Seattle's sheet
carries an explicit "adequate to size a large error, not to grade a Good city" caveat.

**Every city needs its municipal basemap located and probed before its sheet is worth reviewer time.**
Charlotte's is found and is good (leaf-off, 0.061 m/px); its sheet is built and unreviewed.

## 5. ⚠️ Two disqualifications and a supply correction, found by reading data — no imagery

- **San Francisco is disqualified for Stage 1**: 50,096 rows but only **7,553 distinct coordinates**,
  1:1 with `cnn` (intersection node id) ⇒ coordinates are **intersection centroids**, ~4.7 identical
  labels per pixel. Separately, 14,414 rows are `crexist=0` — confirmed absence, wrong polarity for
  Stage 1 and right for #86.
- **Charlotte's coordinates disagree with its own corner key** (P .621 / R .406): of its multi-record
  published corners, **52.5% span >6 m, 14.4% >20 m, p99 308 m**. Its sheet must test this.
- Charlotte has **5,505 `RP_Type=NoRamp`** rows ⇒ usable 35,095, not 40,601.

**Unassessed pool is 180,673, not ~236,000 (−55,327).** So **Good (276,615) + every unassessed city,
all passing, = 457,288 < 500,000**: the route avoiding both the OK tier and the state DOTs is
**closed**.

Also fixed a silent truncation found the same way: SF names its OID field `ObjectId`, so ID-paging
stopped at 2,000 of 50,096 and wrote a normal-looking snapshot. `find_oid_field` now reads it from
the layer, and a short fetch **refuses to write** without `--allow-partial`.

## 6. ⚠️ Denver's "2022 imagery" claim does not survive contact with the data

96.2% of records are `UPDATE_STATUS=NC`, **74.4% carry a 2015 `CREATEDATE`, only 84 carry 2022**, and
`Modify` is used **zero** times across 72,770 records. Either everything was re-verified against 2022
imagery or a 2015 delineation is being carried forward; zero `M` favours the second, which is a ~7-year
gap comparable to what disqualified DC. **§5c's ✅ for Denver is now marked unconfirmed.** Nothing in
the schema records removals either, so phantoms have no upper bound from this data.

Footprint **passes**: 85.2% strictly inside Denver County, 98.8% of the rest within 1 km, 128 records
(0.18%) beyond 2 km — the mountain parks.

## 7. Frozen Stage 1 inputs — `data/inventories/`

The three paper inventories were never in this repo; the README points at live portals serving current
data. **Eleven snapshots are now frozen**, each with endpoint, exact query, fetch date,
declared-vs-retained count and a sha256.

| Snapshot | Records | vs paper Tab. 1 |
| :--- | ---: | :--- |
| `nyc-ny` | 217,679 | −0.0005% — effectively frozen |
| `portland-or` | 46,101 | **+1.7% drift** |
| `bend-or` | 14,805 | **+8.8% drift** |
| + Denver, Seattle, SF, Charlotte, Boston, Sioux Falls, Minneapolis, Arlington | | candidates |
| + `seattle-wa-centerlines` (34,484), `denver-co-centerlines` (7,866) | | **reference geometry, not supply** |

Two gzip header fields are pinned (`mtime=0`, `filename=""`) so identical records hash identically — a
real bug the test caught, since otherwise the digest tracked *where the file was written*.

Centrelines are frozen on the same terms because an analysis that exonerates a city's coordinates must
not depend on a live endpoint for the reference it exonerated them against. **Each city's centrelines
must come from the same publisher and CRS as its ramps** — Seattle's SND and curb ramps are both
EPSG:2926 from org `ZOyb2t4B0UYuYNYH` via the same server, which is what makes a shared datum fault
cancel and a ramp-layer defect show.

**Two limits stated rather than implied:** these are *not* the paper's files (Portland and Bend have
drifted, so a re-run reproduces today's dataset, not ICCV's — recovering the paper-exact files from the
supplemental stays open), and basemap imagery is not redistributable, so `verdicts.json` records tile
URLs and keys instead of shipping tiles.

## What this does *not* settle

- **Recall is structurally invisible** to this instrument — it samples published records, so it can
  only find records with no ramp, never ramps with no record. §5e says so explicitly.
- **One rater, no second pass.** The benchmark's standing top follow-up applies here too.
- **Denver's phantom interval reaches 14.9%**, which across 72,770 records is the difference between
  ~4,000 and ~10,800 false labels. Narrowing it needs a larger sample, not a better instrument.
- **Our Good/OK/Poor scale is still a scale of one.** NYC is the remaining route to placing it on the
  paper's, and its imagery service is dynamic-only — the `/export` fetcher is not built.

## Review

- `docs/curb_ramp_data_sourcing.md` §5d–§5i and §9 carry the numbers and caveats.
- `data/inventories/README.md` documents the snapshot format and how to add a city.
- Everything is regenerable; the review sheet is not committed because it embeds municipal tiles.


jonfroehlich:
**Correction: Denver is mid-pack, not weak — and the unassessed supply is 23% smaller than counted.**

Ran the geometry gate over all nine frozen inventories (pushed to PR #97). Two things I said in the
comment above are now wrong, and one of them is my error rather than new data.

## ⚠️ The Denver framing was wrong

I compared Denver against NYC alone. Against the full set, **NYC is the outlier**:

| City | Records | Rec/corner (6 m) | Singleton | vs published corner key |
| :--- | ---: | ---: | ---: | :--- |
| **NYC** — Good, in training | 217,679 | **1.61** | 0.402 | P .976 / R .973 |
| **Portland** — Good, in training | 46,101 | **1.38** | 0.635 | — |
| **Bend** — Good, in training | 14,805 | **1.31** | 0.726 | — |
| Sioux Falls | 19,991 | 1.22 | 0.785 | — |
| **Denver** | 72,770 | **1.21** | 0.794 | — |
| Arlington | 10,342 | 1.12 | 0.887 | — |
| Charlotte | 40,600 | 1.09 | 0.917 | P .621 / R .406 ⚠️ |
| Minneapolis | 18,453 | 1.09 | 0.917 | P .924 / R .859 |
| Boston | 24,022 | 1.08 | 0.924 | — |
| San Francisco | 50,096 | **6.64** | 0.038 | ⚠️ see below |

Denver is **second-highest of the candidates**. The actual result is about the corpus, not the city:
**all three training cities sit above every candidate**, so any addition dilutes pairing density
relative to what RampNet trains on today. Minneapolis's own published corner key confirms 1.09 is
real, so this is not a clustering artifact.

**And the mechanism is mostly benign.** The two cities publishing ramp type let it be decomposed:
Sioux Falls' `Diagonal`-only corners hold **1.004** records, `Directional`-only hold **1.365**;
Charlotte agrees from the other side (1.059 with a diagonal type vs 1.176 without). A low ratio is
largely **ramp-design vocabulary, not under-recording** — which is the diversity §8 asks for, and
plausibly the thing Paterson and Gainesville punished us for lacking. The counter-risk is real
though: fewer paired examples to learn the pair separation #46 found us failing.

## ⚠️ San Francisco is disqualified for Stage 1 — no imagery needed

Its 50,096 records carry only **7,553 distinct coordinates**, mapping 1:1 to `cnn` (SF's intersection
node id). Every ramp at an intersection is stamped with *the intersection's* point — modal 6–8 rows
per coordinate, up to 29. Stage 1 would project ~4.7 identical labels onto one pixel, none on a ramp.
Separately, **14,414 rows are `crexist = 0`** — confirmed absence, wrong polarity for Stage 1 (right
polarity for #86).

## ⚠️ Charlotte's coordinates disagree with Charlotte's own corner key

Recovery is P .621 / R .406, far below NYC and Minneapolis. Cause is spread: of its 3,839
multi-record published corners, **52.5% span >6 m, 14.4% >20 m, p99 = 308 m.** Big arterial corners
explain ~30 m, not 308 m. Charlotte's review sheet must test this specifically.

## Supply correction — and it closes a route

| City | Listed in §3 | Corrected | Why |
| :--- | ---: | ---: | :--- |
| **San Francisco** | 50,096 | **0** | intersection centroids |
| **Charlotte** | 40,601 | **35,095** | 5,505 rows are `RP_Type = NoRamp` |
| others | | | live drift only |
| **Total** | ~236,000 | **180,673** | **−55,327** |

**Good (276,615) + every unassessed city, all passing, = 457,288 — still short of 500,000.** The
route that avoided both the OK tier and the state DOTs is now closed: 500k requires accepting the
tier the paper rejected, or the state-DOT tail with its Richmond/NYC clipping hazards.

Both corrections came from *reading the data*, not reviewing imagery — the automated checks pay for
themselves before a reviewer is booked. Also fixed a silent fetch truncation found the same way: SF
names its OID field `ObjectId`, so paging stopped after 2,000 of 50,096 and wrote a normal-looking
snapshot. Short fetches now refuse to write.

jonfroehlich:
# Deep Code Review: PR #97

**PR**: [Assess Denver, and freeze the Stage 1 inventories for replication (#96, #59)](https://github.com/ProjectSidewalk/RampNet/pull/97)
**Branch**: `feat/location-precision-96` → `main`
**Size**: 93 files, +20,737 / −10 lines
**Author**: claude-opus-5

---

## Executive Summary

This is a massive, cohesive PR that adds a complete positional-quality analysis framework for curb-ramp inventories: 12 new analysis scripts, 14 new test files, frozen snapshots for 11 city inventories, and ~1,200 lines of documentation updates. The scientific methodology is strong (the "triangle must close" validation, half-width-cancellation trick, and permutation tests are all sound). Code quality is generally high with thorough test coverage. I found **3 critical issues**, **13 warnings**, and **11 notes**.

---

## 🔴 Critical Issues (3)

### C1. Float-precision counter is unreliable for IEEE 754 floats

**File**: [inventory_geometry.py](file:///D:/Git/RampNet/scripts/analysis/inventory_geometry.py) ~line 180–200

`_count_decimal_places` converts floats to strings and counts digits after the decimal point. IEEE 754 representation artifacts can produce spurious trailing digits (e.g., `repr(39.7654)` → `'39.765400000000004'`), inflating the reported precision to 15 decimal places. The heuristic truncation at "3+ trailing zeros" is fragile.

**Impact**: Could misreport coordinate precision in the geometry analysis. The SF disqualification finding (based on duplicate counts via exact equality) is unaffected, but the precision metric itself may be wrong for some cities.

**Fix**: Use `decimal.Decimal(str(x)).as_tuple()` to extract the true significant-digit count, or round-trip through `f"{x:.10f}"` and strip trailing zeros.

---

### C2. Centerline segment digitization direction can scramble east/west offset signs

**File**: [inventory_centerline_offset.py](file:///D:/Git/RampNet/scripts/analysis/inventory_centerline_offset.py) ~line 160–180

The signed offset (east vs. west of centerline) is determined by the cross-product of the segment direction vector and the ramp-to-segment vector. This requires consistent segment digitization direction. The code filters to near-cardinal segments (10°–30° range) and applies `_orient_segments` to normalize direction, which mitigates the issue.

**Impact**: Low in practice because the orientation normalization handles it, and the sensitivity sweep confirms robustness. But if a future city's centerline dataset has inconsistent digitization even within the cardinal window, results could be wrong without warning. The ~44% segment utilization rate (two 20° windows out of 90°) is a meaningful data reduction.

**Fix**: Add a diagnostic that counts how many segments were flipped by `_orient_segments` — if it's close to 50%, the cardinal filter isn't sufficient for that dataset.

---

### C3. Tile fetch doesn't catch HTTP 200 with HTML body (the "grey tile" problem)

**File**: [inventory_review_sheet.py](file:///D:/Git/RampNet/scripts/analysis/inventory_review_sheet.py) ~line 320–360

When a tile server returns HTTP 200 with `Content-Type: text/html` (out-of-bounds placeholder), the code attempts `Image.open()` which crashes with `PIL.UnidentifiedImageError`. This is exactly the failure mode described in the PR body for Esri World Imagery at z21 over Denver.

**Impact**: Crashes the review sheet generation for cities using tile servers that return HTML error pages instead of proper HTTP status codes.

**Fix**: Check `response.headers.get('content-type', '').startswith('image/')` before `Image.open()`, or wrap in `try/except UnidentifiedImageError` and substitute a diagnostic "no tile" placeholder.

---

## 🟡 Warnings (13)

### W1. No request timeouts across all scripts

**Files**: All scripts using `requests.get()`

No `timeout` parameter is set on any HTTP request. A hanging government endpoint will block the script indefinitely.

**Fix**: Add `timeout=30` (or a CLI-configurable value) to all `requests` calls.

---

### W2. O(n×m) brute-force nearest-centerline search

**File**: [inventory_centerline_offset.py](file:///D:/Git/RampNet/scripts/analysis/inventory_centerline_offset.py) ~line 120–150

Every ramp is checked against every centerline segment. For Denver (49,594 ramps × thousands of segments), this could be very slow.

**Fix**: Use `shapely.strtree.STRtree` for O(n log n) nearest-neighbor queries.

---

### W3. Permutation test uses random sampling when exact test is feasible

**File**: [inventory_review_summary.py](file:///D:/Git/RampNet/scripts/analysis/inventory_review_summary.py) ~line 80–120

For n=11, there are only 2^11 = 2,048 possible sign assignments. An exact permutation test is trivial and would give an exact p-value.

**Fix**: When `n ≤ 20`, enumerate all 2^n assignments instead of random sampling.

---

### W4. Tile-fetching code duplicated across 4 scripts

**Files**: `inventory_review_sheet.py`, `probe_basemap.py`, `probe_basemap_at_sites.py`, `verify_chip_georeference.py`

Tile coordinate computation and fetching logic is copy-pasted. Bug fixes in one won't propagate.

**Fix**: Extract a shared `_tiles.py` utility module.

---

### W5. Unknown verdict categories silently dropped

**File**: [inventory_review_summary.py](file:///D:/Git/RampNet/scripts/analysis/inventory_review_summary.py) ~line 300–330

`_classify_verdict` maps text labels to offsets via a hardcoded table. Unknown labels are silently counted as "unclassified" without warning.

**Fix**: Log a warning (or raise) on unknown verdict categories.

---

### W6. Basemap configs require source-code edits to add a city

**File**: [inventory_review_sheet.py](file:///D:/Git/RampNet/scripts/analysis/inventory_review_sheet.py) ~line 1000–1050

City basemap configs are hardcoded dicts. Adding a city requires editing the script. The `data/inventories/README.md` "How to add a city" section doesn't mention this step.

**Fix**: Move basemap configs to a JSON sidecar or add the step to the README.

---

### W7. Denver imagery-age concern undocumented as a tracking issue

**File**: [curb_ramp_data_sourcing.md](file:///D:/Git/RampNet/docs/curb_ramp_data_sourcing.md) §5d

Denver's "unconfirmed" imagery-age status has no linked GitHub issue. Per CLAUDE.md, in-flight questions should be discoverable.

---

### W8. §9 missing exact replication commands

**File**: [curb_ramp_data_sourcing.md](file:///D:/Git/RampNet/docs/curb_ramp_data_sourcing.md) §9

The frozen-snapshot documentation says to "re-run `fetch_inventory.py` with these parameters" but doesn't give the exact CLI commands. CLAUDE.md requires "exact commands in order."

---

### W9. ~52 MB of binary files committed without LFS

**Files**: `data/inventories/*.jsonl.gz` (11 files, up to 17.5 MB each), `analysis_out/georef_check_*/*.png` (14 files, ~1 MB each)

These permanently inflate repo clone size. Git LFS would be more appropriate, especially for the gzip files.

---

### W10. Test for `_count_decimal_places` uses only exact-representation floats

**File**: [test_inventory_geometry.py](file:///D:/Git/RampNet/tests/test_inventory_geometry.py)

Tests use values like `39.76` and `39.7654` that have exact float representations. Doesn't catch the IEEE 754 adversarial case.

---

### W11. Hub search pagination termination could be fragile

**File**: [discover_inventories.py](file:///D:/Git/RampNet/scripts/analysis/discover_inventories.py) ~line 130–140

Pagination terminates when `len(results) < num`. If the API returns fewer results for a non-terminal page (e.g., due to filtering), the loop exits early and misses subsequent pages.

---

### W12. Bearing residual doesn't filter for match quality

**File**: [stage1_bearing_residual.py](file:///D:/Git/RampNet/scripts/analysis/stage1_bearing_residual.py) ~line 120–140

Ramps matched to wrong panoramas produce meaningless bearing residuals. No filtering on match confidence.

---

### W13. Georeference verification re-downloads tiles instead of comparing against review tiles

**File**: [verify_chip_georeference.py](file:///D:/Git/RampNet/scripts/analysis/verify_chip_georeference.py) ~line 400–430

If the tile server updates imagery between the review and verification, the check could pass even if the original review used different imagery. The script doesn't compare tile URLs against those recorded in `verdicts.json`.

---

## 🔵 Notes (11)

| # | File | Note |
|---|------|------|
| N1 | `discover_inventories.py` | ArcGIS hub/group search URLs are undocumented endpoints that could change |
| N2 | `fetch_inventory.py` | `_normalize_record` strips hardcoded server-field names; new inventories may have unlisted server fields |
| N3 | `probe_basemap.py` | Entropy threshold (4.0 bits) for tile validity is undocumented |
| N4 | `probe_basemap_at_sites.py` | Bounding box radius is in degrees (physical size varies with latitude) |
| N5 | `inventory_precision_score.py` | Good/OK/Poor thresholds calibrated on 3 cities only |
| N6 | `inventory_review_sheet.py` | HTML built via string concatenation; Jinja2 would be more maintainable at 1,345 lines |
| N7 | `inventory_review_summary.py` | Bootstrap CI uses percentile method; BCa would be more reliable for n=11 |
| N8 | `data/inventories/README.md` | Verification command uses `sha256sum` but repo is Windows-primary; mention `Get-FileHash` |
| N9 | `curb_ramp_data_sourcing.md` §5e | Charlotte P/R metric needs inline clarification (self-consistency, not detection) |
| N10 | Manifests | `notes` field is present in some manifests but not others; should be standardized as optional |
| N11 | PR scope | 20,737 lines across 93 files is very large; inventory freezing could have been a separate PR |

---

## ✅ Positive Highlights

| Area | What's good |
|------|------------|
| **Scientific methodology** | The "triangle must close" validation (ramps→imagery = ramps→centrelines + centrelines→imagery) is rigorous and well-documented |
| **Half-width cancellation** | The centerline offset trick is elegant, correct, and validated by the Denver calibration |
| **SF disqualification** | Evidence trail is airtight: 50,096 → 7,553 distinct → 1:1 with intersection nodes |
| **`--allow-partial` safety net** | `fetch_inventory.py` refuses to write truncated snapshots without explicit opt-in |
| **Caveats inline with numbers** | Every rating has its limitations stated in the same paragraph, per CLAUDE.md |
| **"Refuses to call a shift"** | `inventory_review_summary.py` presents the test without making the decision — rare and correct |
| **Deterministic gzip** | Pinning `mtime=0` and `filename=""` for reproducible hashes; the test caught this |
| **Test coverage** | 14 new test files with 3,735 lines. Tests catch real edge cases (gzip determinism, bearing wraparound, permutation symmetry) |
| **Diagnostic images committed** | Registration overlays in `analysis_out/` make the georeference verification visually inspectable |
| **Sensitivity sweeps** | Cardinal cutoff sweep (5°–45°) and offset tolerance sweep (0.005–0.050) validate robustness of choices |

---

## Recommended Priority for Fixes

1. **C3** (grey-tile crash) — straightforward defensive fix, blocks reliable sheet generation
2. **C1** (float precision) — could misreport metrics; simple fix with `decimal.Decimal`
3. **W1** (request timeouts) — easy to add, prevents indefinite hangs
4. **W5** (unknown verdict warning) — one-line fix, prevents silent data loss
5. **C2** (digitization direction) — add the diagnostic counter; the existing mitigation works but needs monitoring
6. **W4** (tile code dedup) — maintenance burden grows with each new city; extract now while it's manageable

---

🤖 Generated with Google Antigravity (Claude Opus 4.6)

jonfroehlich:
Worked through this review finding by finding, in 0f7065c. **Most of it does not survive checking** — including both remaining "critical" issues, which describe functions that are not in this branch. Recording that in full, because a review nobody re-checked is indistinguishable from a review that was right.

## What was real, and is fixed

**C3 — a tile that is not an image.** Real, though not by the stated mechanism, and worse than described. `_fetch_tile` wrote the response body to the tile cache and *only then* handed it to `Image.open`. So a service answering an out-of-range tile with an HTML page at status 200 raised out of the whole sheet build **and** left the bad bytes on disk, meaning every later run failed on the cached file rather than on the network. Nothing is cached now until it has decoded; an undecodable body is retried like a 404, then raised as `TileUnreadable` (a `TileMissing` subclass, so existing callers still catch it). The caller counts it separately and the manifest records it separately — "outside the basemap footprint" is a fact about the city's records, this is a fact about the run, and folding the second into the first is exactly the silent coverage shrink that produced the 1-of-60 Charlotte sheet. Cache entries written by older runs are dropped and re-fetched instead of failing forever. Six new tests, none touching the network.

**W11 — pagination.** Real. The Hub filters after paging, so a page of 80-of-100 can be followed by a full one and a short page is not proof of the end. It stops on an empty page now, still bounded by `--pages`. Three new tests.

**W6 — basemaps need a source-code edit.** Real, and the README did not say so. `data/inventories/README.md` now has the step, with the `probe_basemap.py` call first — never read a service's declared maxLOD and believe it. `TILE_SOURCES` deliberately stays in source: each entry carries the vegetation cover, m/px and "what this can and cannot grade" note the verdicts have to be read against, and those belong next to the URL they qualify.

**W8 / N8 — replication commands.** Fair. The README now carries a verification recipe in **both** bash and PowerShell (this repo is Windows-primary), each checked against all 13 committed snapshots — 13/13 OK — plus how to recover the exact fetch command from a manifest, and a note that re-running it reproduces the fetch and not the file, because these inventories drift.

Separately, §9's second blocker is restated as the practical fact — we do not re-host the tiles.

## What does not check out

| Finding | Status |
| :--- | :--- |
| **C1** float-precision counter | **Not in this branch.** There is no `_count_decimal_places`, and no decimal-place precision metric anywhere in the PR. `grep -rn 'decimal' scripts/analysis/` returns nothing relevant |
| **C2** digitization direction scrambles offset signs | **Cannot apply.** There is no `_orient_segments` and no cross-product sign rule. `perpendicular_offset` returns the vector from the foot of the perpendicular to the point in absolute local ENU metres — reversing a segment leaves the foot, and therefore the sign, unchanged |
| **W1** no request timeouts | **Wrong.** No script here uses `requests` at all. All six network calls are `urllib.request.urlopen(req, timeout=...)`, every one with an explicit timeout |
| **W2** O(n×m) brute-force nearest-centreline | **Wrong.** `SegmentIndex` is a uniform-grid spatial hash, bucketed per axis. An STRtree would be a lateral move |
| **W3** use an exact permutation test | **Not applicable.** The null randomises *directions* — `rng.uniform(0, 2π)` — not signs. There is no 2^11 space to enumerate; that would be a different test |
| **W4** tile code duplicated across 4 scripts | **Mostly wrong.** `verify_chip_georeference.py` imports `inventory_review_sheet` and reuses its constants and `_fetch_tile`. Only `probe_basemap.py` has its own small `metres_per_pixel` |
| **W5** unknown verdicts silently dropped | **Wrong.** `classify` is total over four fields — `unjudgeable`/`phantom`/`measured`/`todo`. No label table, no unclassified bucket |
| **W10** test only uses exact-representation floats | Moot — it tests C1, which does not exist |
| **N3** undocumented 4.0-bit entropy threshold | **Not in the file.** `probe_basemap.py` measures vegetation cover and excess-green, not entropy |
| **W12** bearing residual has no match filter | **Already handled.** `match_bearings` caps separation at `max_sep_deg`, and the module docstring states the truncation and says to read `matched_frac` beside it |

## Declined, with reasons

- **W9 (Git LFS for the 53 MB of snapshots).** These are the frozen inputs every §5 number is computed from — committed is the strongest form of "retrievable by someone else". Converting them now would not shrink a clone without rewriting history, and would trade a plain `git clone` for an LFS dependency on exactly the files a replicator most needs.
- **W13 (compare tile URLs against the review's).** The concern is real but the fix proposed is not the one that would work: the tile *source* is already pinned per city with a comment saying it must match the sheet, and what is unverified is whether the *bytes* changed between review and verification. Doing that properly means a per-chip imagery hash — the `imagery_manifest.py` pattern — which is a change to the instrument, not a review fix.
- **W7 (file an issue for Denver's imagery age).** The caveat travels with the number in §5d, which is where CLAUDE.md wants it.

577 passed, 1 skipped.

PR #99, opened 2026-07-31
Far-field anomaly Phases 0+1 (#46): the pixel floor falls, and 'silent' is almost never absent

jonfroehlich:
Phases 0 and 1 of the four-phase far-field study designed on #46 (2026-07-31): before the 34/36-`visible` verdicts are allowed to mean anything, check the **sample** they were made on and the pixel-starvation **framing** against the model's own behaviour (Phase 0), then read the model's **heatmap** inside every silent miss's match window to ask whether "silent" ever meant "saw nothing" (Phase 1). Phase 0 is committed data only — no GPU, no network, no imagery. Phase 1 ran the published checkpoint (`projectsidewalk/rampnet-model`, the same weights every committed cache came from) on a local RTX 3070, single-pass fp32 matching `op_cache`.

## Findings

**1. Survivorship (H1) is real, mild, and now quantified.** The rated 37 are the ≥30-source-px slice of the 46 unwitnessed far-field silent misses, and the floor is not one floor — it is 7.5 model px on the 16384-px splits and 30 on morgantown, so the deck's composition follows stored width (27 GSV / 10 Mapillary; gainesville alone contributes 14). Net bias: AUC(rated px vs unrated far-silent px) = **0.600**, rated median at the **62nd percentile**. The 9 below-floor exclusions are the extreme tail (median 117 m, 3 above-horizon clamps). So the 94% visible rate generalizes to the far-silent **core (~18–50 m)** and says nothing about the deep tail.

**2. The model's own hits refute a hard pixel floor.** The decisive readout needs no reviewer:

| band | GT | recall | silent | rated `visible` |
| 18–25 m | 395 | 0.777 | 25 | 12 |
| 25–40 m | 226 | 0.549 | 32 | 13 |
| 40–150 m | 74 | 0.284 | 23 | 9 |
| clamp ≥ 150 m | 5 | 0.200 | 3 | 0 |
| **total** | **700** | | **83** | **34** |

For each `visible` miss, the model's recall over all far-field GT within ±20% of that miss's apparent size is **median 0.57** (q1 0.31, q3 0.74). A hard pixel floor would put these near zero. Recall declines 0.777 → 0.549 → 0.284 across the bands. **Far-field failure is graded sensitivity, not a cliff** — a silent far miss is the unlucky tail of a process that succeeds on most same-sized ramps.

**3. Guards hold.** Zero above-horizon clamps among the rated 37; the deck is majority GSV, where flat-ground distance is trustworthy (Spearman 0.95).

**4. Phase 1: `silent` was peak bookkeeping, not blindness.** Decomposing the in-window heatmap mass for all **128 pooled silent misses**: **absent** (act < 0.01) **10** — the only true "saw nothing" cases; **faint local** (0.01–0.05) **39** — a real on-site response below the floor (mass at the ramp in 30/39, nothing else within ~3.8 R), the `sub_threshold` continuum; **tail** (≥ 0.05) **79** — a *confident* adjacent mode's tail (nearest cached floor peak ~1.4 R away at median score **0.685**; argmax in the window's outer quarter in 75/79), the σ/representation family whichever of neighbour-TP / FP / self-just-outside-radius it is. The anomaly population itself — the 34 far-field rated-`visible` — splits **3 / 12 / 19**: the model responds at or next to ~91% of the far ramps a human called resolvable. And of the 45 *near-field* silent misses under the 0.013 sourcing estimate, only **6** are heatmap-absent — the rest are the calibration and σ families §0b already prices, so **0.013 stays unrevised in either direction** until Phases 2–3. The per-pano null confirms the classes rather than merely accompanying them: median percentile **0.495 absent / 0.780 faint-local / 0.915 tail**, and `absent` sitting at chance is exactly what a flat heatmap should read. (Read the percentile, not the p95 flag: with ~23 non-overlapping windows per elevation band the p95 is the band *maximum*, so only 31/128 clear it by construction.)

**The takeaway (now stated explicitly at the end of §0c):** distance is a **stressor on failure families the taxonomy already prices, not a new category of failure** — 62% σ/representation, 30% sub-threshold attenuation, 8% genuine absence. The implied lever is decoder- and representation-side (target σ, peak spacing, threshold calibration, Phase 2's scale question for the residual), **not far-field training vocabulary**; multi-view remains the one remedy that sidesteps all three mechanisms at once.

## What this changes in the docs

- New **§0c** in `docs/curb_ramp_data_sourcing.md` with the tables above and the boundaries of the claim: §0a's far/near **measurement stands**, its "fixable by" **binary falls** (the 18 m boundary marks difficulty, not reachability); the §0b sourcing bracket's exclusion of all 83 far-field silent misses is no longer safe, but **0.013 is deliberately not revised** until Phases 2–3 run (scale counterfactual, decoy control); multi-view's remedy logic is untouched, though the "MV ceiling" column shares the binary assumption.
- §0a gets an in-place qualifier pointing to §0c.
- Two stale §0b claims fixed: the reviewer pass **is** done (one rater), and the point estimate exists (0.013, `docs/replication.md`).
- Phase 3 design consequence recorded: the decoy deck should stratify by distance band and oversample 40–150 m, where `visible` went 9-of-9 down to 10.5 model px — the spot where pointed-verification bias would show most.

## Post-review changes (2026-08-18)

Reviewed on this PR; fixes in `5809126` and `4d2bdbd`, with `main` merged in `8e896ca`.

**One number moved.** `band_of` dropped far-field rows clamped to exactly 150 m while *below* the horizon — `geom()` reaches 150 m by two routes and only the above-horizon one is a `y` tell, so a half-open [40, 150) top band lost them while `y > 0.5` also kept them out of `clamp`. Two GT rows, one of them a silent miss, which is why the band table above used to sum to 698 GT / 82 silent against a population of 700 / 83. The 40–150 m row is now **74 / 0.284 / 23**; no conclusion changes, and the script now refuses to print a band table that does not sum to its population.

Everything else was already correct and is now harder to get wrong: the seam divergence between this window (which wraps at x=0/1) and the matcher's (which does not) is flagged per row — 9 of 128, none changing bucket; `null_pct` is reported rather than computed and ignored; `--limit` and a `--cities` subset can no longer overwrite a committed artifact with a partial run; the rater is a flag and travels in the result JSON; and both scripts pin their JSON output to LF, which a Windows re-run had been silently breaking.

## Mechanics

Phase 0: `scripts/analysis/farfield_forensics.py` + 41 tests (`tests/test_farfield_forensics.py`), result JSON at `analysis_out/farfield_forensics.json`. Phase 1: `scripts/analysis/silent_activation.py` + 41 tests (`tests/test_silent_activation.py`), result JSON at `analysis_out/silent_activation.json` (both whitelisted in `.gitignore` like the other result files). The tests pin the population arithmetic (83 = 37 rated + 9 below-floor + 37 witnessed), cross-check the gallery manifest's widths against the imagery manifests, and pin **every figure §0c quotes** to the two result JSONs — Phase 0's by re-running it byte-for-byte, Phase 1's by value, since it needs a GPU and the native-res panoramas. One instrument catch for the record: the first azimuth-null implementation could read a site's own faint bump back as "chance", and the null now self-excludes a 2 R zone around the site. Full suite: **1085 passed**.

Phase 1's imagery is the Hugging Face dataset `projectsidewalk/rampnet-benchmark` (`--panos-root` points at the checkout holding it); everything else both phases read is committed. Both take `--rater` (default `jonf`), so the second pass #46 keeps asking for is a flag rather than a patch.


jonfroehlich:
Scope grew from Phase 0 to Phases 0+1 in this branch — Phase 1 (activation forensics over all 128 silent misses, local GPU) landed as 59d5a9e + 64763d0. Headline: only 10/128 silent misses are heatmap-absent; 62% are a confident adjacent mode's tail (~1.4R, median score 0.685); the far-field rated-visible population splits 3 absent / 12 faint-local / 19 tail. Full numbers in the #46 comment and docs §0c. Suite 533 passed.

jonfroehlich:
## Deep code review — Phases 0+1

Read all 8 changed files and every helper they import (`miss_taxonomy`, `miss_decomposition`, `miss_gallery`, `stage1_label_recall.geom`, `threshold_sweep`, `rampnet/detection_eval`, `rampnet/metrics`); re-ran Phase 0 from the committed caches; re-derived every number in the PR body and §0c from the two committed JSONs; ran both new test files and the full suite (**533 passed**, matching the PR body); test-merged against `main`.

**Verification first, because it is the headline: I could not find a single number in the PR body or §0c that fails to reproduce.** Phase 0 re-ran identically off committed inputs. Every Phase 1 claim I checked independently — 30/39 faint with on-site mass, 1.38 R / 0.685 for the tail, 75/79 argmax in the outer quarter, 70/79 within 2 R, 3/12/19 for the far-field visible, 6-of-45 near-field absent, 9-of-9 visible down to 10.5 model px, the 247/180 split — lands exactly, as do the `act` quartile table and the class table, cell for cell. The pure-core / I-O separation with unit tests on every load-bearing function is the right architecture for this.

Findings, ordered by what I would actually fix.

---

### 1. `band_of` silently drops rows clamped to exactly 150 m below the horizon — and the published band table does not reconcile with its own section

`scripts/analysis/farfield_forensics.py:110-122`

`stage1_label_recall.geom` clamps to 150 m in **two** places:

```python
if depression <= 1e-4:                          # above horizon
    return 150.0, ...
d = min(CAM_H / math.tan(depression), 150.0)    # <-- and here
```

`band_of` recognizes only the first (`row["y"] <= 0.5`), and its last band is half-open `[40, 150)`. A row *below* the horizon but far enough that `min(..., 150.0)` bites therefore returns `None` — no band at all. Confirmed against the committed caches:

```
far GT 700 | rows in NO band: 2
   richmond    y=0.501962 dist=150.0 px=5.22 hit=False bucket=sub_threshold
   annapolis   y=0.502979 dist=150.0 px=5.22 hit=False bucket=silent
```

**Consequence in the doc.** §0c's band table sums to 698 GT and **82 silent misses**, while the survivorship table two paragraphs above states **83**. The section contradicts itself, and the band table is the one the PR body calls the decisive readout. Corrected, the bottom row is **GT 73 / recall 0.288 / silent 23**.

Nothing else moves: the AUCs, the matched-size rate, the survivorship quartiles and the guards all run off `far` / `far_silent` directly, not off the bands.

### 2. The tests pin the population arithmetic but not the band partition — which is why 1 got through

`tests/test_farfield_forensics.py:76-89`

`test_band_edges_are_half_open` checks 18.0, 25.0 and 40.0 and never the 150 upper edge; `test_near_field_rows_fall_in_no_band` establishes `None` as a legitimate return, so the orphaned rows look like intended behaviour. The file's own docstring says it exists to pin "the population arithmetic the write-up quotes" — the bands are part of that arithmetic and are not covered. Wanted: a `band_of(dist=150.0, y=0.7)` case, and an integration assert that the bands sum to `len(far)` and their silent counts sum to 83.

### 3. "the window is exactly the matcher's" is not true at the 360° seam

`scripts/analysis/silent_activation.py:17-19, 66-73, 108-126`; §0c ("the window is exactly the matcher's")

`site_profile` wraps columns (`col % W`) and `nearest_peak` wraps (`dx = min(dx, W - dx)`). `rampnet.metrics.greedy_match` — which produced the `hit`/`silent` labels — does **not**; it is a plain `(pred_x - gt_x)**2`. So for a miss near x=0 or x=1 the two read different windows. **9 of 128 misses sit within the match radius of the seam** (bend ×2, gainesville ×2, morgantown ×3, paterson, richmond), with `act` up to 0.946; for those rows roughly half the scanned window is imagery the matcher never considered.

Severity, to be clear: **no classification flips.** The nearest cached floor peak is ≥ 23.4 px in all 9, so none would have been `merged`. (I also chased the one row whose reported `nearest_peak_px` is 22.5 against a 22.528 px radius — annapolis at x=0.078, not a seam case; the exact value is 22.532157, correctly outside. Display rounding only.)

The wrap is the *physically right* geometry for an equirectangular panorama — the matcher is the one that is wrong at the seam, and that is pre-existing. So keep the wrap and drop the parity claim, plus a `seam` flag on the affected rows so a future population can be re-checked cheaply.

### 4. The null is built as the instrument, then never used by the conclusion — and it points the other way for the class that needs it most

`scripts/analysis/silent_activation.py:129-156, 294-318`

The module docstring makes the null the basis of the claim: "`act` is reported as a percentile of its own panorama's null, so 'there is signal here' is a claim against that pano's actual noise floor, not against zero." The headline decomposition then uses raw cutoffs (0.01 / 0.05) and consults the null nowhere. Measured on the committed JSON:

| class | n | above own null p95 |
| :--- | ---: | ---: |
| absent | 10 | **0** |
| faint_local | 39 | **2** |
| tail | 79 | 29 |

`faint_local` is described as "a real on-site response below the floor" — and **37 of 39 fail the script's own significance test**. §0c discloses the pooled 31/128 in one line and calls the bar deliberately hard, which is honest, but it leaves the tension sitting exactly where a reader will push.

**The data already resolves it, for free.** `null_pct` is computed and stored per row and reported nowhere:

| class | null_pct q1 / med / q3 |
| :--- | :--- |
| absent | 0.445 / **0.495** / 0.650 |
| faint_local | 0.600 / **0.780** / 0.855 |
| tail | 0.860 / **0.915** / 0.990 |

Clean monotone separation, with `absent` sitting at chance (0.495) as it should. That is the number that supports the decomposition.

Worth stating alongside it *why* p95 is such a hard bar, because it is not obvious: median `null_p95` is **0.595** against a median `null_med` of **0.003**. With a 22.5 px radius on a 1024-wide grid there are only ~23 non-overlapping windows per elevation band, so the 95th percentile of 200 draws is effectively the band maximum — the test asks "is this site the strongest thing on its horizon row", not "is it above the noise floor". A compounding factor goes unmentioned: the null draws can land on *other GT ramps* in the same band (only the site's own 2 R zone is excluded), so real ramp responses inflate the null.

None of this changes a number. It changes whether the interpretation survives contact with a skeptical reader.

### 5. `--limit` can silently overwrite the committed artifact with a partial run

`scripts/analysis/silent_activation.py:182-184, 255-257, 329-337`

`--limit` breaks out of the pano loop, but the payload records only `n` and `skipped_no_imagery` — nothing marks the run as truncated, and the flag's help text (`"Stop after this many panos (smoke test)"`) actively invites the combination. Since `analysis_out/silent_activation.json` is a committed artifact of record and the file every §0c number derives from, `--limit 3 --json-out analysis_out/silent_activation.json` produces a file that looks complete and is not. `--cities` is the same vector.

### 6. Single-rater filename is hardcoded, in both scripts and both test files

`scripts/analysis/farfield_forensics.py:178` — `os.path.join(gallery_dir, "silent__jonf.json")`

CLAUDE.md: *"When a task involves human raters, assume from the start that a second rater will repeat it and the two will be compared."* A second rater's verdict file cannot be run through either phase without editing source, and the same filename is baked into `tests/test_farfield_forensics.py:139`. A `--verdicts` arg costs nothing today.

Related latent trap in the same function: an item in the gallery manifest but missing from the verdicts file gets `verdict: None` and still counts toward `rated_rows` and the per-split table. All 50 are rated today, so nothing is wrong; mid-pass with a second rater the counts would inflate silently.

### 7. Phase 1's input is not named anywhere — the clean-clone test fails on it

§0c is careful to say Phase 0 is "Committed inputs only", which throws the contrast onto Phase 1: it needs `benchmark/<city>/panos`, which is git-ignored. Neither the script docstring nor §0c names where those come from. The replication rule is explicit that an unstated gap is indistinguishable from a hidden one, and this one has a published answer.

### 8. Smaller things

- **Selective agreement with E1.** §0c: "the pooled 25–40 m rate agrees with E1's gold-set 0.49 at the same range" (0.549 vs 0.49). The adjacent band, which `FAR_BANDS`' own comment names as equally comparable, is **0.777 pooled vs 0.90 gold** — a 12-point gap, unmentioned. It is explainable (the gold set is in-distribution; the pooled seven are OOD), but citing only the band that agrees reads as selective.
- **`rated visible` column mixes near and far.** The class table's column reads 5 / 13 / 23 (= 41, all fields) while the bullet directly below quotes 3 / 12 / 19 (far only). Both correct, adjacent, and a reader will trip.
- **`scripts/analysis/README.md` run-order table** does not get the two new scripts. Pre-existing drift (5 others are missing too), but Phase 1 is one of the few GPU-requiring scripts and that table is where that is documented.
- **An observation, not a defect:** `absent` is 5 rated `visible` out of 10. Half the genuinely flat-heatmap sites are ones a human called resolvable — the cleanest "human sees it, model has no response at all" cases in the study, and the sharpest target for Phase 2's scale counterfactual. Currently only a table cell.

**Deliberately not changing**, recorded so the decision is not relitigated:

- **`null_percentile` recomputes the site window** (`radius_max` inside it, `site_profile` again at the call site). Measured at ~21 s across all 128 misses against ~10 s/pano inference — not worth a signature change.
- **Shared RNG across misses.** `random.Random(NULL_SEED)` is consumed in `sorted(by_pano.items())` order, so per-row `null_pct` is reproducible only for the full seven-city run. Consistent with `miss_taxonomy.null_in_radius`; a per-site seed would be marginally better and would invalidate the committed artifact for no change in the numbers.

### Merge state

`mergeStateStatus: DIRTY`. Test-merged: **`docs/curb_ramp_data_sourcing.md` auto-merges cleanly** (§0c lands at line 342, between §0b and §1, ahead of all of main's new §5d–§5o). The only conflict is `.gitignore`, where both sides appended whitelist entries — resolution is "keep both".

### Verdict

The science is sound and the arithmetic holds. **Finding 1 is the only one I would call blocking**, and only because the published table contradicts its own section by one miss. Findings 3, 4 and 7 are doc and reporting changes that make the claims survive a hostile read; 4 is the one to prioritize after 1, because the null result as currently reported looks like it undercuts the faint-local claim when the `null_pct` already computed clearly supports it.

Fixes for all of the above are going up on this branch next.

jonfroehlich:
## Review addressed — `5809126`, plus a merge of `main` in `8e896ca`

All eight findings applied, majors first. CI green on 3.10 and 3.12; suite is **1049 passed / 1 skipped** on the merged tree (**544** on the pre-merge branch, up from 533 — +4 farfield, +7 silent_activation).

**One correction to my own review first.** I wrote that the fixed 40–150 m row would be "GT 73 / recall 0.288 / silent 23". It is **74 / 0.284 / 23** — *both* orphan rows land in that band and I carried only one. The band table below is the measured output, not my arithmetic.

### 1. `band_of` — the one real bug

Top band closed at its upper edge. `geom()` reaches 150 m by two routes and only the above-horizon one is a `y` tell, so `min(d, 150.0)` rows with `y > 0.5` fell through every band. Corrected table, now with a totals row:

| band | GT | recall | silent | rated | rated `visible` |
| :--- | ---: | ---: | ---: | ---: | ---: |
| 18–25 m | 395 | 0.777 | 25 | 14 | 12 |
| 25–40 m | 226 | 0.549 | 32 | 14 | 13 |
| **40–150 m** | **74** | **0.284** | **23** | 9 | 9 |
| clamp ≥ 150 m | 5 | 0.200 | 3 | 0 | 0 |
| **total** | **700** | | **83** | **37** | **34** |

The 0.777 → 0.549 → **0.284** sequence and "roughly 3 in 10" both survive, so no conclusion moves.

⚠️ **The table in the PR description above still shows the old 72 / 0.292 / 22 row.** I did not edit your PR body — say the word and I will.

### 2. The invariant that would have caught it

The script now **refuses to print a band table that does not sum to its population** (`SystemExit`, not a warning), and prints the reconciliation line under the table. Pinned by `test_the_bands_partition_the_far_field`, plus a `band_of(150.0, y=0.502)` unit case — `test_band_edges_are_half_open` had only ever tested the lower edges.

### 3. Seam

Kept the wrap, dropped the parity claim. Every row now carries a `seam` flag, and the script reports both counts. Measured: **9 of 128 seam rows, 0 that would change bucket** (minimum nearest floor peak among them is 23.4 px against a 22.5 px radius). No number moves; the flag makes that a query next time rather than a re-derivation.

### 4. The null

`null_pct` is now a column in the printed table and in §0c:

| class | n | null pct q1 / med / q3 | > own p95 |
| :--- | ---: | :---: | ---: |
| absent | 10 | 0.445 / **0.495** / 0.650 | 0/10 |
| faint local | 39 | 0.600 / **0.780** / 0.855 | 2/39 |
| tail | 79 | 0.860 / **0.915** / 0.990 | 29/79 |

`absent` landing at 0.495 is chance, which is exactly what a flat heatmap should read — so the null now *confirms* the class boundaries instead of appearing to contradict them. §0c and the docstring both explain why the p95 column is low and should not be read as the noise floor (median `null_p95` 0.595 against median `null_med` 0.003 — with ~23 non-overlapping windows per elevation band the p95 is the band maximum), and note the second conservatism: a draw can land on another GT ramp in the same band.

### 5. Partial runs

`--limit` with `--json-out` is refused outright; a `--cities` subset needs an explicit `--allow-partial`. `build_payload` is one function now, and records `cities` and `panos`, so the file states its own scope. Two tests assert the refusals.

### 6. Second rater

`--rater` on both scripts (default `jonf`) via `verdicts_path`; the rater is recorded in the Phase 0 JSON. Tests no longer hardcode the filename. Also fixed the related trap: a manifest item with **no** verdict is no longer returned as rated — harmless today at 50/50, but mid-pass it would have inflated `rated_rows`, the per-split table and the survivorship AUC with no visible symptom.

### 7 & 8. Record

Phase 1's imagery input named as `projectsidewalk/rampnet-benchmark` in the docstring and §0c; both E1 bands stated rather than only the agreeing one (25–40 m 0.549 vs 0.49, **and** 18–25 m 0.777 vs 0.90, with the in-distribution reason); the class table's `visible` column disambiguated; both scripts added to the analysis README run-order table; and the `absent` cell's 5-of-10 `visible` written up as Phase 2's sharpest target.

### Found while regenerating

Both scripts wrote their result JSON in Python text mode, so a **Windows re-run produced CRLF** and could not be byte-compared against the committed LF artifact — a content hash that only holds on one OS. Pinned to `newline=""`. Phase 0's artifact now reproduces byte-for-byte on a re-run (verified on the merged tree).

### Provenance of the two artifacts

- **Phase 0 re-ran in full** from the committed caches. Diff is 3 corrected band values + the `rater` key.
- **Phase 1 was NOT re-run** — no local panoramas, no GPU. Its JSON gains only *derived* keys: `seam` from `x` via the shipped pure function, `cities` and `panos` from the rows. The regeneration asserted every pre-existing key and value byte-identical before writing, and the emitted `seam` values were re-checked against a fresh derivation. No measured quantity changed.

### Merge

`main` merged in (`8e896ca`). Only `.gitignore` conflicted — both sides had appended whitelist entries, so both are kept. `docs/curb_ramp_data_sourcing.md` and `scripts/analysis/README.md` auto-merged; §0c sits at line 342 between §0b and §1. **`mergeable: MERGEABLE`.**

jonfroehlich:
## Test coverage — `4d2bdbd`. PR description updated.

**82 tests across the two files, up from 35.** CI green on 3.10 and 3.12; suite **1085 passed / 1 skipped**; `mergeStateStatus: CLEAN`.

### Two refactors first, because the most important behaviours were structurally untestable

- **`null_azimuths`** split out of `null_percentile`. The self-exclusion is the whole point of the null — a draw whose window overlaps the site would read the site's own bump back as "chance", flattening exactly the sparse-heatmap case Phase 1 exists to detect — and **nothing tested it**. Rejection resampling is preserved exactly, so the RNG stream is unchanged and the committed nulls still reproduce (asserted against a copy of the pre-refactor loop for three values of `x`).
- **`partition_check`** split out of `main`, so the band-table guard's failure path can be exercised.

### What is now covered

**Phase 0 (41).** `partition_check` accepting and rejecting — a dropped GT row and a dropped silent miss *independently*, since the real bug lost one of each and a table can drop a miss while its GT column still looks right. `row_key` rounding and city-scoping. `stored_widths` degrading to `{}` rather than inventing a width. `load_rated`'s field filter partitioning the 50 into 37 + 13. And **`main` reproducing the committed result JSON byte for byte** — byte-level, not value-level, because the writer pins LF and a platform quietly writing CRLF would still pass a value comparison.

**Phase 1 (41).** The self-exclusion three ways: no draw lands in the zone, the zone wraps the seam, and — behaviourally — a lone bump at the site never appears in its own null. Remaining geometry: bottom-row clamp, negative values reading as no response, `nearest_peak` choosing the closest of several, and that rows scale by 512 while columns scale by 1024 (getting that backwards would silently halve or double every "nearest floor peak" in the write-up). Plus: **every derived field regenerates from the measured ones**, which is what licenses adding `seam` without re-running the model.

**Both:** every figure §0c quotes is pinned to its result JSON — the band table and its sums, both AUCs, the matched-size quartiles, the tier and above-horizon guards, the per-split floors, the 10/39/79 decomposition, the null percentile medians, the tail's 75/79 and 70/79, faint-local's 30/39, the far-visible 3/12/19, the near-field 6, the strata partition, and the seam count.

### Mutation-checked rather than assumed

| mutation | tests that fail |
| :--- | ---: |
| top band back to half-open `[40, 150)` | 3 |
| `ABSENT_MAX` 0.01 → 0.02 | 4 |
| drop the wrap from `seam_of` | 2 |
| zero the null's exclusion zone | 2 |

The last row is the one that matters most: that behaviour had shipped with no test at all.

### On re-running Phase 1 — not needed, and it would muddy the record

Nothing in the compute path changed. `class_of` extracts the same two cutoffs (pinned), `seam_of` is a pure function of `x` (pinned against every stored value), `build_payload` writes only the header, and the CLI guards run before inference. The one change that *could* have touched results — `load_rated` dropping unrated items — is inert at 50/50 rated, and a test now asserts that. Meanwhile the model runs on a GPU, so a re-run may not be byte-identical for reasons unrelated to this review, and we would be explaining that instead of the science. An end-to-end Phase 1 replication is worth doing once, deliberately, as its own experiment — not as a side effect of a review.

### PR description

Updated: the band table carries the corrected 40–150 m row (74 / 0.284 / 23) plus the clamp and totals rows, the null percentiles are in finding 4, the Mechanics paragraph reflects 41+41 tests and 1085 passing and names the imagery source and `--rater`, and there is a short post-review section. Original authorship line left as it was.

Issue #18, opened 2026-07-03
Stage 1 dataset evaluator ignores redundant points; re-quantify agreement precision


jonfroehlich:
Follow-up from #9 (this part was intentionally split out).

`stage_one/dataset_evaluation/evaluate.py` matches one-to-one on the TP side (claims one GT, then breaks — correct), but routes redundant generated points near an already-matched GT to an "ignored" counter instead of counting them as false positives. This makes the published Stage 1 agreement precision (94.03%) slightly optimistic; recall (92.45%) is unaffected.

Fix: count those redundant points as FPs (align with `rampnet/metrics.py` semantics), re-run against `manual_labels/`, and record the delta. Expected to be small — redundant generated points within 90 px of the same ramp are rarer for Stage 1 output than for model peaks — but it should be quantified rather than assumed.

Context: README erratum and `docs/eval_protocol_verification.html`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TAaPQDYhNDmYmxvrz4Lgm7

jonfroehlich:
Closing note for whoever picks up the re-measurement: **the corrected number will move for two reasons, not one.**

#33 landed the code fix but deliberately did not re-run the number (it needs the generated Stage 1 dataset, which isn't in git — `download_dataset.py`, or run it on makelab2). In the review round it also did something this issue didn't anticipate: rather than adding a third matcher, it unified Stage 1 onto `rampnet.metrics.greedy_match`. So `stage_one/dataset_evaluation/evaluate.py` now differs from the published script in **two** ways:

1. **Redundant points count as FPs** (this issue). Lowers precision.
2. **Nearest-unclaimed matching instead of first-in-list-order.** These differ whenever a generated point is in range of more than one GT ramp — common: ~41% of gold-set points have a neighbour within 2× the matching radius. This can *raise* precision, by assigning points to ramps more sensibly and leaving fewer false positives.

The two pull in opposite directions, so the net delta from 0.9403 shouldn't be attributed to redundancy alone. Recall (0.9245) is unaffected by (1) but **is** affected by (2), contrary to what this issue originally said — nearest-unclaimed changes which GT points get claimed, so the TP count itself can shift.

README (erratum + the Stage 1 evaluation section) now says this too, in #42.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DcegCW7sp1ZaSnR916r7QN

PR #104, opened 2026-08-03
Measure Stage 2 training cost (#59, #84): 3.5 h/epoch not 36, and the paper run was ~12 epochs

### LINKED ISSUE #84: Stage 2 trains for 1 epoch at constant LR (~9.4k steps): measure the epoch curve, and with it the auto-label ceiling

jonfroehlich:
Corrects a ~10× error in `docs/curb_ramp_data_sourcing.md` §7, and rescues the only surviving copy of RampNet 1.0's training telemetry into the repo. Prompted by a planning question — *how long would retraining take if we hit the 500k target?* — which §7 could not answer honestly, because it had no measurement behind it.

**Headline: one Stage 2 epoch is 3.49 h on 16 GPUs (~56 GPU-h), not the ≳36 h / ≳580 GPU-h §7 estimated.** That estimate read the README's *"will take a very long time (> 24 hours)"* as a per-epoch figure. The ">24 h" in fact describes the paper's **whole ~12-epoch, preemption-riddled run** — 44.7 h of active compute over 74.6 h of calendar time.

## What this adds

The paper run's TensorBoard events survived at `/gscratch/makelab/jsomeara/RampNet/stage_two/runs/experiment_1` — another student's scratch directory, one `rm -rf` from gone, and the only copy. **All 18 files are committed here** under `docs/data/rampnet1_stage2_run/` (6.4 MB) with a `SHA256SUMS` manifest and a `.gitattributes` binary guard, so the measurement is replicable from a clean clone with no cluster access.

`scripts/analysis/stage2_train_cost.py` parses the TFRecord and protobuf framing directly — **no TensorFlow, no TensorBoard, no GPU, no network**:

| | |
| :--- | ---: |
| Median step time (rank 0) | **1.341 s** (p25 1.339 / p75 1.345, n = 119,902) |
| Global batch | 16 panoramas (16 GPUs × batch 1) |
| Steps per epoch | **9,378** (150,063 train panos ÷ 16) |
| **Wall-clock per epoch** | **3.49 h** (~56 GPU-h) |

The interquartile band is **6 ms wide across 119,902 samples** — the signature of a hard **I/O bound**, not a compute bound (~7.7 TFLOPs/pano against an A100 is roughly **3% MFU**; the cost is 8.4 MP JPEG decode + resize on ~3 cores per rank). Two consequences: cost scales with **panoramas, not labels**, which is what makes the projection below meaningful; and it is **fixable** — pre-resized panos or more dataloader CPU could cut it several-fold. Nothing here is a floor imposed by the model.

## Two things the record did not have

**The run went ~12 epochs, not 1.** Max step 112,434 = 11.99 × 9,378, cancelled mid-epoch-12 across 15 preemptions on `ckpt-all` (**×1.67** calendar overhead). Validation scalars land on exact multiples of 9,378, so the epoch length is confirmed independently of the step-rate arithmetic.

**This is half of #84's epoch curve, already run:**

| epoch | 1 | 2 | 3 | 4 | **5** | 6 | 7 | 8 | 9 | 10 | 11 |
| :--- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| auto-label val loss | .000520 | .000478 | .000463 | .000466 | **.000458** | .000468 | .000470 | .000473 | .000484 | .000487 | .000487 |

Auto-label val loss bottoms at **epoch 5 (−12% vs epoch 1)**, then rises monotonically through epoch 11. So at 1 epoch the model is **not** converged — but the headroom is small and it reverses.

**The released model is nonetheless epoch 1.** `best_model.pth` is **byte-identical** (`cmp`) to `checkpoints/epoch_1_step_9378.pth`, copied back by hand on 2025-06-21 over the val-best file the script had saved automatically. The documented "1 epoch" recipe describes the released artifact **correctly**; what it omits is that 11 further epochs were run and discarded.

## Projection to the 500k target (#59, #96)

**"500k" counts government ramp records, not labels** — RampNet 1.0 is 278,544 records → 214,376 panos carrying **849,895 point labels** (~3 views per ramp). 500k records is **1.80×** the records and implies ~1.5M labels.

| Schedule | Compute (16 GPUs) | Calendar on preemptible `ckpt` |
| :--- | ---: | ---: |
| 1 epoch (the released recipe) | **6.3 h** (~100 GPU-h) | ~10.5 h |
| 5 epochs (1.0's auto-val optimum) | 31.4 h | ~2.2 d |
| 12 epochs (what 1.0 actually ran) | 75.2 h | ~5.2 d |

**Stage 2 retraining is an overnight job, and the epoch count is worth ~12× what the data doubling is worth.** §7's "about a week" stands as an order of magnitude — but it is essentially **all Stage 1** (~170k new panos × 32 tiles ≈ 5.5M tile requests, rate-limit bound, still unmeasured).

## Replication

```bash
python scripts/analysis/stage2_train_cost.py                  # the measurement
python scripts/analysis/stage2_train_cost.py --records 500000 # the projection
python scripts/analysis/stage2_train_cost.py --verify         # re-check SHA256SUMS
```

## What this does not claim, and what is missing

- **Does not close #84.** The *human* half of the epoch curve is still unrun — evaluating epochs 1/3/5/8 against `manual_gold`. If auto-label val loss keeps improving while human F1 does not, that is the Stage 1 label ceiling. **The epoch-N weights are gone** (a 2025-07-11 cleanup left only epoch 1), so that half needs a re-run — ~28 h of compute on the 1.0 data at the measured rate.
- **The code that produced the run is not in git.** The public history begins with a squashed "Initial Commit" dated 2025-07-15, a month after the run, already carrying `num_epochs = 1`. The run-time value was ≥ 12 and is unrecoverable.
- **The Slurm record is incomplete.** `sacct -u jsomeara -S 2025-06-16 -E 2025-06-22` shows only three `train_curb_ramp_detector` jobs of 2–3 h each, because `--requeue` overwrites the accounting record; the event files are the only evidence of the other 12 segments. The event-file hostnames (g3050, g3060, g3082) are what tie the logs to those 4-node jobs.
- **The hardware claim is looser than the README's.** README says 16× L40s; the surviving Slurm records show those incarnations holding **A100s**, under `--constraint='l40s|l40|a40|a100'`. Immaterial while I/O-bound (1.34–1.53 s/step across segments on mixed hardware), but the paper's phrasing is more specific than the evidence supports.
- **The projection carries 1.0's ratios.** 0.770 panos/record and the 70.0% train share will shift with city density; and at a fixed 1 epoch, 1.8× the data silently becomes a 1.8× longer schedule — a change in training regime, not only in cost.

No code paths change; `docs/` and `scripts/analysis/` only.


jonfroehlich:
## Deep Code Review: PR #104

### Overview & Summary
This PR provides critical empirical ground truth for Stage 1 dataset generation and Stage 2 model training costs, replacing previous order-of-magnitude estimates with measurements derived directly from rescued training telemetry.

Key insights verified:
1. **Stage 2 Training Cost**: Measured at **3.49 h/epoch on 16 GPUs (~56 GPU-h)** (1.341 s/step median), correcting the prior estimate of ≳36 h/epoch (~10× overestimate caused by conflating the total multi-epoch preemption run with a single epoch).
2. **Multi-Epoch Training History**: The paper run actually executed **11.99 epochs (112,434 steps)** with validation loss bottoming out at **epoch 5 (0.000458, -12% vs epoch 1)** before overfitting through epoch 11. `best_model.pth` was verified byte-identical to `epoch_1_step_9378.pth`.
3. **Stage 1 Generation Yield**: **97.91% yield** (214,599 written out of 219,170 intended), with 4,570 of the 4,571 missing panoramas directly attributed to Google Street View tile fetch rejections. Disk quota errors (`[Errno 122]`) were verified to have caused zero data loss.
4. **Stage 1 Precision Upper Bound**: Accounting for 119 redundant detections as false positives adjusts precision from **0.9403 to 0.9121 (-2.8 pts)** on the gold set, providing the redundant-detection bound for #18.

---

### Critical Issue / Blocker

#### 🔴 `download_dataset.out` is gitignored and missing from the repo, causing `stage1_yield.py` to crash
- **Symptom**: Running `python scripts/analysis/stage1_yield.py` immediately fails with:
  ```
  FileNotFoundError: [Errno 2] No such file or directory: '.../docs/data/rampnet1_stage1_run/download_dataset.out'
  ```
- **Root Cause**: `.gitignore` line 3 contains `*.out`. Even though `download_dataset.out` is listed in `docs/data/rampnet1_stage1_run/SHA256SUMS` and documented in `docs/stage1_generation_cost.md`, git ignored the file when adding `docs/data/rampnet1_stage1_run/`.
- **Remediation**:
  1. Add an exception in `.gitignore`:
     ```gitignore
     !docs/data/rampnet1_stage1_run/*.out
     ```
  2. Track and commit `docs/data/rampnet1_stage1_run/download_dataset.out` (matching hash `d0956c263b90aaae9bcaccb5a06ac5799d3e21e14d1da1a9c28d252b52c86a6d`).
  3. Ensure `.gitattributes` flags it properly if line-ending normalization is needed.

---

### Code Quality & Architectural Assessment

#### 1. Pure Python Protobuf Parser in `scripts/analysis/stage2_train_cost.py` (🌟 Excellent)
- **Zero Heavy Dependencies**: Directly parsing TFRecord frames and wire format varints/fields (`_varint`, `_fields`, `_records`, `read_scalars`) with `struct.unpack` avoids requiring `tensorflow` or `tensorboard` dependencies.
- **Robustness**: Handles truncated records gracefully (when jobs were killed mid-write) and accurately extracts summary scalars.
- **Verification**: `python scripts/analysis/stage2_train_cost.py --verify` and `--records 500000` run quickly, cleanly, and deterministically. All 18 SHA256 checksums match `docs/data/rampnet1_stage2_run/SHA256SUMS`.

#### 2. Test Coverage & CI Alignment
- The existing pytest suite (501 tests) passes.
- **Recommendation**: In line with other analysis scripts in `scripts/analysis/` (e.g. `test_stage1_label_recall.py`, `test_miss_taxonomy.py`, `test_temporal_gap.py`), add unit tests under `tests/` for:
  - `tests/test_stage2_train_cost.py`: Test protobuf decoding, TFRecord framing, step gap filtering, and metric calculation against committed fixtures or synthetic messages.
  - `tests/test_stage1_yield.py`: Test log parsing, progress tracking, and missing pano calculation.

#### 3. Graceful Error Handling & CLI Validation
- In `scripts/analysis/stage1_yield.py`: Add an explicit check for the presence of required input files with a descriptive error message explaining how to obtain or verify them before calling `load_log(...)`.
- In `scripts/analysis/stage2_train_cost.py`: Consider guarding against empty step gap lists when processing directories without `Loss/train_step` scalars.

---

### Documentation & Scientific Rigor Review

- **Documentation Quality**: Outstanding clarity in `docs/stage2_training_cost.md`, `docs/stage1_generation_cost.md`, and the updates to `docs/curb_ramp_data_sourcing.md` §7.
- **Integrity of Provenance & Caveats**:
  - Explicitly documents known gaps (e.g., Slurm `--requeue` accounting loss, hardware discrepancies where surviving jobs ran on A100s rather than the README's L40s, lost intermediate checkpoints).
  - Accurately distinguishes between government ramp records (~278k) vs panoramas (~214k) vs point labels (~850k).
  - Explicitly states that the 0.9121 precision number is an upper bound (only accounting for redundant detections, not nearest-neighbor metric matching).

---

### Recommendation

**Approve with minor fixes**: Once `.gitignore` is updated to unignore `docs/data/rampnet1_stage1_run/download_dataset.out` and the file is committed so that `stage1_yield.py` executes successfully out of the box, this PR is ready to merge.

🤖 Generated with Google Antigravity (Gemini 3.7 Flash)

jonfroehlich:
Blocker fixed and both follow-ups done, in 13c10c4.

**The gitignored artifact.** Confirmed exactly as reported: `*.out` on line 3 of `.gitignore` swallowed `docs/data/rampnet1_stage1_run/download_dataset.out`, so the file was checksummed in `SHA256SUMS`, quoted in `docs/stage1_generation_cost.md`, and read by default by `stage1_yield.py` — while not being in the repo at all. It is tracked now through a negation, and the committed blob is byte-identical to the manifest entry (`d0956c26…`), verified out of the git object rather than off disk.

**One thing the review did not catch, in the same area.** The other three files in that directory were committed without a line-ending attribute. They verify here only because this machine runs `core.autocrlf=input`; on a default Windows clone (`autocrlf=true`) git would have checked them out with CRLF and **every entry in `SHA256SUMS` would have failed for files that are in fact byte-correct** — the reproducibility check reporting corruption that is not there. `.gitattributes` now marks the whole stage 1 evidence directory `-text`. All four blobs re-verified against the manifest.

**Graceful failure.** `stage1_yield.py` names the missing files and how to verify them; `stage2_train_cost.py` says which scalar tag it looked for. Neither divides by a count that can be zero any more (no missing panos, a single uninterrupted training segment, quartiles from one gap).

**Tests: 26 new, 527 passed.** The hand-written TFRecord/protobuf reader is the part worth protecting — a wire-format slip would not raise, it would quietly yield a different `s/step`, and every cost figure in both docs is derived from that one number. So the tests encode events themselves rather than depending on tensorflow, cover the framing edge cases (truncated tail from a preempted job, header-only file, preemption gaps excluded from the median), and then re-derive the published numbers from the committed run: **1.341 s/step, 3.49 h/epoch, 11.99 epochs, val bottoming at epoch 5, 97.91% Stage 1 yield with 4,570 of 4,571 losses Google's**. If the rescued telemetry is ever truncated or re-encoded, the suite says so instead of the docs quietly ceasing to be reproducible.

Heads-up for whoever merges second: this and #109 both create `.gitattributes` as a new file, so one of them will hit an add/add conflict. Resolve it by keeping both sides.

PR #108, opened 2026-08-04
Make a klone YOLO checkpoint resumable on Tillicum, and move y11x_pano there to finish its schedule (#51)

jonfroehlich:
Adds `scripts/model_comparison/retarget_yolo_checkpoint.py`, the tool that let `y11x_pano` continue on Tillicum as one continuous training trajectory rather than a restart. Job **207774** is running; full context and numbers are in the [#51 status comment](https://github.com/ProjectSidewalk/RampNet/issues/51#issuecomment-5181927292).

## Why

`resume=True` reuses **every** training arg saved in the checkpoint. That is what keeps the LR schedule and epoch counter continuous across a preemption — and it is exactly what makes a checkpoint non-portable, because `data` / `project` / `name` / `save_dir` / `model` / `resume` are all absolute paths on the cluster that wrote them.

Two failure modes, one of them silent:

1. the saved `data` path is missing on the new host → the run dies at dataset load (obvious); or
2. a copied `save_dir` **does** exist → Ultralytics writes results back into the **original** run directory and corrupts the source run.

(2) is the trap already recorded in the #51 weight-snapshot MANIFEST. This replaces a warning that describes it with a tool that avoids it.

## What it does and does not touch

Rewrites exactly those six path keys. Every hyperparameter is deliberately left alone — `epochs`, `patience`, `lr0`/`lrf`, `batch`, `imgsz`, `workers`, `close_mosaic`, `seed`, `optimizer` define the pre-registered #71 schedule, and changing one mid-run would make the resumed arm a different config rather than a continuation.

Worth stating for `epochs: 60` specifically: it is the **denominator of the LR decay**, not a label. Rewriting it would not shorten the experiment — it would change what every remaining epoch does. That is also why a run cannot be retroactively redeclared as a shorter one.

Dry run by default. `--apply` writes, keeps a `.preretarget` backup, and reloads the written file to assert the keys persisted rather than trusting `torch.save`.

## Verified in use

- checkpoint sha256 identical klone → workstation → Tillicum (`8f85791587dd…729c`, matching `.sha256.verified`)
- pano val **labels and JPEGs md5-match byte-for-byte** across clusters — stronger than the sorted-filename-list md5 recorded in `docs/tillicum.md`
- resume confirmed: `Resuming training … from epoch 22 to 60 total epochs`, with `optimizer=auto → MuSGD(lr=0.01, momentum=0.9)` identical to the klone runs

## Dependency and a follow-up for #91

The run uses `scripts/model_comparison/run_yolo_train_tillicum.slurm`, which lives on `docs/tillicum-onboarding` (#91) and **not** on `main` — so reproducing this run today needs both branches. Merging #91 removes that.

Also for #91: its claim *"The YOLO stack, however, is VERIFIED (2026-07-31)"* should be narrowed. It was established by a **data-prep** job; no training job had ever run there. The first training submit (job 207761) failed at 4 min with `RuntimeError: operator torchvision::nms does not exist` — Tillicum had the generic `torchvision 0.28.0` wheel against `torch 2.13.0+cu126` where klone has `0.28.0+cu126`. Prep never calls NMS, so prep could not have caught it. Cost of the failure: \$0.06. `numpy` also differs (2.4.6 vs 2.4.4).

## Test plan

No automated test: the script's input is a multi-hundred-MB Ultralytics checkpoint, which does not belong in fixtures. It self-verifies instead — reloads after writing and asserts all six keys persisted, exiting non-zero otherwise — and was exercised end to end on the real checkpoint (dry run, then `--apply`, then a resumed job that reported the correct epoch).

### DIFF
--- a/scripts/model_comparison/retarget_yolo_checkpoint.py
+++ b/scripts/model_comparison/retarget_yolo_checkpoint.py
@@ -33,6 +33,14 @@ experiment, it changes what every remaining epoch does.
 keeps only ``last.pt``/``best.pt``, and resume honours the saved value, so per-epoch
 weights cannot be recovered for an arm that did not start with it.
 
+Where to run it
+---------------
+On the TARGET cluster, after the checkpoint has been copied there. That is the assumed
+workflow, and it is why ``--data`` is checked for existence: on the target host a
+missing data.yaml is a typo caught in a second rather than a job that dies at dataset
+load minutes later. To prepare a checkpoint somewhere the target paths do not exist
+yet -- staging on a laptop before an rsync, say -- pass ``--no-check-data``.
+
 Usage
 -----
     python retarget_yolo_checkpoint.py CKPT.pt \
@@ -50,20 +58,56 @@ import argparse
 import hashlib
 import shutil
 import sys
-from pathlib import Path
+from datetime import datetime, timezone
+from pathlib import Path, PurePosixPath, PureWindowsPath
 
 # The six keys that carry a cluster-absolute path. `resume` and `model` both point at
 # the checkpoint itself; ultralytics rewrites `model` internally on load, but we set
 # both so the file is self-consistent if inspected.
 PATH_KEYS = ("data", "project", "name", "save_dir", "model", "resume")
 
+_CHUNK = 1 << 20  # 1 MiB
+
 
 def sha256(path: Path) -> str:
     h = hashlib.sha256()
     with path.open("rb") as fh:
-        for chunk in iter(lambda: fh.read(1 << 20), b""):
+        for chunk in iter(lambda: fh.read(_CHUNK), b""):
             h.update(chunk)
     return h.hexdigest()
+
+
+def _flavour(p: str):
+    """Which path flavour the TARGET cluster speaks, inferred from the path itself.
+
+    Not the host's: staging a Tillicum checkpoint from a Windows laptop
+    (``--no-check-data``) would otherwise write ``\\gpfs\\projects\\...`` via
+    ``pathlib.Path``, and the resumed run would look for a directory that cannot
+    exist on Linux. Judged on the string given, so the same command produces the
+    same checkpoint wherever it is run.
+    """
+    return PureWindowsPath if ("\\" in p or PureWindowsPath(p).drive) else PurePosixPath
+
+
+def retarget_paths(data: str, project: str, name: str) -> dict[str, str]:
+    """The six path values a checkpoint needs to resume under ``project/name``.
+
+    Coupled to ultralytics' run layout: the trainer derives its output directory as
+    ``project/name`` and writes weights to ``<save_dir>/weights/last.pt``. If that
+    layout ever changes upstream, this function is the one place to fix -- writing
+    these keys by any other rule would produce a self-inconsistent checkpoint.
+    """
+    pure = _flavour(project)
+    save_dir = str(pure(project) / name)
+    weights = str(pure(save_dir) / "weights" / "last.pt")
+    return {
+        "data": str(data),
+        "project": str(project),
+        "name": name,
+        "save_dir": save_dir,
+        "model": weights,
+        "resume": weights,
+    }
 
 
 def main() -> int:
@@ -77,6 +121,9 @@ def main() -> int:
                     help="write here instead of in place")
     ap.add_argument("--apply", action="store_true",
                     help="actually write; without this the script only reports")
+    ap.add_argument("--no-check-data", action="store_true",
+                    help="skip the --data existence check, for staging a checkpoint "
+                         "on a host where the target cluster's paths do not exist")
     args = ap.parse_args()
 
     import torch  # imported late so --help works without a torch install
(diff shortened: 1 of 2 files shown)


jonfroehlich (commented):
## Deep Code Review — 
etarget_yolo_checkpoint.py

### Overall Assessment

This is a well-written, focused utility script. The PR description is exemplary — it explains the *why* in depth, documents both failure modes it prevents, and records the verification evidence. The code is clean, the docstring is thorough, and the dry-run-by-default / self-verification pattern is exactly the right approach for a tool that mutates multi-hundred-MB artifacts. That said, a few things are worth flagging:

---

### 🟢 Strengths

1. **Dry-run by default** — correct default for a destructive operation. The `--apply` gate and the reload-and-assert pattern after writing are both excellent.

2. **Disciplined scope** — the script touches exactly the six path keys and explains *why* every other key is left alone. The `epochs=60` LR-denominator note in the docstring is particularly valuable context that will save someone real confusion later.

3. **Self-verifying write** — reloading the written checkpoint and asserting all six keys persisted catches silent `torch.save`/pickle failures. Returning non-zero on mismatch means this composes well with `set -e` in a Slurm script.

4. **Late `import torch`** — so `--help` works without a torch install (useful on login nodes). Consistent with how other scripts in this repo handle heavy imports.

5. **`.preretarget` backup** — one-shot, won't clobber an existing backup, so applying twice doesn't lose the original.

---

### 🟡 Suggestions (non-blocking)

**1. `--data` existence check may be too eager for cross-cluster prep**

Lines 88-91 validate that the `--data` file exists *on the machine running the script*. If someone prepares the retargeted checkpoint on their laptop before `rsync`-ing it to Tillicum, the data.yaml won't exist locally. The check will block the retarget even though the path is correct for the target cluster.

Consider making it a warning in dry-run mode and only fatal with `--apply`, or adding a `--no-check-data` escape hatch:

\\\python
if not data_yaml.is_file():
    if args.apply:
        print(f"WARNING: --data does not exist on this host: {data_yaml}", file=sys.stderr)
        print("         The checkpoint will reference this path on the target cluster.", file=sys.stderr)
    else:
        print(f"NOTE: --data not found locally (expected if retargeting for a remote host)", file=sys.stderr)
\\\

That said, if the workflow is *always* to run the script on the target cluster, the current check is the right trade-off (catches typos early). Just worth documenting that assumption.

**2. `save_dir` construction assumes Ultralytics' `project/name` layout**

Lines 96-97 build `save_dir = project/name` and `weights = save_dir/weights/last.pt`. This matches current Ultralytics layout, but it's an assumption worth calling out — if Ultralytics ever changes this layout, the script would write self-inconsistent paths. A comment noting this coupling would help future-proof.

**3. No `--version` or provenance stamp in the modified checkpoint**

After retargeting, there's no record *in the checkpoint itself* that it was retargeted, or from which source paths. If someone later inspects `train_args` and wonders why the paths differ from the original cluster, there's no breadcrumb. Consider adding a key like `retarget_info` to the checkpoint dict (not inside `train_args`, to avoid confusing Ultralytics):

\\\python
ckpt["retarget_info"] = {
    "script": "retarget_yolo_checkpoint.py",
    "original_save_dir": ta.get("save_dir"),
    "timestamp": datetime.now().isoformat(),
}
\\\

Low priority — the `.preretarget` backup serves a similar purpose if it's kept alongside.

**4. `sha256` printed for input but semantics differ for output**

The input sha256 is printed to enable cross-cluster integrity verification (matching the MANIFEST). The output sha256 is printed too, but it will differ from the input by definition (contents changed), and from any future re-save (`torch.save` isn't bit-reproducible across Python/torch versions). Worth a brief comment noting the output hash is useful for *this specific written file*, not for cross-host comparison.

---

### 🔵 Nits (take or leave)

- **Line 106**: `1 << 20` is idiomatic for 1 MiB but a named constant or comment (`# 1 MiB`) would help readability for anyone unfamiliar with the pattern.

- **Lines 115-116**: The `print` formatting hardcodes column widths (58 chars for the "from" column). Long Slurm paths (common on HPC clusters) could overflow and misalign the table. Consider using dynamic widths:

  \\\python
  w = max(len(str(ta.get(k, ""))) for k in PATH_KEYS)
  \\\

- **Missing `__all__` or explicit "not a library" note**: The script is entry-point-only (`raise SystemExit(main())`), so it's not importable as a module in any useful way. This is fine — it matches the pattern used across this repo. No action needed.

---

### ⚪ Questions for the author

1. The PR body mentions the checkpoint sha256 is identical across three clusters *before* retargeting. After retargeting, is the expectation that only the `.preretarget` backup matches the original MANIFEST hash? If so, should the MANIFEST be updated to record the retargeted hash too?

2. The PR notes that reproducing this run requires both this branch and `#91` (`docs/tillicum-onboarding`). Is there a merge-order dependency, or can they land independently? (If #91 should land first, a GitHub label or PR dependency link would make that visible.)

---

### Verdict

**Approve** — this is clean, well-scoped work. The suggestions above are quality-of-life improvements, none are blocking. The script does exactly what it says, the self-verification is thorough, and the documentation is strong.

🤖 Generated with Google Antigravity (Claude Opus 4.6)

jonfroehlich:
All four suggestions and both nits are in 83cf6f3 — and writing the tests for the first one turned up a real bug.

**The `--data` check stays, with an escape hatch.** The assumed workflow is to run this on the target cluster, where a missing `data.yaml` is a typo caught in a second rather than a job that dies at dataset load. `--no-check-data` covers the staging case, and the docstring now says which is which instead of leaving it implicit.

**That escape hatch exposed a portability bug.** The six paths were built with `pathlib.Path`, which emits *host* separators — so staging a Tillicum checkpoint from this Windows desktop would have written `\gpfs\projects\...` into `train_args`, and the resumed run would have looked for a directory that cannot exist on Linux. The flavour is inferred from the target path now, so the same command produces the same checkpoint wherever it runs. Every retarget so far ran on the target cluster, so this never bit; the escape hatch is what would have made it reachable.

**The rest:**
- The ultralytics `project/name` coupling is stated once, in `retarget_paths`, rather than assumed inline.
- `retarget_info` is stamped at the top level (not inside `train_args`, so ultralytics never sees a key it does not know): where the paths came from, where they went, when. Worth saying in the comment and here — it survives only until ultralytics writes its own next checkpoint over `last.pt`. It identifies the file handed to the job; `.preretarget` keeps the original bytes.
- Both sha256 prints now say what they are for. Input matches the snapshot MANIFEST; output identifies that file only and is not comparable across hosts, since `torch.save` is not bit-reproducible across torch/python versions.
- `1 << 20` is a named constant, and the before/after table sizes its column from the data.

**`tests/test_retarget_yolo_checkpoint.py`** (10 tests) covers the pair of silent failures this script exists to prevent: the six path keys must change, and the schedule must not — `epochs` is the LR-decay denominator, so rewriting it would change what every remaining epoch does rather than relabelling the run. 511 passed.

**Your two questions:**

1. **No, the MANIFEST should keep the pre-retarget hash only.** That hash identifies the training run's own bytes, which is what makes "the same checkpoint reached three clusters" checkable. The retargeted file is a per-target derivative — and not a reproducible one, since `torch.save` output varies with torch and python versions, so a recorded hash for it would fail for someone doing the same retarget correctly. `.preretarget` is what matches the MANIFEST, and that is the right invariant.
2. **No merge-order dependency.** #91 is the cluster documentation and this is the tool; each stands alone, and reproducing the run wants both only in the sense that one explains the machine and the other the command.

PR #109, opened 2026-08-04
Publish the paper's Stage 1 inputs, and plan the rest onto Hugging Face

jonfroehlich:
`v1.0-iccv2025` shipped the Stage 1 code but not the data it consumes. This publishes what can go
in git, proves what the published files became, and writes down where the rest goes.

## The gap this closes

The three open-government curb ramp inventories the paper ran on were excluded from git by a
`location_data/*` line in `stage_one/dataset_generation/.gitignore`. That line is how the omission
happened, and it is **not** a gap re-downloading can close: the city portals serve *current* data
and it drifts (Bend +8.7%), so a download made today reproduces a different experiment.

All three are now committed (71.8 MB) and hash-pinned in `docs/data_provenance.md` §3, with a new
`.gitattributes` marking them `binary` so no contributor's `core.autocrlf` can silently invalidate
those hashes.

## Which government records became training labels

`combine_location_data.py` reduces the inventories to `(latitude, longitude, date)` and shuffles,
discarding NYC's `RampID`/`CornerID`, the geojson `OBJECTID`/`FacilityID`/`NonAssetID`, and even
the city label — so `all_locations.csv` cannot answer "is this ramp in the training set?".

`scripts/analysis/gov_provenance.py` rebuilds the mapping and verifies it: **276,071 / 276,071 rows
resolved, 0 unmatched.**

| | Bend | Portland | NYC | total |
| government records | 13,357 | 45,035 | 217,679 | 276,071 |
| consumed by a generated panorama | 5,110 | 21,075 | 130,527 | **156,712** |
| consumption rate | 38.26% | 46.80% | 59.96% | **56.77%** |

**43.23% of the government records never became a training label.** That belongs next to any claim
about how much open-government data this pipeline converts.

## Three findings that constrain any Stage 1 re-run

1. **The paper-era shuffle was unseeded.** `random.seed(42)` was added later, so the published row
   *order* is unrecoverable. My first implementation replayed the seeded shuffle; verification
   caught it on row 0, and the working method is a coordinate join instead. This is the Stage 1
   counterpart of the split-seeding caveat in `data_provenance.md` §4.
2. **`convert_date` changed after the paper.** Unknown install dates were `"2000-01-01"`, now `""`.
   Measured: 276,066/276,071 dates reproduce under paper-era logic vs **252,983/276,071 (91.64%)**
   under current `main`. The 23,088-row gap is exactly the records with **no install date** —
   8.36% of the corpus, every one silently admitted to the installed-before-captured check.
3. **8 coordinates are shared by two government records** (16 rows), and that is necessarily the
   whole of the residual date disagreement.

## Street centrelines: 801.6 MB → 18.7 MB, proven equivalent

`street_data/` could never live in git — `New York - Streets.geojson` alone is 669 MB. But its only
consumer, `generate_negative_panos.py`, reads the geometry plus **one** name field used solely as
an emptiness test. Route numbers, ZIP, MSAG, ESN, road class and county are never touched.

`scripts/build_street_derivative.py` strips the rest: **42.9× smaller**, New York alone 67.5×. The
name field is kept rather than dropped because Portland has 4,192 features with an empty
`FULL_NAME` that `load_city_streets` skips — a pure-geometry file would silently re-admit them.

Equivalence is proved, not asserted. `verify` computes a **consumer fingerprint** — sha256 over the
ordered (name, geometry) pairs of every feature surviving the filter, which is exactly what the
length-weighted sampling index is built from. All three match, and reproduce identically on Windows
and on the cluster. `generate_negative_panos.py` prefers a full download when present, so existing
checkouts are unaffected.

## The Hugging Face plan

`docs/replication.md` now carries the full plan and `README.md` a "Published Artifacts" table, so
no input is discoverable only by asking us. Ordering is driven by what blocks what, and the top
item was not the obvious one:

- **`rampnet-crop-model` (720.7 MB) first.** `inference_isolator.py` hardcodes a path to the
  round-2 crop checkpoint — the model that turns every government coordinate into a pixel keypoint
  — and it exists only on lab storage. Publishing `location_data/` without it hands someone the
  inputs to a pipeline they still cannot run.
- **`rampnet-stage1-inputs` (1.06 GB), manifests first.** `finaldataset.jsonl` and
  `negativepanosSHORTENED.jsonl` are the only way to reproduce the paper's negatives, because the
  sampler is unseeded.
- **`rampnet-benchmark` (~12 GB, #21) incrementally.** Benchmark imagery is immutable once fetched
  and the revisable parts live in git, so the set only grows and a new city costs one folder.
  Folders named `panos_native/` and `panos_4096x2048/` — by resolution, not by consumer, so the
  name cannot rot when the model's input size changes.

Checked the two live repos against the July 2025 terms rather than assuming we had followed them:
`rampnet-dataset` is Parquet in 128 shards per split as advised; `rampnet-crop-model-dataset` is
1,214 loose JPEGs, under the per-folder limit but not the recommended shape.

Also checked the whole plan against HF's current limits, and nothing is within an order of
magnitude of one: largest file 669 MB (limit 500 GB), largest folder 125 entries (limit 10,000),
largest repo ~2,550 files (recommended <100k). The per-city subdirectory layout is what keeps the
folder count trivial. The real consideration is the storage quota — `projectsidewalk` is a *free*
org, whose public storage is "best-effort", and the existing 463 GB rests on the July 2025 human
approval rather than an entitlement. Our +29 GB is +6%, but a courtesy note on that thread belongs
in the publish steps.

## Published to Hugging Face

All three planned repos are live, in the [RampNet collection](https://huggingface.co/collections/projectsidewalk/rampnet):

| repo | size | notes |
| :--- | ---: | :--- |
| [`rampnet-crop-model`](https://huggingface.co/projectsidewalk/rampnet-crop-model) | 720.7 MB | rounds 1+2; **Stage 1 could not run without round 2** |
| [`rampnet-stage1-inputs`](https://huggingface.co/datasets/projectsidewalk/rampnet-stage1-inputs) | 1.03 GB | inventories, raw street data, manifests |
| [`rampnet-benchmark`](https://huggingface.co/datasets/projectsidewalk/rampnet-benchmark) | 11.41 GB | Parquet; configs `native` / `4096x2048` / `galleries` — the imagery half of #21 |

Uploads were verified rather than trusted: HF stores each LFS object under its sha256, so oids were compared directly against local hashes (8/8, then 27/27). Files too small for LFS were re-downloaded and hashed. For the benchmark, `load_dataset()` against the Hub returns bytes whose sha256 matches both the row and the committed `imagery_manifest.json`.

**All three human review passes are now redoable by someone outside the lab.** GT verification and the #55 A/B previously needed imagery that existed only on lab machines. In particular the second-rater pass on Budapest — the top outstanding follow-up in the model comparison — is now blocked on a person, not an artifact.

Remaining: the 15 GB round-1 Project Sidewalk crop set, planned for `rampnet-crop-model-dataset/round1_ps/`.
## Verification

- `pytest -q` — **501 passed**, at each commit.
- `gov_provenance.py` re-run under both paper-era and current `convert_date`; both figures quoted
  above come from that run.
- `build_street_derivative.py verify` — 3/3 fingerprints match, on two platforms.
- The three inventories' sha256 confirmed identical to the cluster copies after transfer.


jonfroehlich:
## Round 1 is published, and the crop-repo pair now reads correctly

Status as of 2026-08-05 ~15:00 PDT. Two Hub-side changes shipped today, both recorded in `docs/replication.md` §4 (commits `6a5b554`, `2fd7be9`):

### 1. The round-2 repo was renamed: `rampnet-crop-model-dataset` → [`rampnet-crop-model-dataset-round2`](https://huggingface.co/datasets/projectsidewalk/rampnet-crop-model-dataset-round2)

The separate-repos decision (`11e6125`) had left the naming backwards — the unmarked repo held round *2* while the new repo carried the `-round1` suffix. A rename fixes that only if it breaks nobody, so that was measured before it was done: `scripts/analysis/hf_move_redirect_check.py` (committed, same throwaway-repo shape as the config-mixing check) proves `HfApi.move_repo` leaves a redirect that keeps the old id working at the `dataset_info`, `/resolve/`-URL, and `load_dataset(OLD_ID)` layers. The real rename was then verified content-identical: same git sha before and after, all 1,214 files, downloads/likes carried, and the collection entry followed automatically.

**Standing obligation the redirect leaves behind (now in §4): the org must never create a new repo named `rampnet-crop-model-dataset`.** A new repo at the freed name would shadow the redirect and every stale link would silently point at the wrong dataset.

### 2. [`rampnet-crop-model-dataset-round1`](https://huggingface.co/datasets/projectsidewalk/rampnet-crop-model-dataset-round1) is live — the last unpublished paper-era input

27,704 crops, 11 Parquet shards, 13.37 GB, in the RampNet collection (7 items), viewer up. Verified at every hop:

| hop | check | result |
| :--- | :--- | :--- |
| klone → local | tar sha256 vs job log | `7fd446c1…8ec9`, identical |
| crops → shards | `verify` re-hashes every crop | 27,704/27,704 byte-identical, 35,757 keypoints independently re-parsed |
| local → Hub | all 11 shards vs Hub LFS sha256 (content-addressed, no re-download) | identical |
| Hub → user | remote parquet footers; one live-streamed row re-hashed | 19,392/4,155/4,157 = 27,704 rows; sha matches |

Cards: the round-1 card carries the `crop_uid` warning (the filename token is **not** a panorama id) and the per-axis table for the 3.1% x-only label/image scaling mismatch, both found while building this export. The round-2 card (README-only — no `configs:` block can ever be added there, per the mixing check) gained its identity banner, the two-round cross-link table, and front-matter metadata; the `rampnet-crop-model` card's three dataset links now carry the canonical `-round2` name.

With this, every input in the replication ledger's HF table is ✅. The one item still owed from the 2025 correspondence is the courtesy note to the datasets team, which §3 records.

jonfroehlich:
## Review findings addressed — all 15, plus main merged

Two fix commits (b31e3ad code, 213c3bf docs) and a merge of `main` (0d3275e). Suite went 510 → 535 on the fixes, and is **986 passed / 1 skipped** on the merged tree; both CI legs pass.

### The three that would have reached the Hub

Every finding in the code commit shares a shape: the artifact uploads, nothing raises, and the damage is visible only to whoever tries to use it.

1. **A records-only re-push dropped three configs.** `card()` rebuilt the README's `configs:` block from whatever Parquet happened to be in `--out`. But `records` mode exists so a verdict fix costs megabytes instead of 11.41 GB — so it runs against a directory holding only `data/records/`, rendering a one-config card and orphaning the image parquets: still on the Hub, unreferenced, and `load_dataset(repo, "4096x2048")` starts failing. `build` now writes `build_manifest.json`, `card` unions it with the disk scan, and a card that would still drop an imagery config is **refused rather than rendered**.
2. **`benchmark/*/panos` swept in the paper's gold set.** `manual_gold/panos` holds 1,000 panoramas already published inside `rampnet-dataset`; the glob made them a 10th `native` split. The manifest cross-check couldn't catch it, because that directory has no `imagery_manifest.json` — it printed "skipped" and exited PASS. Now an explicit allowlist, kept in step with `miss_decomposition.ALL_SPLITS` by a test, and the skip is announced.
3. **`galleries` published the crop tag in a column named `pano_id`.** `low_floor_sweep.py` names a crop `{pano}_{x:.5f}_{y:.5f}`, so the join the card documents returned zero rows — the same trap `test_the_opaque_token_is_never_published_as_a_panorama_id` was written for on the crop set, repeated in the sibling exporter. Rows now carry `crop_id`, a `pano_id` parsed back out of it (right-anchored, so panorama ids containing their own underscores survive), and the detection's x/y. A stem that isn't a tag gets a null and a printed count, not a guess.

### The rest

- `verify` FAILED only on a sha256 mismatch, so an unpinned split still printed "the exact bytes reviewers judged". Missing manifests and absent panoramas now fail; `--allow-unpinned` for a split still being assembled.
- Neither exporter cleared `--out`, so shards from a shorter previous build were counted into the card totals and uploaded — duplicating rows in the released split, every orphan still matching its own sha256. Now cleared per config/split, never wholesale, so a config this run doesn't rebuild survives.
- The `native` branch accepted every regular file while the others filtered by extension; one `Thumbs.db` reached PIL and killed an 11 GB build with no resume.
- `all_locations.csv` was staged unverified, though current `main` provably regenerates a different one (seeded shuffle, changed date semantics — 8.36% of rows). Now hash-pinned like `location_data/`, with `--expect-sha256` to add more as they become known.
- The card's build-date range was hardcoded prose, wrong since São Paulo landed on 2026-08-01. Now computed from `verdicts.json` (2026-07-22 → 2026-08-01).
- The GeoJSON readers used the platform's locale codec, so the fingerprint the street-derivative equivalence proof rests on was machine-dependent. Pinned to UTF-8 on both sides; the three committed derivatives fingerprint unchanged, and their sha256s still match `data_provenance.md`.
- `git_commit()` and a chunked sha256 existed in six and four copies. One `scripts/hf_export_common.py` now, used by all eight consumers — and the fix owed to all of them, a `-dirty` marker so a card built from edited sources stops claiming clean provenance, landed once.
- Both HF probe scripts leaked their `mkdtemp` workdir plus a full datasets cache on every run, and one leaked a file handle.
- Three documentation defects where a reader is told something the repo itself disproves: §9 of `curb_ramp_data_sourcing.md` still recommended what this PR does (a reader landing there re-downloaded drifted portal data — the exact failure the PR prevents); `replication.md` described three configs of 9 splits where there are four and `galleries` has 8, and named folders the Parquet exporter never creates; and the stage-1 card claimed one provenance standard for artifacts held to two.

### Two things worth your attention

1. **The 276,071 vs 276,615 gap is documented, not closed.** The committed inventories hold 544 fewer records than Table 1 (Bend −254, Portland −289, NYC −1). I verified the pipeline discards nothing — every geojson feature has ≥2 coordinates, every NYC row parses as a POINT — and `all_locations.csv`, the run's own recovered output, has exactly 276,071 rows that all resolve against these three files with none unmatched. That is good evidence these are the files the paper consumed. What I could not determine is where Table 1's number came from; my best reading is portal-advertised counts read at survey time, but that is a hypothesis and §3.3 says so rather than asserting it. Whoever compiled Table 1 can settle it.
2. **`galleries` needs a 244 MB re-push.** The join-key fix above changes the published schema, and it is in code only — the Hub still has the broken column. Recorded in `docs/replication.md` so it doesn't get lost.

### On the merge

`main` had moved 96 commits ahead and both sides edited the same three files. None of the conflicts were mechanical — each was the same claim at two vintages, so keeping both sides would have shipped self-contradicting docs:

- **`.gitattributes`** (add/add) — a genuine union; disjoint paths, and main's last-match-wins ordering for its `SHA256SUMS` lines is preserved.
- **`replication.md`** — main still listed the benchmark panoramas as a 9.0 GB git-ignored blocker; 348eee2 published them, so that row is gone. Both sides' detection-corpus sizes had also gone stale against their own trees (18.0 MB here for 20.1 MB of files; 20.2 MB on main for 22.9 MB), so it is now the measured 22.9 MB with the file count beside it.
- **`curb_ramp_data_sourcing.md` §9** — both sections kept, since they answer different questions: `location_data/` is what the paper consumed, `data/inventories/` is the frozen 2026-07-31 state the 2.0 supply analysis runs on. Main's caveat that "recovering and committing the paper-exact files is still open" is what this PR closed, so it now says where they are and that they came from the paper run's scratch directory, not the supplemental material. One more instance of that same contradiction, in §5, survived 213c3bf and is fixed here.

Issue #9, opened 2026-07-02
Fix prediction-GT matching in both model evaluators (TP double-count / dropped duplicate FPs); re-quantify gold-set P/R


jonfroehlich:
Both model evaluators implement prediction→GT matching without one-to-one discipline, and the errors bias the published numbers upward:

- `stage_two/evaluate.py:419-433`: the inner GT loop has **no `break`** — a single prediction within the matching radius of two ground-truth points appends **two TP entries** (inflates recall and the TP mass under the PR curve). Worse, a prediction whose only in-radius GTs are already claimed sets `passes=True` but appends **nothing** — standard protocols (VOC/COCO) count duplicate detections as false positives; here they vanish, inflating precision.
- `stage_one/crop_model/ps_and_manual_model/evaluate.py:365-385`: same duplicate-drop pattern (duplicate matches are neither TP nor FP).
- By contrast, `stage_one/dataset_evaluation/evaluate.py:144-163` does this **correctly** (breaks after claiming the first unclaimed GT, buckets duplicates separately) — so the repo contains both the right and wrong pattern, and the Stage-2 evaluator (the one producing the paper's model PR curves) is the buggy variant.

Related metric issues to fix in the same pass:
- AP is computed with **raw, uninterpolated** precision (`stage_two/evaluate.py:217-250`) — not comparable to VOC/COCO AP.
- The 0.022 "radius" is applied as `0.022 * heatmap_w` against x scaled by w=1024 and y scaled by h=512 (`evaluate.py:326,420-421`) — it's actually an **ellipse** 2× more permissive vertically. Consistent with Stage 1, but should be documented.

Fix: greedy one-to-one matching (predictions sorted by confidence, each claims at most one unclaimed GT, duplicates counted as FP), optional interpolated AP. Then **re-run the 1k-manual-gold-set evaluation with corrected matching and quantify the delta vs the published precision/recall**; if material, add an erratum note to the README (and consider an arXiv v2 note).

From the post-release code review (July 2026).

jonfroehlich:
Resolved. Summary of what landed, plus three corrections to this issue's original claims after a full re-verification.

**What landed**
- One-to-one greedy matching (each detection claims at most one unclaimed GT; duplicates are FPs) is implemented in `rampnet/metrics.py` and used by `stage_two/evaluate.py`.
[...]

PR #110, opened 2026-08-05
Restate the verification report's preamble, and declare its charset

jonfroehlich:
`docs/eval_protocol_verification.html` opened by quoting the chat instruction that produced it, verbatim and attributed by name:

> Requested by Jon: "Are you absolutely certain there was an error in the original paper and our experimental approach? Double- and triple-check, produce visual examples of what we double-counted, and explain in greater detail."

That was fine for an internal note. It is now linked from the `v1.1-corrected-eval` release notes and from [#9](https://github.com/ProjectSidewalk/RampNet/issues/9), so it reads as a public document, and an instruction quoted out of its conversation is both informal and less informative than the thing it asked for. The preamble now states the work's scope directly, keeping every element of the original ask — certainty, independent confirmation rather than a single line of evidence, worked visual examples, and a quantified effect:

> A re-examination of the matching rule used to score RampNet's detections, prompted by a discrepancy identified after publication. Its scope: to establish whether predictions were genuinely over-counted, to confirm that finding by independent means rather than rest on a single line of evidence, to document the over-count with worked examples drawn from the released model's own output, and to quantify its effect on the published metrics.

The `<h1>` is neutralised in the same spirit (`624056d`): "Did we really over-count? A re-verification of the published evaluation protocol" → **"Re-verification of the published evaluation protocol"**. The old heading was the question the investigation started from, phrased in the first person and addressed to whoever commissioned it. The subject it carried — the over-count — is stated by the preamble immediately below, so the document loses nothing. The `<title>` was already neutral and is unchanged.

**Second change, one line:** the file never declared a charset. It *is* valid UTF-8, but with no `<meta charset>` a browser opening it from `file://` falls back to the locale's legacy encoding and renders every em dash and curly quote as `â€"`. Since the release notes tell readers to "open it in a browser", that is the path most likely to be hit. The bytes are untouched — this adds a declaration, it does not re-encode.

No change to any finding, number, figure or section of the report; the diff is the two header lines plus the meta tag.

Worth landing before the `v1.1-corrected-eval` tag is cut, since the tag will freeze whatever this file says.

### DIFF
--- a/docs/eval_protocol_verification.html
+++ b/docs/eval_protocol_verification.html
@@ -66,7 +66,7 @@
 
 <header class="col">
   <div class="eyebrow">RampNet · internal verification · 2026-07-02</div>
-  <h1>Did we really over-count? A re-verification of the published evaluation protocol</h1>
+  <h1>Re-verification of the published evaluation protocol</h1>
   <p class="sub">A re-examination of the matching rule used to score RampNet's detections, prompted by a
   discrepancy identified after publication. Its scope: to establish whether predictions were genuinely
   over-counted, to confirm that finding by independent means rather than rest on a single line of
--- a/docs/eval_protocol_verification.html
+++ b/docs/eval_protocol_verification.html
@@ -1,3 +1,4 @@
+<meta charset="utf-8">
 <title>RampNet eval protocol — verification report</title>
 <style>
   :root{
@@ -66,8 +67,11 @@
 <header class="col">
   <div class="eyebrow">RampNet · internal verification · 2026-07-02</div>
   <h1>Did we really over-count? A re-verification of the published evaluation protocol</h1>
-  <p class="sub">Requested by Jon: “Are you absolutely certain there was an error in the original paper and our
-  experimental approach? Double- and triple-check, produce visual examples of what we double-counted, and explain in greater detail.”</p>
+  <p class="sub">A re-examination of the matching rule used to score RampNet's detections, prompted by a
+  discrepancy identified after publication. Its scope: to establish whether predictions were genuinely
+  over-counted, to confirm that finding by independent means rather than rest on a single line of
+  evidence, to document the over-count with worked examples drawn from the released model's own
+  output, and to quantify its effect on the published metrics.</p>
 </header>
 
 <section class="col">
--- a/docs/eval_protocol_verification.html
+++ b/docs/eval_protocol_verification.html
@@ -1,3 +1,4 @@
+<!doctype html>
 <meta charset="utf-8">
 <title>RampNet eval protocol — verification report</title>
 <style>
@@ -281,8 +282,8 @@ pred#3 (conf 0.62) duplicate of an
 </section>
 
 <section class="col">
-  <h2><span class="no">7</span>Corrections to my own earlier claims</h2>
-  <p>You asked me to triple-check; the check cut both ways. Three things in the drafted issue #9 / README erratum
+  <h2><span class="no">7</span>Corrections to earlier claims in this investigation</h2>
+  <p>The re-check cut both ways. Three things in the drafted issue #9 / README erratum
   need fixing before anything is posted:</p>
   <div class="tblwrap"><table>
     <tr><th>Earlier claim</th><th>Verdict on re-check</th></tr>


jonfroehlich (commented):
## Deep Code Review: PR #110

**Status**: ✅ **Approve / LGTM** (with minor polish suggestions)

---

### Summary & Context
This PR refines [`docs/eval_protocol_verification.html`](docs/eval_protocol_verification.html) by:
1. Adding `<meta charset="utf-8">` to ensure correct rendering of typographic characters.
2. Replacing the conversational preamble quoting the original chat prompt with an objective, well-structured scope statement.
3. Neutralizing the main `<h1>` heading to align with the `<title>` tag and public documentation standards.

Given that this verification report is referenced directly in the project `README.md`, GitHub issue [#9](https://github.com/ProjectSidewalk/RampNet/issues/9), and the upcoming `v1.1-corrected-eval` release, transitioning it from an internal conversational memo to a self-contained, public-facing technical artifact is an important improvement prior to cutting the tag.

---

### Detailed Review

#### 1. Charset Declaration (`<meta charset="utf-8">`)
- **Problem addressed**: The report contains extensive multi-byte UTF-8 typography—including em dashes (`—`), typographic curly quotes/apostrophes (`“`, `”`, `’`), math/delta symbols (`×`, `−`, `Δ`), and arrows (`→`). When opened directly in a browser from the local filesystem (`file://`) or a simple static server without an explicit `Content-Type: text/html; charset=utf-8` header, browsers default to the platform legacy encoding (e.g., Windows-1252), causing mojibake (e.g., `â€"`).
- **Implementation**: The one-line `<meta charset="utf-8">` declaration at the top of the document immediately forces standard UTF-8 parsing without altering existing file bytes or inline base64 image data.

#### 2. Preamble & Heading Restatement
- **Preamble**:
  - *Before*: Quoted the private prompt verbatim (`"Requested by Jon: 'Are you absolutely certain...'"`).
  - *After*: *"A re-examination of the matching rule used to score RampNet's detections, prompted by a discrepancy identified after publication. Its scope: to establish whether predictions were genuinely over-counted, to confirm that finding by independent means rather than rest on a single line of evidence, to document the over-count with worked examples drawn from the released model's own output, and to quantify its effect on the published metrics."*
  - *Assessment*: Captures all four core pillars of the original inquiry (certainty, multi-method verification, visual demonstration, and metric quantification) in clean, formal, parallel prose.
- **Heading**:
  - *Before*: `"Did we really over-count? A re-verification of the published evaluation protocol"`
  - *After*: `"Re-verification of the published evaluation protocol"`
  - *Assessment*: Aligns the visual header with `<title>RampNet eval protocol — verification report</title>` and maintains an objective, archival tone.

#### 3. Verification & CI Impact
- **Non-breaking**: No changes to experimental data, evaluation tables, metric ledgers, code snippets, or inline figures.
- **CI Status**: Pytest test suite passes across Python 3.10 and 3.12 without regression.

---

### Minor Observations / Suggestions for Future Polish (Non-blocking)

1. **Add `<!doctype html>` Declaration**:
   - The file currently begins directly with `<meta charset="utf-8">` (and lacks `<!doctype html>`).
   - Without a doctype declaration, browsers may default to **Quirks Mode** instead of **Standards Mode**, which can introduce subtle layout discrepancies in table border rendering or box sizing across different browser engines.
   - *Suggestion*: Consider placing `<!doctype html>` on line 1 above `<meta charset="utf-8">` (as is done in [`benchmark/richmond/box_gallery/index.html`](benchmark/richmond/box_gallery/index.html)).

2. **Remaining Conversational Phrasing in §7 (Line 281)**:
   - While the header and preamble are now neutral, Section 7 still begins with direct conversational address:
     > *"You asked me to triple-check; the check cut both ways. Three things in the drafted issue #9 / README erratum need fixing before anything is posted:"*
   - *Suggestion*: If full third-person neutrality is desired throughout the entire document, this could be phrased as:
     > *"A thorough re-examination revealed three points in earlier draft notes that warranted correction:"*

---

### Recommendation
The changes are clean, focused, and ready to merge before cutting the `v1.1-corrected-eval` release tag.

🤖 Generated with Google Antigravity (Gemini 3.7 Flash)

jonfroehlich:
Both suggestions taken, in e12059c.

- **`<!doctype html>`** on line 1. Without it browsers parse the report in quirks mode, and this file is linked from the README and from #9, so it should render the same wherever it is opened.
- **§7 is neutral now**, heading included: "Corrections to earlier claims in this investigation", and the opening is "The re-check cut both ways" rather than addressing the reader.

The other first-person passages are left alone deliberately — "I ported the loop character-for-character" records who did what, which is provenance rather than conversational residue, and removing it would make the method vaguer, not more formal.

Issue #114, opened 2026-08-14
Score crop-window rules against the manual_gold boxes: containment, margin, and context ratio (crop-clarity acceptance test)


jonfroehlich:
Part of the crop-clarity acceptance test (recorded in sidewalk-auto-labeler's `PLAN-crop-clarity.md`): before any AI-submitted label ships, the crop-window rule that will produce its Gallery image ([ProjectSidewalk/SidewalkWebpage#4865](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4865) plans to port `predict_crop_size` as "sizing v1") gets **scored against gold boxes**, not eyeballed. This repo holds the only gold extent data in the ecosystem — `manual_labels/` has 3,919 human-drawn curb-ramp boxes on the 1,000 manual_gold panos, consumed until now only as centers (`rampnet/detection_eval.py` drops w/h at parse time; #83 already flags them as the eval set for extent work).

## What to build

`scripts/analysis/crop_window_eval.py` — pure geometry, no GPU, no network, no pixels needed for the numbers (pano dims come from `benchmark/manual_gold/records.jsonl`):

For each gold box, place each candidate crop window at a prompt point and score:

- **containment** — gold box fully inside the window (seam-wrap aware; the visibly-broken failure mode);
- **margin** — signed min distance from box edge to window edge, normalized by window side (containment is margin ≥ 0, and violation depth is visible instead of binary);
- **context ratio** — object max side / window side, against the ~10–15% band the consumer survey converged on (`sidewalk-panorama-tools` `reports/2026-08-09-cropper-consumer-requirements.md` — a survey, not a measurement; treated as a target band only);

stratified by depression angle / flat-ground distance band. Two prompt modes:

- `gold-center` — window centered on the box center; isolates **sizing** error from placement;
- `detection` — window centered on the matched RampNet detection (standard 0.022 radius greedy match); the production-realistic number, since real crops are cut at detection points.

## Candidates scored

| candidate | what |
|---|---|
| v1-raw | `predict_crop_size` exactly as CropRunner runs it today (pixel-linear in native y) |
| v1-norm | the resolution-normalized port specified for SidewalkWebpage's CropService (compute in 6656-height reference space, scale back) |
[...]

Issue #83, opened 2026-07-29
Extend RampNet output from points to boxes/polygons (point-prompted SAM2; CenterNet-style size head; segmentation)


jonfroehlich:
## Context

RampNet's output is points — by design, because Stage 1 translates government GPS locations into pixel *points*, and the model regresses fixed-sigma Gaussians (sigma 10 on the 512×1024 heatmap) around them. Extent (bounding box / polygon) is nowhere in the supervision, so the predicted blob size reflects target sigma and confidence, not ramp size. Thresholding the current heatmaps cannot recover object extent, and nothing in the 850k-label training set can teach it.

But extent would be valuable: it upgrades the output from "where" to "where and how big," enables area/geometry-based condition features (RampNet-2.0-adjacent), and makes RampNet directly comparable to box-native baselines (#51) on their own terms.

### 1. Point-prompted SAM2 — boxes and polygons with zero training
RampNet's high-precision points (P 0.949 at the 0.55 operating point) are exactly what SAM-family models accept as prompts:

- RampNet detects a point on the pano
- crop a local region around it at full pano resolution (prompting on a crop sidesteps equirectangular distortion and the small-object problem)
- SAM2 with the point prompt → instance mask → polygon; tight bbox for free

RampNet stays the detector; SAM adds extent. This also doubles as a **label-minting machine**: run it over the ~850k Stage 1 training points to generate pseudo-boxes/masks at scale for paths 2–3.

**Main risk:** segmentation quality on low-contrast ramps that blend into surrounding concrete. This is an empirical question, and we can answer it cheaply — see "First experiment" below.

### 2. CenterNet-style size head on the existing architecture
Keep the ConvNeXt V2 backbone and center heatmap; add two regression channels (w, h at the peak). Minimal delta to native boxes, preserves the one-pass story. Needs box supervision at scale — which path 1 mints (SAM pseudo-boxes, filtered by mask quality). The 1k gold boxes are too few to train on and are the benchmark; they stay eval-only.

### 3. Full segmentation output
If polygons become first-class: Mapillary Vistas includes a curb-cut class with polygon annotations (coverage/quality to be verified before committing) — a fine-tune source that dovetails with the Mapillary direction (#25, #48) and sidewalk-auto-labeler. SAM pseudo-masks scale that supervision the same way Stage 1 scaled points.

## First experiment (cheap, decisive)

[...]

Issue #101, opened 2026-08-03
Flat-ground distance carries a ~11% range-proportional scale error, not just scatter — multi-view self-consistency measures it and DA3 can anchor it


jonfroehlich:
## The measurement

Multi-view association in the auto-labeler ([sidewalk-auto-labeler#27](https://github.com/ProjectSidewalk/sidewalk-auto-labeler/issues/27)) fuses detections of the same physical ramp seen from several panos. That makes a new check possible: **the same ramp, raycast from cameras at different distances, should land in the same place.** It does not.

For every fused site, take each member's *signed along-ray residual* — how far that view's ground point sits beyond (or short of) the fused position, measured along its own ray — and regress it on that member's range. Demeaned within each site (fixed effects, so per-site offsets cancel), pooled across sites with ≥3 refit members spanning ≥4 m of range:

| city | source | sites | members | slope (m per m) |
| :--- | :--- | ---: | ---: | ---: |
| paterson | GSV | 4,433 | 19,588 | **+0.1159 ± 0.0021** |
| gainesville | GSV | 1,774 | 7,033 | **+0.1179 ± 0.0038** |
| bend | GSV | 9,388 | 42,240 | **+0.1082 ± 0.0013** |
| richmond | Mapillary | 890 | 7,040 | **+0.1275 ± 0.0058** |
| sao_paulo | GSV | 1,877 | 8,156 | **+0.0572 ± 0.0034** |

All 50–90σ. The discrepancy is not scatter — it is **proportional to range**, the signature of a *scale* error in `d = camera_height / tan(depression)`, at roughly **11%** almost everywhere.

At 20 m that is a ~2.3 m radial displacement, and because a street's cameras mostly sit along one line, it does not average out — it pushes detections consistently away from the roadway.

## This is the same effect §0 of `detection_recall_analysis.md` already saw

That doc reports flat-ground geometry and DA3 metric depth agreeing **to within 6.5–8.5%**, and notes geometry placing four Richmond ramps *above the horizon*. Those were read as agreement plus rig noise. The multi-view result says the residual has **structure**: it is a proportional scale error, not a spread. A scale error is a one-constant fix; noise is not.

Two independent methods — an external monocular depth model, and pure self-consistency across views — now point the same way, and neither shares the other's failure modes.

## What it is not

[...]

PR #115, opened 2026-08-14
Score crop-window sizing rules on manual_gold; find the boxes are not extent gold

jonfroehlich:
## Summary

Builds the crop-window acceptance-test harness for #114 and runs it against the `manual_labels/*.txt` boxes (3,919 boxes / 1,000 panos). The most important result is about the boxes, not the rules. Full report with overlay evidence: `docs/crop_window_eval.md`.

- **Finding 1 — `manual_labels` w/h is NOT object-extent gold.** Median box max-side is 12.8 px on a 4096×2048 equirect; box size tracks distance ~7× slower than a physical ramp's apparent size; only ~9.8% of boxes are plausibly full-apron. Visually confirmed: the typical "box" is a near-point mark on the detectable-warning pad. Center points are unaffected. Consequence: no usable curb-ramp extent gold exists anywhere in the ecosystem, and #83's plan to use these boxes as the SAM2/extent eval set needs a convention audit first (cross-posted there).
- **Finding 2 — the v1 resolution defect is confirmed and quantified.** At 2048-height panos the raw `predict_crop_size` window is **1.97×** the normalized one (median 294 vs 149 px) — direct evidence the CropService port's normalization (ProjectSidewalk/SidewalkWebpage#4865) is load-bearing. Containment of the *marked point* is a non-problem for every rule (≥97.5%); the real acceptance criteria — whole-apron containment and context ratio — wait on real extent gold.
- **Finding 3 — the 10–15% context band is an ML-consumer number.** Padding a full-extent object model to it implies ~70° FOV windows. The human-facing Gallery card's right context level has never been specified anywhere; the acceptance test should set the band per consumer class.
- **Detection coverage at the 0.55 floor:** 87.3% of gold points are covered by an operational detection (those labels would exist and get crops); 190 operational detections match no gold point (would-be FP crops).

## What's in the diff

- `scripts/analysis/crop_window_eval.py` — the harness: parses gold boxes, places windows per CropRunner's geometry (seam wrap, clamp-by-shift, dim cap), scores v1-raw / v1-norm / geo-v1.5, writes summary JSON + per-box CSV, optional overlay gallery (fresh view-only GSV fetches)
- `tests/test_crop_window_eval.py` — 18 tests (full suite: 514 pass)
- `docs/crop_window_eval.md` — the report; evidence images under `docs/assets/`
- `analysis_out/crop_window_eval.json` — committed summary with a content hash of the per-box table

No imagery is needed for any headline number; the gallery is optional. The harness re-runs unchanged once a proper extent-gold set exists, so I'd keep #114 open until the box-annotation pass produces one.


jonfroehlich:
## Deep review — `crop_window_eval.py` + the Round-2 report

Read the harness line by line, re-ran both evals from a clean checkout, and re-derived Finding 4 from the committed per-box table. The engineering is solid and the reproducibility is genuinely good: composing this branch with #116's `boxes.json` and re-running reproduces **both** committed summaries byte-identically (modulo CRLF). Finding 1 and Finding 2 hold up.

**Finding 4 does not, in the form it is written.** The ranking it reports is a scale artifact, and the plan item it feeds ("geo-v1.5 presumptive; SW#4865's planned port is the worst rule for Richmond") inverts once the scale is controlled for. Detail first, because it changes what ships.

---

### 1. The three rules have the *same* scale-free accuracy. Only their scale differs.

The per-box table already contains everything needed to separate "wrong size constant" from "wrong shape." With the window centred on the prompt, `min_margin = s/2 − R` where `R` is the half-side needed to reach the box's furthest extremity, so

```
required_side = predicted_side × (1 − 2 · margin_norm)
```

exact for every unshifted row (0/227 shifted for both v1 rules, 6/227 for geo-v1.5). `required_side` is a property of the *box and the prompt* — it does not depend on the rule. So `predicted_side / required_side` is a scale-free calibration score, and its dispersion is the rule's actual accuracy:

| rule (detection mode, n=227) | pred/req p10 | p50 | p90 | **p90/p10** |
|---|---:|---:|---:|---:|
| v1-raw | 0.55 | 0.95 | 1.79 | **3.23** |
| v1-norm | 0.46 | 0.80 | 1.41 | **3.03** |
| geo-v1.5 | 2.02 | 3.58 | 6.24 | **3.09** |

Identical to within noise — and `v1-norm` is nominally the best of the three, which is the opposite of the report's ranking. The 0.996 vs 0.260 containment gap is **entirely** the p50: geo is 4.5× larger, nothing more.

Sweeping a constant multiplier `k` (with the scorer's own `min(k·s, pano_w, pano_h)` cap applied):

| rule | k | containment | ctx p50 | side p50 | capped |
|---|---:|---:|---:|---:|---:|
| v1-raw | ×2.5 | 0.978 | 0.350 | 850 | 0 |
| v1-raw | ×3.0 | **1.000** | 0.291 | 1020 | 0 |
| v1-norm | ×3.5 | **0.996** | 0.290 | 1064 | 0 |
| v1-norm | ×4.0 | **1.000** | 0.254 | 1216 | 0 |
| geo-v1.5 | ×1.0 | 0.996 | 0.229 | 1458 | 0 |

`v1-norm × 3.5` matches geo-v1.5's containment exactly (0.996) with a **27% smaller median window**. `v1-raw × 3.0` reaches 1.000 with a 30% smaller one. Geo buys more *context* per contained ramp, not more containment — and whether that extra context is wanted is precisely the per-consumer question Finding 3 says is unanswered.

So the conclusion I would draw from this data is nearly the reverse of the one in the doc:

- **The resolution normalization in ProjectSidewalk/SidewalkWebpage#4865 is fine.** It is the best-calibrated of the three rules on the only real extent gold that exists. Its constant is ~3.5–4× too small for whole-apron containment. That is a one-line change to the planned port, not "ship geometry instead."
- The 2013 formula being "object-sized, not window-sized" is right and is the useful half of Finding 4 — but it is a statement about one multiplicative constant, not about the formula's shape.
- **The number that actually deserves the headline is the one no rule beats: `p90/p10 ≈ 3` on all three.** Nothing is within ~1.7× of the right size at either tail, at any scale. That residual is the entire case for an extent-aware rule (SAM2, #83), and it is invisible in the current table because containment and ctx are both monotone in `k`.

Concretely, I'd suggest: add `required_side` and `k = required/predicted` as CSV columns, put the `p90/p10` dispersion in `_stats`, and add a `--rescale-sweep` so the k-table above is a first-class output. Then Finding 4 can say the honest thing — "all three rules are equally accurate in shape; pick a scale constant, and here is the containment/context curve" — which is also a much easier thing for #4865 to act on.

### 2. geo-v1.5 lands at 1.83× its own target, and the doc's apron-size sanity check does not reconcile with it

Measured ctx p50 is 0.229 against a `GEO_TARGET_RATIO` of 0.125 — 1.83×, and 1.53–1.77× within every distance band, so it is not a tail effect. Only 2/227 windows touch the pano-dimension cap, so it is not the clamp either.

The report explains this with "real aprons measure ~1.2–1.3× the 1.5 m nominal," but 1.25× predicts ctx = 0.156, not 0.229. Those two numbers come from the same script and disagree by 1.5×. Worth showing the derivation of the 1.2–1.3× figure — I get 1.5–1.8× from the committed CSV by every route I tried.

The most likely culprit is `GEO_CAM_H_M = 2.5`. Richmond is Mapillary; at ~1.7 m the two numbers reconcile exactly (0.125 × 1.25 × 2.5/1.7 = 0.23). That matters beyond bookkeeping: if the height is wrong, geo-v1.5's *only* free parameter is absorbing it, and the "12.5% target" knob is not doing what it says. A `--cam-height` sweep would settle it in one run, and it ties straight into ProjectSidewalk/sidewalk-auto-labeler#40 / #101 — per-pano camera height is now measurable on GSV, which removes the parameter rather than retuning it.

### 3. The gold says aprons are ~3.4:1, but every candidate cuts a square

Recovering box height from `road_margin_ratio` in gold-center mode (`box_h = side / (2·road + 1)`), over all 299 boxes:

| | p10 | p50 | p90 |
|---|---:|---:|---:|
| box max side (px) | 128 | 270 | 693 |
| box height (px) | 31 | 75 | 242 |
| aspect (max/h) | 1.97 | **3.44** | 5.92 |

Containment is therefore decided by horizontal extent alone, and a square window sized to contain the width over-provisions vertically by ~3.4×. At geo's ctx p50 of 0.229 the *vertical* context ratio is ~0.067 — roughly 15 box-heights of sky and sidewalk. That is most of the "enormous windows" complaint in Finding 3, and it is a property of the square-window assumption rather than of any sizing rule. Reporting horizontal and vertical context ratios separately (and scoring a non-square candidate) would be cheap and would change what #4865 should cut.

### 4. The planned GSV arm cannot run the comparison the doc says it will

Two independent blockers, both worth knowing before the annotation session:

**(a) manual_gold is 4096×2048, not native GSV.** All 1,000 records carry `4096×2048` (checked). So `--bundle benchmark/manual_gold` scores at `pano_h = 2048` — *below* the 6656 calibration height, the same regime as Richmond's 5500/2880/2048. The doc's framing —

> GSV native panos (8192/16384) sit above it, where the raw formula *over*-sizes instead

— is not what that arm will measure. Probing the >6656 regime needs records carrying native GSV dimensions (and imagery to match, per the note on #116).

**(b) `gold:` items can never be detection-prompted.** `load_bundle_boxes` sets `is_detection = key.startswith("det:")`, and `box_gallery --from-manual-labels` emits `gold:<i>`. So the GSV arm produces gold-center rows only: detection mode gets `n = 0` and `summarize` silently omits the whole mode, while every item lands in the `missed_only` counter and the header reports "missed (gold-center only): 3919". Since the Richmond headline is the det-prompted table, the two arms will not be comparable.

The manual-mode `match_detections` already does exactly the needed matching (greedy, 0.022 radius, 1024/512) — running it for `gold:` items and recording the matched point would close this. Otherwise the arm needs an explicit "gold-center only, not comparable to Richmond's headline" caveat.

### 5. The v1 sign convention is the load-bearing detail and nothing pins it

`_v1_reference_size(pano_h/2 − y_px)` is the whole port. `test_v1_raw_pinned_values` pins the implementation against its own output, so a flipped convention passes all 22 tests and inverts Finding 4. The direction is physically right (nearer → larger window), so I do not think it is wrong — but a permalink to `CropRunner.predict_crop_size` **and its call site** in the constants comment would make the port auditable by someone who is not re-deriving it. Same for the `elif`→`if` change in the clamp (equivalent as written; worth saying so).

---

### Smaller things

- **The scorer cannot tell an incomplete `boxes.json` from a complete one.** `load_bundle_boxes` reports `n_boxed + n_cant` over whatever is in the file, so an abandoned session silently shrinks the denominator with no warning. `verdicts.json` sits in the same directory — cross-checking the adjudicated count (237 `True` dets + 73 sure missed = 310 = 299 + 11 ✅ for Richmond) would make that loud.
- `run_bundle_eval`: a `det:` item with no recorded `point` falls into the `missed_only` bucket and quietly loses its detection-mode row. Worth a third counter rather than folding it into a mislabelled one.
- In gold-center mode `road_margin_ratio` reduces to `(side/box_h − 1)/2` — a deterministic function of the window side and the box height, carrying no placement information. It only becomes evidence in detection mode; the gold-center "road p50" column reads like an independent measurement and isn't one.
- `analysis_out/*.json` is written in text mode, so a Windows re-run dirties the tree with pure CRLF churn. `newline="\n"` in `write_outputs` makes the committed artifacts platform-stable.
- Merge order: this PR commits `analysis_out/crop_window_eval_richmond.json`, whose declared input `benchmark/richmond/boxes.json` only exists on #117. Merging #117 first avoids a window where main holds a summary of a file it does not have.

### Verified

- `pytest tests/test_crop_window_eval.py` → 22 passed.
- Re-ran both evals from a clean checkout of this branch (with #117's `boxes.json` copied in): `analysis_out/crop_window_eval.json` and `analysis_out/crop_window_eval_richmond.json` both reproduce byte-identically. The Round-2 table in the doc matches the regenerated output exactly.
- `benchmark/manual_gold/records.jsonl`: 1000/1000 rows at 4096×2048.
- `benchmark/richmond/verdicts.json`: 124 panos, 0 partially judged, 237 `True` dets + 73 sure missed.

jonfroehlich:
## Review fixes applied — 80edc05

All six findings. 32 tests here (from 22), 528 overall; both committed summaries regenerated.

### 1. Finding 4 rewritten — the ranking was a scale artifact

`required_side` (the side box and prompt actually demand — a property of the gold, not of any rule) is now a per-box column, and `size_ratio = predicted / required` is in the summary. Detection mode, n=227:

| rule | containment | size ratio p10 / p50 / p90 | **p90/p10** |
|---|---:|---:|---:|
| v1-raw | 0.449 | 0.55 / 0.95 / 1.79 | **3.24** |
| v1-norm | 0.260 | 0.46 / 0.80 / 1.40 | **3.02** |
| geo-v1.5 | 0.996 | 2.03 / 3.59 / 6.26 | **3.08** |

Equally accurate; one constant apart. The new rescale sweep (on by default, re-scored with the real window geometry incl. the pano-dim cap) puts them at matched containment:

| rule | k for ≥99.5% | ctx p50 | side p50 |
|---|---:|---:|---:|
| v1-raw | ×3.0 | 0.291 | 1020 px |
| v1-norm | ×3.5 | 0.291 | 1063 px |
| geo-v1.5 | ×1.0 | 0.229 | 1458 px |

So the recommendation for ProjectSidewalk/SidewalkWebpage#4865 inverts back to: **ship the normalization, give it a scale constant.** Upstream's own docstring settles the provenance question too — `old_pano_y` "converts pano_y and pano_height to the OLD version of pano_y that we had when this alg was written", and that conversion is precisely what is missing, so v1-norm is the faithful port rather than a variant of it. The headline is now the spread: nothing is within ~1.75× of the right size at either tail, at any scale, which is the real case for #83.

### 2. geo-v1.5's 1.83× overshoot was the camera height

Not the apron-size claim. `--cam-height 1.7` (Mapillary, vs GSV's ~2.2–2.5) lands ctx p50 at **0.156** — exactly what the gold's 1.2–1.3× apron measurement predicts against the 0.125 target. The two numbers reconcile; 2.5 m was over-estimating distance and undersizing the window. Height is a flag now and is recorded in the summary. The doc also says plainly that only the *product* of camera height and nominal ramp width is identified from richmond alone.

### 3. Aspect is measured and the context ratio is split per axis

Median box aspect 3.29 (p10–p90 2.0–5.9). Containment is decided by width alone, and at geo's ctx p50 0.229 the vertical ratio is 0.068 — ~15 box-heights. Most of Finding 3's "enormous windows" is the square-window assumption; `ctx_h`/`ctx_v` are now reported so a non-square candidate can be scored against the same gold.

### 4. The GSV arm

- `gold:<i>` items are matched to the record's operational detections (same greedy 0.022 match), so that arm produces a det-prompted table instead of silently falling back to gold-center-only.
- The doc no longer claims it will probe the >6656 regime. All 1,000 manual_gold records are 4096×2048, so it scores below the calibration height in the same regime as richmond — and annotates at 1024 px. Both stated up front rather than discovered mid-session.

### 5. Provenance pinned

The v1 constants now carry upstream's actual `predict_crop_size` body, its call site, and the two issue threads the fit came from. The pinned test asserted the port against its own output, so a flipped sign convention would have passed all 22 tests and inverted the conclusion. (It is correct — `old_pano_y = pano_height / 2 - pano_y`.)

### 6. Smaller

Completeness cross-check against `verdicts.json` (richmond: 299 + 11 = 310 ✅) so an abandoned session can't silently shrink the denominator; a `det:` item with no recorded point gets its own counter instead of being misreported as a model miss; the gold-center `road p50` column is annotated as `(side/box_h − 1)/2` by construction rather than reading as independent evidence; `newline="\n"` on both writers so a Windows re-run stops dirtying the tree with CRLF churn.

Merging after #116/#117, so `benchmark/richmond/boxes.json` is on main before the summary that names it.

Issue #43, opened 2026-07-24
360 pano seam: curb ramps split across the left/right edge get double-counted in GT scoring


jonfroehlich:
## Problem

A 360° equirectangular pano has a hard seam at its left/right edge (x≈0 and x≈1 are the
same physical location — the pano wraps there). A curb ramp that straddles that seam gets
bisected: one half at the far right, one at the far left. This creates a ground-truth
scoring hazard that the current verdict schema and viewer don't guard against.

**First real case — Clovis `150809456962552` (captured 2019-07), benchmark/clovis:**

- The model emitted `det1` at **x=0.9883**, y=0.586, conf 0.580 — the right-hand half of a
  curb ramp on the corner directly behind the car.
- Reviewing the *left* edge (x≈0.003), the same corner's other half looks like an
  undetected ramp, so the natural move is to drop a **missed** mark there.
- Result: the same physical ramp is scored as **both a true positive** (det1 correct)
  **and a false negative** (the missed mark). Incoherent, and it docks recall for a ramp
  the model actually found. (The other detection on this pano, det0 at x=0.3545, and a
  separate genuine miss at x≈0.28, are unaffected — those aren't near the seam.)

## Why it's easy to miss

The GT viewer (`scripts/gt_gallery.py`) treats the pano as a flat, hard-edged image:
`clampPan()` clamps horizontal pan to `[w − w·zoom, 0]`, so you **cannot pan across the
seam**. The two halves appear at opposite ends of the strip with no visual cue they're the
same corner. The model sees the same hard seam (no wrap-padding on its 4096×2048 input), so
this is a shared artifact of how the pano is cut, not just an annotation slip.

## Scope (what this is and isn't)

- **GT scoring — the real issue.** The schema auto-dedups redundant *detections* (the
  `duplicate` verdict), but there is no dedup between a *detection* and a *missed mark*
  (different categories), so the seam case relies entirely on reviewer discipline.
- **Production / PS submission — fine, no action.** A single detection at x≈0.99 submits
  normally, and even a both-halves double-detection projects to ~the same geographic point,
  so PS clustering absorbs it (same as any duplicate label).
- **Model limitation — separate thread.** Seam-straddling ramps are genuinely harder for the
  model (no horizontal wrap). Worth tracking as a model/pipeline question (wrap-pad the input,
  or detect on a seam-centered rotation and merge), but out of scope here.

## Reviewer rule (mentally wrap the seam)

[...]

PR #117, opened 2026-08-14
Add box_gallery.py: whole-apron extent annotation over adjudicated verdicts

### LINKED ISSUE #116: box_gallery.py: second-pass whole-apron box annotation over adjudicated verdicts (extent gold)
## Why

#114 measured that the `manual_labels/*.txt` w/h are near-point marks, not object extents (median max-side 12.8 px; ~90% of boxes; see `docs/crop_window_eval.md` on #115). Consequence: **no usable curb-ramp extent gold exists anywhere in the ecosystem**, so neither a crop-sizing rule (the Richmond ship gate) nor an extent model (#83''s SAM2 path) can currently be validated. The gap is both providers: richmond''s 124 benchmark panos (Mapillary, ~388 confirmed points) and a manual_gold subset (GSV, re-annotation with an explicit box rule — the existing near-point marks are good prompts to box around).

## Design: a separate second-pass tool, not a mode in gt_gallery

`scripts/box_gallery.py`, sibling of `gt_gallery.py`, consuming a finished `verdicts.json` and writing a separate `boxes.json` sidecar. Rationale:

1. **Sequencing** — boxes are drawn around *adjudicated* ramps (TP det verdicts + sure missed marks), which only exist after point review completes.
2. **Protocol integrity** — nine cities of verdicts were produced under the current point protocol; a drawing mode inside the verdict UI risks bias (extent ambiguity leaking into existence judgments) and protocol drift for future cities.
3. **Schema** — `verdicts.json` is sync-locked to `rampnet.validation.collect`; boxes live in `boxes.json` keyed by `(pano_id, det:<idx> | missed:<idx>)`, so validation code never changes and point-only cities stay first-class.
4. **Resolution rules differ** — point review deliberately renders at model resolution (the #26 recall-fairness rule); box drawing *wants* native resolution for tight edges, which is legitimate because extent annotation is not a recall judgment. Two tools make each rule load-bearing.

## Requirements

- **Native-res, crop-centered view per ramp**, target point pre-marked, sibling points/boxes shown dimly; drag-to-rect + handle editing; zoom to native 1:1 and beyond (magnification only).
- **Equirectangular seam** (#43): the crop wraps x around the pano edge, so a seam-split ramp is drawable as one box; boxes export **pano-normalized** `{cx, cy, w, h}` (resolution-independent, cx may wrap).
- **Blind by construction**: the tool never renders a predicted crop window — the annotator already sees the point, so a window would leak exactly a size prior, and size is the quantity under measurement. The algorithm-vs-gold eyeball belongs in #114''s overlay gallery, not here.
- **An explicit whole-apron BOX_RULE in the UI**, versioned and embedded in the exported `boxes.json` — so the new gold cannot reproduce the convention drift #114 found.
- Escape hatch per ramp: "can''t determine extent" (occlusion/blur), exported as a status rather than a guessed box; auto-flag boxes that touch the crop edge (apron may extend beyond the view).
- `--from-manual-labels` mode: same viewer over a sampled manual_gold pano subset, items from the near-point marks (`gold:<idx>`).
- Autosave to localStorage; prefill from an existing `boxes.json` for revision — same round-trip contract as gt_gallery.

jonfroehlich:
Closes #116. The second pass that produces the extent gold #114 showed is missing everywhere: a sibling of `gt_gallery.py` that turns adjudicated ramp points into whole-apron boxes.

## What it does

- Reads a finished `verdicts.json` and enumerates one item per adjudicated ramp — `det:<i>` for detections judged correct, `missed:<i>` for sure missed marks — mirroring `rampnet.validation.collect`''s semantics (False isn''t a ramp, `unsure` abstains, `duplicate` is a ramp another item already covers). Verdict/records mismatches are skipped with the same warning collect() gives.
- Cuts a **native-resolution** square crop per ramp (default 90° of pano longitude), stitched across the equirectangular seam (#43) so a seam-split ramp is drawable as one box. Geometry comes from the records'' stored dims, and the on-disk image must match exactly — a non-native image fails loudly rather than annotating at the wrong scale.
- The viewer: drag-to-draw, 8-handle resize, move; Space/middle/right-drag pans; wheel zoom with a pill showing distance from native 1:1. The adjudicated point is a crosshair; the pano''s other ramps show as dim dots with their boxes as dashed outlines, so a shared apron never gets boxed twice. Per-ramp "Can''t determine extent" and note. Autosave to localStorage; an existing `boxes.json` prefills for revision.
- **Blind by construction**: no predicted crop window is ever rendered — the annotator already sees the point, so a window would leak exactly a size prior, and size is what this gold measures.
- **The box convention is explicit and versioned** (`BOX_RULE` v1, shown in the viewer and embedded verbatim in every export) so the new gold cannot reproduce the convention drift #114 found in `manual_labels`.
- Exports **pano-normalized** `{cx, cy, w, h}` (resolution- and fov-independent; a box may wrap the seam) to a separate `boxes.json` keyed `(pano_id, det:<i>|missed:<i>)` — the point protocol and validation code stay untouched. Boxes touching a non-pano crop edge auto-flag (`edge_flag`) since the apron may continue beyond the view; `--fov 120` re-renders wider while keeping every annotation.
- Prefill reconciliation: `boxes.json` stores each ramp''s prompt point, so a revised verdicts.json that renumbers `missed:<i>` keys can''t silently re-attach a box to a different ramp — mismatches drop loudly.
- `--from-manual-labels [DIR] --sample-panos N`: the manual_gold re-annotation mode, near-point marks as prompts, `gold:<i>` keys.

## Verified

- 13 new tests (`tests/test_box_gallery.py`); full suite 514 pass.
- Rendered the real Richmond gallery: **310 adjudicated ramps on 92 panos**, zero skips (every archive image matched its record''s native 12288×6144), 3072 px crops, 221 MB — ready to annotate.
- Viewer JS syntax-checked under node, and the seam-wrap geometry round-trips numerically (crop crossing the pano edge: point → pano → crop is identity; a rect straddling the seam exports `cx≈0.004` and reimports stably).


jonfroehlich:
## Deep review — `box_gallery.py`

Read the Python and the viewer JS line by line, re-ran the tests (13 pass), and re-rendered the geometry by hand. The design calls are right: separate second-pass tool rather than a mode in `gt_gallery.py`, blind-by-construction (no predicted window rendered), versioned rule embedded in every export, pano-normalized boxes so a `--fov` re-render keeps the work, native-res crops with a hard dimension check. The gold it produced is real and it reproduces (see the note on #114 below).

Five things to fix, one of them a genuine data-loss path.

---

### 1. Prefill merges per-**pano**, not per-**key** — `boxes.json` entries can be silently deleted on export

```js
let state = JSON.parse(localStorage.getItem(STORE) || '{}');
for (const pid in INITIAL) if (!(pid in state)) state[pid] = INITIAL[pid];
```

If localStorage holds *any* key for a pano, every other key `boxes.json` carries for that pano is skipped. That is not just a display problem — those boxes are then **dropped from the next export**:

- export loop 1 walks `ENTRIES` and does `const s = itemState(e); if (!s || !s.status) continue;` → the un-merged keys are `undefined` and get skipped;
- the round-trip loop below it only rescues keys present in `state` *and* absent from `ENTRIES`, so it does not catch them either.

Net effect: open the viewer in a browser that has partial local state, hit Export, and the download is `boxes.json` minus the other person's (or the other session's) boxes for every pano you had touched. Single-annotator so far, so it has not bitten; it will the first time a second annotator's file is merged and anyone re-renders.

Fix is a per-key merge:

```js
for (const pid in INITIAL)
  for (const key in INITIAL[pid])
    if (!(state[pid] || {})[key]) (state[pid] = state[pid] || {})[key] = INITIAL[pid][key];
```

### 2. The stale-box guard is defeated end-to-end

`reconcile_initial`'s docstring promises "a stale box does NOT survive the next export unless redrawn." Two independent holes mean it usually does:

- **It only filters `INITIAL`.** The revision path that matters — same annotator, same browser — reads the box from localStorage, which reconciliation never sees. Any box already in local state survives a renumbered `missed:<i>` untouched.
- **The export re-stamps the point from the current entry**: `const rec = {point: {x: e.x, y: e.y}, status: s.status}`. So the exported `point` is whatever the *new* verdicts.json says, not the point the box was drawn against. Once a stale box round-trips once, it matches forever and the check can never fire again.

Fix: record the prompt point in state at draw time (in `rectState`, and on the "can't determine" path) and export *that*; then run the same tolerance check against localStorage entries on load, not only against `INITIAL`. Without both halves the sidecar's `point` field is decorative.

### 3. `enumerate_items` does not mirror `collect` on partially judged panos

`rampnet.validation.collect` drops a pano **entirely** when any det verdict is `None` ("partially judged: unusable for either metric"). `enumerate_items` has no such skip, so a half-reviewed pano still contributes its `True` dets and its sure missed marks — including missed marks on a pano nobody finished scanning.

Latent, not active: I checked `benchmark/richmond/verdicts.json` — 124 panos, **0 partially judged**, so the shipped gold is unaffected. But the docstring claims a mirror that the code does not implement. Either add the `any(d is None …)` skip or narrow the docstring to "the what-is-a-ramp side of a fully judged pano."

### 4. A note typed before a box exists is silently discarded

The `#itemnote` handler writes `{note: …}` with no `status`; the export skips anything without `s.status`. So "can't tell yet, here's why" — exactly the note you want on a hard item — persists to localStorage, shows in the UI, and then vanishes from the file. Either export note-only items, or don't accept a note until the item has a status.

### 5. `--from-manual-labels` will not be native resolution — and that is the module's stated reason to exist

Every one of the 1,000 rows in `benchmark/manual_gold/records.jsonl` is `4096×2048` (checked). So the GSV arm gets `crop_side(4096, 2048, 90) = 1024 px` crops — a third of Richmond's 3072 px — and `render_pano_items`' native check will *reject* a true-native GSV pano because the record says 4096×2048.

The docstring's justification ("crops here are cut at **native resolution**: extent annotation is not a recall judgment, and tight edges need real pixels") does not hold for that arm as it will actually run. Under a rule that says "every edge should touch the ramp's boundary at native zoom," drawing at model resolution is a different instrument. Worth deciding before the session: native-res manual_gold imagery + records that say so, or an explicit "GSV gold is drawn at 4096×2048" caveat carried into the comparison. (This also lands on #114 — see the comment there.)

---

### Nits

- `renderLight` rebuilds the progress counter **without** the `(N can't)` suffix that `render` adds, so the can't-determine count blinks out on every state change and comes back on navigation.
- `#cropview` is `height: min(76vh,880px)` + `aspect-ratio: 1/1` + `max-width: 100%`, and `#croplayer img` is `width:100%;height:100%`. On a narrow window `max-width` wins and the crop is stretched anisotropically, while `updatePill` reports the horizontal ratio only. A stretched view under a "tight at native zoom" rule biases edges quietly. `width: min(76vh, 880px, 100%)` with height derived, or a warning when `r.width !== r.height`, closes it.
- `--from-manual-labels` on a missing/empty labels dir exits with "No adjudicated ramps to annotate," which does not point at the actual problem.
- `items_from_manual_labels` parses all 1,000 label files before `--sample-panos` throws away 95% of them. Cheap, just noting it.

### Verified

- `pytest tests/test_box_gallery.py` → 13 passed.
- `boxes.json` is internally consistent with `verdicts.json`: 237 `True` dets + 73 sure missed = 310 adjudicated; the export carries 299 boxed + 11 can't = 310. ✅
- Composed this branch's `boxes.json` with #114's scorer and re-ran `crop_window_eval --bundle benchmark/richmond`: the summary committed on #114 reproduces **byte-identically**. Worth merging this PR **before** #114, since that PR commits a summary whose named input (`benchmark/richmond/boxes.json`) only exists here.

jonfroehlich:
## Review fixes applied — 3fc4db8

All five findings plus the nits. Suite: 18 box_gallery tests (from 13), 514 overall, viewer JS `node --check` clean.

**1. Prefill merged per-pano → per-key.** One locally-touched ramp no longer suppresses the rest of a pano's prefill, so Export can't write `boxes.json` minus another session's boxes.

**2. The stale-box guard now works on the path that matters.** Annotations carry `px`/`py` — the prompt point they were made against. Export writes *that* as `point` (re-stamping from the entry's current position was laundering stale boxes past every future check), and load reconciles local state too, not just the prefill. Mismatches drop with a banner; annotations predating the stamp are adopted at their current point rather than destroyed, and counted separately so they can be spot-checked.

The bootstrap is extracted as `STATE_BOOTSTRAP_JS` and tested under node — it is the one piece of viewer JS that can lose gold, and both failure modes were silent.

**3. `enumerate_items` skips partially judged panos**, which is what `collect()` does. Latent on richmond (0 such panos), not latent in general.

**4. Notes on unboxed ramps survive** — they get `status: "note"` instead of being dropped at export.

**5. Resolution is reported, not assumed.** New `resolution_note` prints what the bundle actually annotates at and embeds it as `crop_px_by_pano_dims`. It turns up something worth knowing beyond the manual_gold case: **9 of richmond's own 92 panos are 4096×2048**, so part of the shipped gold is model-resolution too —

```
Crop resolution: 11000x5500 -> 2750 px (64 panos), 12288x6144 -> 3072 px (10 panos),
                 4096x2048 -> 1024 px (9 panos), 5760x2880 -> 1440 px (9 panos)
```

Backfilled into the committed richmond gold so the one export that exists carries the same provenance a fresh one would.

**Nits:** `#cropview` sized from width so it can't go non-square and stretch the crop while the pill still reads native 1:1; one progress-counter definition (the can't-determine count stopped blinking out); `--from-manual-labels` fails loudly on a missing/empty dir and samples panos before parsing 1,000 label files.

Merging this before #114, as flagged — that PR's committed summary names `benchmark/richmond/boxes.json` as its input.

Issue #58, opened 2026-07-25
Add the 1k manual gold-standard set as a model-comparison split (bigger, independently-labeled, un-anchored GT)


jonfroehlich:
## Motivation

The model-comparison harness (#34 / #36 / #40) scores RampNet vs the VLM/open-detector field on the richmond/bend/clovis **verdict** GT. That GT is deliberately deployment/OOD-focused, but it is explicitly **"RampNet-anchored"** — derived from a review of RampNet's *own* detections plus a missed-ramp pass (the `docs/model_comparison.md` caveat, and the bias behind #55).

We already have a **bigger, independently-labeled, un-anchored** gold set and aren't using it in the comparison: the **1,000-panorama Manual Ground Truth Set** (3,919 curb ramps). It is:

- **Fully and manually labeled from scratch** by humans (not seeded by RampNet) → **no RampNet-anchoring**, so it sidesteps #55 entirely for this data and is *fairer to the VLMs too* (their recall isn't measured against a GT built from RampNet's hits).
- **4× the size** of the current benchmark (1,000 vs 234) → tighter Wilson CIs.
- The **paper's canonical gold standard** — `README.md` (§"Evaluating") says we "place more emphasis on [it]… less prone to errors… directly from a human source."

## Where the data is

- **Images:** in `projectsidewalk/rampnet-dataset` on HF (the manual-eval panos are held out from train/val, so they live in the `test` split — see README data-leakage note).
- **Labels:** in-repo at `manual_labels/*.txt`, YOLO format (`class cx cy w h`, normalized), keyed by pano id.

## Important caveat — it's in-distribution (GSV)

The gold set is GSV, sampled from cities including some in training (NYC/Portland/Bend; `docs/data_provenance.md` flags in-training cities as optimistically biased). It is held out from Stage-2 *training*, but it's the same imagery source + overlapping geography RampNet was built for, so **RampNet has home-field advantage** and its margin here will likely be *larger* than on OOD Mapillary. Frame it as an **in-distribution reference that complements** the OOD cities (richmond/clovis), not a replacement — don't over-read an in-distribution win.

## What it takes (harness is split-generic)

- [ ] Fetch the 1,000 gold-set images from `projectsidewalk/rampnet-dataset` (by pano id from `manual_labels/`) into a `benchmark/manual_gold/` split (`panos/` git-ignored like the cities).
[...]

PR #118, opened 2026-08-14
fetch_manual_gold: --images-only, so re-fetching imagery stops requiring --force (#58)

jonfroehlich:
The manual_gold bundle conflates two products with opposite lifecycles: the imagery is git-ignored and per-machine, while `records.jsonl` is committed and carries the detections `export_gold_records.py` wrote into it. The most common operation — *new machine, records already committed, just fetch the panos* — was exactly the one the script could not do safely: the guard offered only "refuse" or `--force`, and `--force` rebuilds `records.jsonl` (discarding the detections) plus `bundle_meta.json`, leaving a dirty tree someone has to know to `git checkout`. That dance happened for real on 2026-08-14 while staging the pano-trio benchmark eval on makelab2 (#51): the fetch ran under `--force` and the two committed files were then restored by hand.

What this adds:

- **`--images-only`** fetches into `panos/` and touches nothing committed. It cross-checks the fetched ids against `records.jsonl` and refuses on drift (a label/record mismatch means the bundle needs a rebuild, not an image fetch).
- **The guard now steers to `--images-only`** for the usual case instead of funneling everyone to the destructive flag.
- **Source-mismatch refusal, before any download**: an `--images-only` fetch whose `--source` contradicts `bundle_meta.json`''s recorded source is rejected — the two sources are not byte-identical, and a JPEG re-encode alone moved the gold numbers P +2.2 / R −1.8, so imagery from the other source under the committed records would silently change what gets scored.
- **Measured cost note in the docstring**: on makelab2 the `hf` path''s `load_dataset` downloaded and arrow-materialized **all three splits** (~2.5 h end-to-end) despite `split="test"` — not the ~44 GB test-only fetch the docstring used to promise. A shard-scoped fetch via `HfFileSystem` + pyarrow (the `--audit` path already works that way) is named as the follow-up and deliberately **not** done here, to keep this change out of the byte-fidelity-sensitive read path.
- 7 new tests (`tests/test_fetch_manual_gold.py`), all through `--source local` against tmp fixtures — no network, in keeping with the suite''s rules. 508 pass locally.


jonfroehlich:
## Deep code review of this PR, and the fixes for what it found

A high-effort review of this branch turned up one blocking defect and ten smaller ones. All eleven are fixed in 9892bf8; main is merged in 7edc067. The test file goes from 7 tests to 33, and **21 of the new tests fail when run against the pre-fix `fetch_manual_gold.py`** — checked by re-running them against the previous commit, so each new behaviour is genuinely pinned rather than vacuously green. 1,036 tests pass on the merge.

### Blocking: the automated path still called `--force`

`scripts/run_gold_bundle.slurm` — the launcher this PR exists to fix — was never touched. On a fresh clone `records.jsonl` is committed while `panos/` is git-ignored, so `n_panos` is 0, the `records.jsonl && n_panos >= 1000` test fails, and the `elif` branch ran `fetch_manual_gold.py --force`: detections discarded, `bundle_meta.json` rewritten, dirty tree. That is exactly the 2026-08-14 makelab2 incident this PR describes. The in-file comment justifying it — "detections only exist after step 2, and a bundle that got that far is complete and never reaches this branch" — stopped being true once detections started arriving via git rather than only via step 2.

It now runs `--images-only`. The `n_panos >= 1000` skip is gone with it: `--images-only` checks each pano already on disk against `records.jsonl`, which is strictly better than a file count that a 999-pano directory passes silently.

### Blocking: the source guard no-opped when the source was unknown

`built_source = meta.get("source")` followed by `if built_source and ...` meant a missing `bundle_meta.json`, or one without a `source` key, fell straight through and the fetch was accepted. The check was waived precisely when the built source could not be established — the case it exists for. A missing `records.jsonl` three lines above already hard-refused, so this was also internally inconsistent.

An unknown or unparseable source now refuses, with the `git checkout` that restores the committed file.

### Integrity of the fetched imagery

- **The size cross-check was computed and discarded.** `records = dict(rows)` materialised the real width/height from `image_dims()`, and the images-only branch threw the dict away. The committed records already state 4096x2048 per pano, so the comparison is free; it now runs, and a disagreement refuses naming the ids and both sizes. Same ids with different pixels is different imagery, and scoring it would move the gold numbers silently.
- **The sha256 manifest is now wired in.** `scripts/analysis/imagery_manifest.py` already writes and verifies exactly this for the nine city splits, and says so in its own docstring: *"someone who downloads the archive has no way to confirm they received the same imagery the ground truth was built on."* That is what `--images-only` is. The fetch now verifies against `benchmark/manual_gold/imagery_manifest.json` by reusing that module, rather than growing a second, special-cased notion of "the right bytes" here. `imagery_manifest.py`'s `ALL_SPLITS` import moved into `main()` so importing it does not drag rampnet's eval stack into a download script's environment.

  **`benchmark/manual_gold/imagery_manifest.json` does not exist and is not written here** — it needs one run on a machine holding all 1,000 panos. When it is absent the fetch says so and prints the exact command that creates it, and the gap is stated in both docs beside the numbers it affects. It is filed as a follow-up in `docs/model_comparison.md`. Until then this split — the largest, at 1,000 panos — is the only one whose imagery has no committed content hash.

### Resume, crashes, and flag handling

- **No skip-existing on the path that is now the usual one.** A second `--images-only` run re-fetched all 1,000 panos, and this PR measures the hf path at ~2.5 h, so a preempted Slurm job restarted from zero. Panos already on disk and consistent with `records.jsonl` are now skipped; absent, unreadable (a truncated write from a killed job), or wrong-sized ones are re-fetched rather than trusted. `--refetch` overrides the skip. With nothing left to fetch the fetcher is not entered at all.
- **Two uncaught crashes.** `fetch_local` unconditionally opened `<pid>.json` — metadata the images-only path discards — so imagery rsync'd without the sidecars died on a bare `FileNotFoundError` partway through the copy. `--images-only` no longer reads them at all; a full build still needs them and now exits with guidance. `record_ids` indexed `["pano"]["panorama_id"]` raw, so a malformed or legacy line gave a `KeyError` traceback where every other guard in the script exits with "rebuild it".
- **Duplicate ids were silently collapsed** by the set comprehension, letting the drift check pass on an internally inconsistent bundle. `record_panos` treats them as an error.
- **`--audit --images-only` silently ignored `--images-only`** and exited 0, while `--images-only --force` went out of its way to reject the equally contradictory pair. Both `--audit --images-only` and `--audit --force` are now refused; so is `--refetch` outside `--images-only`.
- **`fetch_local` read each whole panorama into memory to get its dimensions**, after `shutil.copyfile` had just streamed the same bytes — GBs of needless I/O across the set, discarded outright on the images-only path. Header-only now, via `image_dims_path`.

### Docs

`benchmark/README.md` and `docs/model_comparison.md` both still documented a bare `python scripts/fetch_manual_gold.py` for the imagery, which now hits the guard and exits on a fresh clone; neither mentioned `--images-only`; and both repeated the "~44 GB test split" figure that this PR's own docstring measures to be wrong. The corrected measurement was living only in a Python docstring. Both now carry the `--images-only` command, the measured ~2.5 h all-three-splits cost, and the missing-manifest caveat — in the same document as the numbers they affect. The shard-scoped `HfFileSystem` + pyarrow fetch and writing the manual_gold manifest are recorded as follow-ups rather than left implicit.

### One finding closed by documentation, not code

`--source local` can never be combined with `--images-only` for this bundle, because the committed `bundle_meta.json` records `source: "hf"`. That refusal is correct and stays — an escape hatch would reintroduce the exact byte-fidelity hazard the guard exists for (a JPEG re-encode alone moved the gold numbers P +2.2 / R −1.8). The docstring now says plainly that `--source local` is a full-build option with no `--images-only` counterpart here, which it previously left for the user to discover from an error message.

### Verification

- 33 unit tests, all through `--source local` against tmp fixtures — no network, in keeping with the suite's rules. New coverage includes the full-build and `--force` paths, which were previously untested even though the diff inserted an early `return` immediately above the code that writes `records.jsonl` and `bundle_meta.json`; and `bundle_meta.json` byte-equality is now asserted on every error path, not just the happy one.
- The three new guards were smoke-tested against the real committed 1,000-pano bundle: source mismatch, `--refetch` without `--images-only`, and `--audit --images-only` all refuse before any network call.

PR #119, opened 2026-08-15
Record what every paid-model run costs, at the time it runs

jonfroehlich:
**Standing rule this implements:** every experiment that spends API money records what it spent, as it runs, so papers can report cost alongside accuracy.

## What this adds

1. **Run-time recording** — `GeminiDetector` accumulates each call''s token counts from the SDK''s own `usage_metadata` (thinking tokens fold into output, matching how they''re billed). After each paid-model run, `compare.py` prints the usage + estimated cost and appends a JSONL record to `--usage-log` (default `<cache-dir>/usage_log.jsonl`, `--usage-log none` to disable). Cached panos make no call, so the record is what *that run* actually paid.
2. **Verified-only price table** — `scripts/model_comparison/pricing.py`. Every entry carries the date it was verified and a note on introductory-rate expiry (the Gemini 3.x flash rates double on 2027-01-01). Token counts are the durable fact; dollars are an estimate; the billing console stays authoritative.
3. **Server-side reconciliation** — `scripts/analysis/vertex_usage.py` pulls the project''s actual billed token counts per model from Cloud Monitoring (daily granularity, ~6 weeks retention). This recovers runs that predate the instrumentation, which is how the table below exists at all.
4. **Docs** — a "Cost accounting" section in `docs/model_comparison.md` with the measured numbers and caveats (the daily rows are query-anchored 24 h windows, not calendar days).

## Measured: the complete Gemini history of the benchmark (Cloud Monitoring, 2026-08-15)

| model | input tokens | output tokens | est. cost |
|---|---:|---:|---:|
| gemini-2.5-flash | 4,811,091 | 1,479,263 | $5.14 |
| gemini-3.6-flash | 17,449,987 | 6,119,688 | $36.04 |
| gemini-3.1-pro-preview | 17,405,320 | 593,380 | $41.93 |
| gemini-3.7-flash (leg in flight at time of measurement) | 14,589,790 | 3,584,187 | $24.38 |
| **total** | | | **~$107** |

Two findings worth keeping: tokenization is deterministic — a 125-pano city leg is exactly **957,000 input tokens** (125 panos x 6 views x 1,276 tokens/view) — and **thinking tokens dominate output cost** and vary enormously by model (3.6-flash emitted 6.1M output tokens on the same input where 3.1-pro emitted 0.59M), so flash legs are not as cheap relative to pro as the rate card suggests.

## Safety of the change

The detection-cache signature is untouched — `test_gemini_cache_key_is_frozen` still pins the hash, so none of the ~paid cached detections are invalidated. Suite: **555 passed**.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,12 @@ analysis_out/op/*
 # Which silent misses another model witnessed (#46). Small, and it carries the
 # unwitnessed list that is the gallery's actual work queue.
 !analysis_out/silent_witness.json
+# What every paid-model run spent, appended at the time it ran (compare.py
+# --usage-log). A few hundred bytes per leg, and it is the only durable record of
+# API spend -- the billing console's retention is ~6 weeks, so an uncommitted copy
+# becomes unrecoverable rather than merely inconvenient. Append-only; never rewrite
+# a past line, since each one is a measurement.
+!analysis_out/usage_log.jsonl
 # Crop-window sizing eval summaries (#114): per-rule containment/context stats plus
 # the sha256 of the per-box CSV each summarizes (the CSVs themselves regenerate).
 # The glob also covers per-bundle runs against box_gallery extent gold (#116),
--- a/docs/model_comparison.md
+++ b/docs/model_comparison.md
@@ -1044,6 +1044,49 @@ public GSV/Mapillary, so residency is not a concern here). Vertex model ids diff
 AI-Studio aliases (`gemini-flash-latest` only resolves on `global`); pin them explicitly with
 `gemini:<model-id>` in `--models`.
 
+## Cost accounting for paid models
+
+**Standing rule: every experiment that spends API money records what it spent, at the time
+it runs.** A paper reports cost alongside accuracy, and a token count that wasn't captured
+at run time has to be reconstructed later (or lost). Three layers:
+
+1. **At run time** — `compare.py` accumulates each paid provider's own usage metadata
+   (currently Gemini; local GPU models are free in API terms) and appends one JSONL record
+   per model run — token counts, panos scored, estimated cost — to `--usage-log`
+   (default `<cache-dir>/usage_log.jsonl`; `--usage-log none` disables). Cached panos make
+   no call, so the record is what *that run* actually paid, not what a full run would cost.
+2. **Prices** — `scripts/model_comparison/pricing.py` is a verified-only table: each entry
+   carries the date it was checked. Prices change (the Gemini 3.x flash rates are
+   introductory through 2026-12-31 and double after); the token counts are the durable
+   fact, the dollar figure is an estimate, and the billing console is authoritative.
+3. **Reconciliation** — `scripts/analysis/vertex_usage.py` pulls the project's *actual*
+   billed token counts per model from Cloud Monitoring (daily granularity, ~6 weeks
+   retention), which also recovers runs that predate this instrumentation. Caveat: its
+   daily rows are 24 h windows ending at the query's time-of-day (UTC), not calendar days —
+   attribute rows to legs from the run record, not by eye.
+
+**Measured: the complete Gemini history of this benchmark** (Cloud Monitoring,
+2026-08-15; every Gemini leg ever run on the project — richmond/bend 07-23/24 onward —
+falls inside the retention window):
+
+| model | input tokens | output tokens | est. cost |
+|---|---:|---:|---:|
+| gemini-2.5-flash | 4,811,091 | 1,479,263 | $5.14 |
+| gemini-3.6-flash | 17,449,987 | 6,119,688 | $36.04 |
+| gemini-3.1-pro-preview | 17,405,320 | 593,380 | $41.93 |
+| gemini-3.7-flash (leg in flight, through 08-15 06:15 UTC) | 14,471,066 | 3,550,843 | $24.17 |
+| **total** | | | **≈ $107** |
+
+Anatomy of a leg: tokenization is deterministic — a 125-pano city leg is exactly
+**957,000 input tokens** (125 panos × 6 views × 1,276 tokens/view at the 1024×1024 view
+size), so input cost is fixed per bundle and only output varies. That puts a 125-pano
+city leg at roughly **$1.60–$2.20** depending on model, and the 1,000-pano manual_gold
+leg at ~8× that. Two output-side facts worth knowing when budgeting: **thinking tokens
+bill as output** and dominate it (the visible JSON is ~50 tokens/call; observed output is
+~250–350), and models differ enormously in how much they think — on identical input,
+gemini-3.6-flash emitted 6.1M output tokens where gemini-3.1-pro emitted 0.59M, which is
+why flash legs are not as cheap relative to pro as the rate card suggests.
+
 ## Running it
 
 ```bash
--- a/docs/model_comparison.md
+++ b/docs/model_comparison.md
@@ -1052,18 +1052,26 @@ at run time has to be reconstructed later (or lost). Three layers:
 
 1. **At run time** — `compare.py` accumulates each paid provider's own usage metadata
    (currently Gemini; local GPU models are free in API terms) and appends one JSONL record
-   per model run — token counts, panos scored, estimated cost — to `--usage-log`
-   (default `<cache-dir>/usage_log.jsonl`; `--usage-log none` disables). Cached panos make
-   no call, so the record is what *that run* actually paid, not what a full run would cost.
+   per model run — token counts, panos actually scored, detector signature, estimated cost —
+   to `--usage-log` (default **`analysis_out/usage_log.jsonl`**, which is committed;
+   `--usage-log none` disables). Cached panos make no call, so the record is what *that run*
+   actually paid, not what a full run would cost. The append happens in a `finally`, so a leg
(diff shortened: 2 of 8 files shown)


jonfroehlich:
## Deep review

Ran the suite on the branch: **555 passed**, confirming the PR body. Every dollar figure in the doc table checks out against `pricing.py` to the cent ($5.14 / $36.04 / $41.93 / $24.17, total $107.28), the 957,000 = 125 × 6 × 1,276 identity holds, and the thinking-token arithmetic is right — the SDK's `candidates_token_count` excludes `thoughts_token_count`, so adding them is correct for billing.

Findings below, ranked.

### 1. The cost record is discarded on exactly the runs that matter — `compare.py:587`

`report_usage` sits *after* the `except Exception: … continue`. If `score_model` raises after the API has been paid — `cache.put` on a full disk, a bad record, a preempted Slurm job — the branch prints "not runnable" and `continue`s straight past it. `KeyboardInterrupt` isn't an `Exception` at all, so Ctrl-C on a long paid leg unwinds `main()` and takes `detector.usage` with it.

The except-branch comment already anticipates this case — it exists for "a multi-model cluster run that has already paid for the models before it" — and that is the run whose spend is dropped. `try/…/finally: report_usage(...)` fixes it; `report_usage` already no-ops at `calls == 0`, so it is safe on every path.

### 2. Bookkeeping can kill the results it is bookkeeping — `compare.py:587`, `310–316`

That same call is unguarded and outside the per-model try. The table, sweep and PR curves only print after the loop, so a `PermissionError` from `open(usage_log_path, "a")` — read-only path, quota, a stale directory of that name — destroys a whole multi-model run's output after the money is spent.

### 3. `usage['thoughts_tokens']` is indexed unconditionally — `compare.py:301`

Confirmed by running it: a provider that sets `usage` without a thinking field raises `KeyError: 'thoughts_tokens'`, and per finding 2 that kills the run. Use `.get(..., 0)`. The base class declares `self.usage = None` but no key contract; an `_accumulate_usage(input, output, thoughts=0)` helper on `_VLMDetector` would settle it.

### 4. The default log lands in the one directory we already know is lost — `compare.py:557`

`--usage-log` defaults to `<cache-dir>/usage_log.jsonl` → `.model_cache/`, which is `.gitignore:12`. The comment at `.gitignore:49` says it outright: *".model_cache itself is gitignored, so without this the table is unreproducible."* That is the same reasoning that produced the `op_cache/` and `fp_taxonomy.json` re-includes, and #120, which exists to rescue detections from this exact directory. The PR calls token counts "the durable fact" and then defaults them where git cannot see them. `analysis_out/usage_log.jsonl` plus a re-include line matches the established pattern.

Related: the default is computed from `--cache-dir` regardless of `--no-cache`, so `--no-cache` still `os.makedirs`es `.model_cache/`.

### 5. The price table cites the wrong rate card — `pricing.py:16`

The source comment points at `ai.google.dev/gemini-api/docs/pricing`, but the runs bill through Vertex: `GeminiDetector` takes the `vertexai=True` + ADC path (`detectors.py:585–590`), and `vertex_usage.py` reads Cloud Monitoring for a GCP project. The two lists agree today, but the file's own rule is "the page it came from", and the introductory-expiry footnote is an AI-Studio fact that need not track Vertex. Cite the Vertex page, or record both with which path each governs.

### 6. `vertex_usage.py` hardcodes the project and ignores the documented env var — line 69

`--project default="rampnet"` never consults `GOOGLE_CLOUD_PROJECT`, which `detectors.py:561` reads and which `docs/model_comparison.md` tells users to set in `.env`. Someone following those instructions gets a 403 against a project that is not theirs, with no hint the flag exists.

### 7. `nextPageToken` is dropped — `vertex_usage.py:64`

`r.json().get("timeSeries", [])` takes page one and stops. Monitoring v3 paginates, and `--days 42` across several models × two `type` labels can exceed a page. The failure is a silent *undercount* in the tool whose entire job is server-side ground truth. Same class: only `type` values `input`/`output` are ever read (lines 92, 109) — a third bucket accumulates into the dict and is never printed or priced.

### 8. `panos_scored` records the wrong denominator — `compare.py:587`

It logs `len(gts)`, every pano in the bundle. On a resumed run where 100 of 125 are cached, the record says `panos_scored: 125` next to `calls: 150` — anyone dividing cost by panos is 5× low, and partial caching is the normal case here. `len(run.scored)`, plus a count of panos that actually issued calls.

### 9. The "anatomy of a leg" figures do not reconcile with the table above them — `docs/model_comparison.md`

Using the doc's own constant (1,276 input tokens/view), the table implies calls = input/1276, so output per call is **392** (2.5-flash), **447** (3.6-flash), **43** (3.1-pro), **313** (3.7-flash). Only one of four lands in the stated "~250–350", and the next sentence — "models differ enormously" — argues against a narrow band. Per-leg cost from the same table is **$1.02 / $1.98 / $2.31 / $1.60**, so "$1.60–$2.20 depending on model" excludes two of four. These are the numbers a paper would quote for budgeting; they should fall out of the table printed directly above them.

### Smaller

- The usage record carries no detector signature, so a `--tiling perspective` run and a `--tiling none` run of the same model on the same bundle log identically at ~6× different input cost. 9e87290 just closed this for the detections artifact.
- `## Files` in `docs/model_comparison.md` is not updated with `pricing.py` or `vertex_usage.py`.
- `vertex_usage.py` imports `google-auth` and `requests` directly, and its error message says to install `requirements-vlm.txt`, which declares neither (both arrive transitively via `google-genai`).
- The tests add a second mid-file import block duplicating the existing `from compare import (…)`, and reach `PRICING` via `__import__("pricing")` one line after importing from it.
- Hardening, not a bug: `_record_usage` could assert `total_token_count ≈ prompt + candidates + thoughts`. The arithmetic is correct on the current SDK; the assert is what catches it if that ever inverts, since the unit test checks a hand-built fake.

jonfroehlich:
Pushed bff1a1e addressing the review above. Suite **561 passed** (555 before; +6 tests covering the `finally` path, an unwritable log, a provider with no thinking field, the signature in the record, and the `total_token_count` cross-check).

Fixed: the `finally` call site and `panos_scored`; the guarded log write and `.get('thoughts_tokens', 0)`; the default log moved to committed `analysis_out/usage_log.jsonl` (which also stops `--no-cache` materialising `.model_cache/`); the detector signature in each record; the usage-shape contract on `_VLMDetector`; the `total_token_count` cross-check; `--project` from `$GOOGLE_CLOUD_PROJECT`; `nextPageToken` and unpriced-token-type reporting in `vertex_usage.py`; the per-leg doc figures now derived from the table above them; `## Files`; `requirements-vlm.txt`; test import hygiene.

**One left for you, deliberately.** I did not touch any price in `pricing.py` — I have no way to check a rate card, and quietly editing a number whose whole contract is "verified, with the date it was checked" would be worse than leaving it. The file now carries the caveat that the values were checked against `ai.google.dev` while every measured dollar was billed through Vertex, and says to re-check against the Vertex page and stamp a new `as_of` before a paper quotes them. Worth ten minutes before the numbers travel.

Two notes on the doc table: the per-leg figures are now `$1.02 / $1.60 / $1.98 / $2.31` (2.5-flash / 3.7-flash / 3.6-flash / 3.1-pro), each derived from that model's own row by dividing input tokens by 1,276. And `analysis_out/usage_log.jsonl` will be empty until the next paid run — the $107 history predates the instrumentation and exists only via `vertex_usage.py`, which is now stated in the doc as a replication limit rather than left implicit.

jonfroehlich:
Pushed 7b66016, which **closes #121**. Landed it here rather than as a stacked PR: the usage record is the artifact build provenance belongs in, and a separate PR would only add a merge ordering to get wrong.

**The problem.** A pinned model id is an alias. `gemini:gemini-3.7-flash` resolves to whatever the provider currently serves under that name, and the provider is free to move it. `cache_key` is `hash(label, signature, city, pano)`, so a run today and a run after a rotation produce the **same key and the same published signature** — two models' detections can land in one file with nothing to tell them apart, and the file carries no date or build to sort it out afterwards. #20 already records an orphaned clovis run that had to be split 50/50 by hand.

**Why the build does not go in `signature()`.** That changes every cache key, misses every already-paid detection, re-bills the run, and trips `test_gemini_cache_key_is_frozen` — "if this fails, the change was not free". So it is recorded *beside* the signature, never inside it, and there is now a test asserting that separation directly, so the reason is guarded rather than only the hash.

**What landed:**

- `GeminiDetector` records `resp.model_version` per call as `{build -> calls}`, and **warns once if it changes mid-run** — the case nobody would otherwise notice, since alias, signature and cache key all stay put.
- Recorded *before* the `usage_metadata` guard: a response can carry the build without carrying usage, and the build is the half that cannot be back-filled.
- `report_usage` prints `served by: <build> (N calls)` and writes `model_versions` into the JSONL record — `None`, not `{}`, when a provider reports nothing, since an empty dict would read as "asked and got nothing", which is the state every existing file is genuinely in.

**Stated in the doc rather than left implicit:** this cannot be reconstructed after the fact — `.model_cache` stores detection points only — so every file currently in `benchmark/model_detections/` carries no build id and never will. An argument for capturing it now, not for skipping it.

Suite: **565 passed** (561 before, +4). `test_gemini_cache_key_is_frozen` still passes, so nothing cached is invalidated.

Separately, I filed #122 on the challenger roster — adding a model currently costs far more than it should, and the expensive part is that it moves the #46 human-pass item list. Details there.

PR #120, opened 2026-08-15
Publish the gemini-3.7-flash detections for the nine city splits (#20)

jonfroehlich:
The gemini-3.7-flash leg''s paid outputs lived only in one desktop''s git-ignored `.model_cache`; anyone re-deriving the interim F1 numbers posted on #20 would have had to re-pay the API calls. This publishes them the established way (`scripts/analysis/export_model_cache.py`, #46 replication pattern): one JSON per split under `benchmark/model_detections/`, detector signature embedded, 0 uncached panos on all 9 splits.

**Verification:** `--verify` re-scored every split from the published files and from the cache — identical TP/FP on all 9 (model, split) pairs.

**What this makes reproducible:** the detections behind the #20 interim table are now committed and verified identical to what produced it. Per the existing contract (docs/replication.md), the analysis scripts (`fp_taxonomy.py`, `silent_witness.py`, etc.) read these published files directly — no cache, no GPU, no credentials. `compare.py` itself still scores from `.model_cache`, so regenerating the F1 table verbatim on a clean clone means scoring the published points (which `--verify` proves is the same computation), not re-running the model — the spend is never repeated either way.

`manual_gold` is deliberately absent — that leg was still running at export time (810/1000); it gets exported the same way once complete, alongside the full write-up in `docs/model_comparison.md`.


jonfroehlich:
## Deep review

This is data, not code, so I validated the files directly out of the git objects rather than reading the diff.

**Clean where it can be checked.** Schema identical to the existing legs; all nine files byte-identical to canonical sorted-compact JSON, matching every other published file; pano sets exactly equal `records.jsonl` *and* the gemini-3.6-flash leg on all nine splits; `n_panos` consistent and `n_uncached: 0` everywhere; every coordinate finite and in [0,1]; the signature byte-identical to the other two Gemini legs apart from `model_id`; null-confidence convention matches. I also checked for cache contamination — only 7 of 1,808 points coincide exactly with a 3.6-flash point (0.39%, consistent with Gemini's 0–1000 box quantization), so the leg is genuinely independent. Suite passes (551).

### 1. Nothing in the repo reads these files

`gemini:gemini-3.7-flash` is not in `CHALLENGERS` (`scripts/analysis/fp_taxonomy.py:69`), and grep finds no mention of the model anywhere on main outside the new filenames. So:

- `export_model_cache.py --verify` — the exact command `docs/replication.md:39` tells a replicator to run — defaults `--models` to `CHALLENGERS` and **skips this leg entirely**. Run against this branch it prints "verified 68 (model, split) pair(s) / published detections score IDENTICALLY to the cache", exit 0, without opening a single new file.
- `silent_witness.py` iterates `CHALLENGERS` and has no `--models` flag at all, so it cannot be pointed at the new leg without editing the file.
- `fp_taxonomy.py` needs an explicit `--models gemini:gemini-3.7-flash`, which is recorded nowhere in the repo.

So "the analysis scripts read these published files directly" is not yet true for this model. Adding the spec to `CHALLENGERS` is the obvious fix, but it would change the committed `fp_taxonomy.json` / `silent_witness.json` derived artifacts, so it needs a re-run rather than a one-line edit — worth deciding deliberately.

### 2. The coverage matrix inverts on merge — `docs/model_comparison.md:11–27, 1007`

The matrix says **"✅ all 8"** on every split, defines the roster as "RampNet, 2 Geminis, 2 Qwens, Molmo, OWLv2, Grounding DINO", and repeats "**Every split now carries the full 8-model roster**" at line 1007 — under the opening guarantee that *"an omission below is a run that hasn't happened, not a result being withheld."* After merge, a third Gemini's detections sit in the repo for nine splits while the document tells the reader it was never run. That is the exact ambiguity the matrix exists to remove, pointed the wrong way.

### 3. `docs/replication.md` is the ledger, and it is now off

It states **"61 human-readable files, 18.0 MB"**, **"61/61 pairs verified identical"**, and lists the set as "(Gemini ×2, Qwen ×2, Molmo, OWLv2, Grounding DINO)". 61 was right at 2811f73; São Paulo pushed it to 68 without an update, and this PR makes it **77 files / 20,221,935 bytes / three Gemini legs**.

### 4. `--verify` is a lossy hash, and it can pass vacuously — `export_model_cache.py:154–163`

Two separate problems in the gate this PR rests on:

- It compares `tp * 1000 + fp` summed per split. OWLv2 and Grounding DINO run **7,300–9,700 detections per split**, an order of magnitude past the modulus, so a compensating difference (+1 TP, −1000 FP) verifies as identical. This PR's data cannot trigger it (136–273 per split), but the check does not hold for the dense legs. Comparing `(tp, fp)` tuples costs nothing.
- `checked += 1` fires regardless of whether any pano was compared, and `if cached is None: continue` skips silently — so with no cache at all it prints "verified 68 pairs / score IDENTICALLY to the cache", exit 0, having compared nothing. Worth failing, or at minimum reporting zero, when no pair had anything to compare.

(`docs/replication.md`'s "Verified by running against a nonexistent `--cache-dir`" is attached to a claim about downstream scripts, which would be a legitimate test — but it is ambiguous which command was actually run, and worth restating.)

### 5. A partial export writes, and then verifies clean — `export_model_cache.py:119, 162`

`export()` writes whenever `dets` is non-empty and ignores `missing` entirely; `verify()` skips any pano the cache lacks, so a pano uncached at export time is absent from *both* sides and never compared. This PR says manual_gold "gets exported the same way once complete" — exported at 810/1000 it would write `n_uncached: 190`, verify identical, and `silent_witness.model_predictions` would narrow to the panos present with no count reported, silently scoring 810 panos as if that were the split. `fp_taxonomy` at least tallies uncached; `silent_witness` does not.

### 6. The gemini-3.7-flash empty-response rate is 2.1× the 3.6-flash rate, and the artifact cannot explain it

`boxes_from_gemini_response` returns `[]` when `resp.parsed` is falsy **and** `resp.text` is empty (`detectors.py:82`), so a truncated, blocked or candidate-less response is cached and published as an authoritative "no ramps here". Malformed JSON raises and is recorded as a failure; this case is silent. Measured across the nine new files:

| | 3.7-flash | 3.6-flash | 3.1-pro |
|---|---:|---:|---:|
| panos with zero detections (of 1,109) | **345** | 163 | 255 |

Worst splits: são_paulo 59/125 (3.6-flash: 11), clovis 47 (28), budapest 46 (16), paterson 41 (18). There are 35 panos where 3.7 emitted nothing while **both** other Geminis emitted ≥2 points (23 of those also have ≥2 GT ramps).

Two things temper the "dropped response" reading, though: `tile: true` means **six independent view calls per pano** merged by `dedup_points`, so an empty pano needs all six to come back empty — not the shape random truncation or safety blocks produce. And 3.1-pro sits at 255, between the two flashes, so "3.7 is conservative" and "3.6 is unusually eager" fit about equally. Worth resolving before the write-up, since it bears on the high-precision reading in #20. #119's numbers are the evidence to hand: 3.7-flash averages ~313 output tokens/call against 3.6-flash's ~447, which argues against truncation on the mean, though not in the tail.

### Smaller

- `tests/test_export_model_cache.py:103` iterates `files[:5]`, which lands on the Molmo entries — none of the nine new files is ever validated, and nothing anywhere asserts `n_panos == len(detections)`, `n_uncached == 0`, or coordinates in range. All hold here; nothing keeps them holding.
- `signature.model_id` pins the floating alias `gemini-3.7-flash` with no run date, SDK version, or resolved build. Vertex resolves it to whatever build is current, so a re-run after Google rotates it writes into the same cache-key space — the variant that produced #20's orphaned clovis run.
- The exact producing command (`--models gemini:gemini-3.7-flash`) is committed nowhere, and the deliberate `manual_gold` omission lives only in this PR body — a clone shows 10 files for five models and 9 for this one, with no in-repo explanation.

jonfroehlich:
Pushed 09c4454 addressing the review above. Suite **553 passed**, and I re-ran `--verify` against the real cache with the stricter comparison: **68/68 on the default roster and 9/9 for gemini-3.7-flash, all identical.** So the data claim in this PR holds up under a check that can actually fail — see below.

Fixed: `docs/replication.md`'s counts (77 files / 20.2 MB) and the exact `--models gemini:gemini-3.7-flash` commands, with a plain statement that the default roster does not cover this leg; the coverage-matrix inversion in `docs/model_comparison.md`; the empty-response question recorded next to the leg; `--verify` now compares per-pano `(TP, FP)` tuples instead of `tp*1000 + fp`, and exits 1 when it had nothing to compare (against a nonexistent `--cache-dir` it now prints `NOTHING VERIFIED` and exits 1, where it used to print a clean pass); `export()` refuses an incomplete split without `--allow-partial`; the test now walks all 77 files and asserts filename/content agreement, `n_panos`, `n_uncached == 0`, point shape and range, and pano-set equality with each bundle.

**Three left open, on purpose:**

1. **Adding `gemini:gemini-3.7-flash` to `CHALLENGERS`** would change the committed `fp_taxonomy.json` and `silent_witness.json`, so it is a re-run and a set of new numbers, not a one-line edit. `silent_witness.py` also has no `--models` flag at all, so it needs that first. Documented as open rather than done.

2. **The empty-response question.** 345 of 1,109 panos with zero detections, against 163 for 3.6-flash and 255 for 3.1-pro, and `boxes_from_gemini_response` cannot distinguish "no ramps" from "no parseable content". I recorded it beside the leg with the two facts that argue against a bug (six view calls per pano must *all* come back empty; 3.1-pro sits between the flashes), but resolving it needs a run that captures finish reasons — worth doing before the precision number is quoted anywhere.

3. **The floating `model_id` alias.** I did not add a run date or resolved build to the signature: `signature()` feeds `cache_key`, so changing it invalidates every cached detection and trips `test_gemini_cache_key_is_frozen` — the paid caches are exactly what we are trying not to throw away. A sibling field outside the signature (`exported_at`, SDK version) would work, but it means re-exporting and re-verifying all 77 files, which is a bigger change than a review fix should make unasked. Happy to do it if you want it.

jonfroehlich:
Pushed b9e4eba. The `manual_gold` leg finished **2026-08-15 08:04 UTC** — a few hours after the first nine splits were exported — so the reason this PR gave for its absence no longer held. Exported it the same way (**1,000 panos, 0 uncached**) and re-verified: **10/10 pairs score identically to the cache.**

That makes gemini-3.7-flash the only Gemini with a `manual_gold` file. The other two are absent for a different reason — their `manual_gold` detections aren't in the cache under current keys (#20) — which `docs/replication.md` now says, rather than leaving an unexplained asymmetry in the directory.

## The silence question is settled: conservative model, not lost responses

With the gold labels available, `scripts/analysis/empty_response_check.py` answers it from committed data alone (published detections + `manual_labels`; no cache, no GPU, no credentials). Three tests, and they agree:

| test | result | reading |
| :--- | :--- | :--- |
| Where it fires, does it find as much? | On the **755** panos where both fire, 3.7 averages **2.38** detections vs 3.6's **3.03** — ratio **0.79** (0.80 vs 3.1-pro, n=727) | A dropped response removes a whole view and leaves survivors untouched. Finding ~21% fewer ramps *where it does fire* is uniform strictness, not concentrated loss. |
| Does silence track the ground truth? | Silent panos carry **0.80** gold ramps on average vs **4.96** where it fires; **186 of 251** silent panos (74.1%) are genuinely empty in the gold set | It goes quiet almost exactly where there's nothing to find. This is also what rules out a *correlated* whole-pano refusal, which the six-view arithmetic can't. |
| Six-view arithmetic | Independent per-call loss would need **p = 82.3%** | Absurd for a leg that exited 0 at token counts in line with its siblings — and the same arithmetic indicts 3.6-flash at 72.6%, which nobody suspects. Rules out independent loss and nothing more. |

So the "conservative, high-precision" reading on #20 is **supported**, and its precision can be quoted.

**The caution isn't free, and that's recorded too:** 201 of `manual_gold`'s 3,919 gold ramps (**5.1%**) sit on panos the model never spoke about, so about five points of recall are unreachable at any threshold. A ceiling on the leg, not a bug in it.

**Caveat kept beside the numbers:** none of this inspects a response that actually failed — the harness records no finish reasons, so this infers from the shape of what came back. A leg run with `finish_reason` captured would settle it directly, and would be worth doing if a future model shows a *higher* silent rate on ramp-dense panos specifically.

`tests/test_empty_response_check.py` guards both measured directions, so a re-export that inverts either fails the suite instead of quietly falsifying the doc. Suite: **555 passed**.

Still open from the earlier review: adding the spec to `CHALLENGERS` (a re-run, since it changes committed derived artifacts) and the floating `model_id` alias (dating it would invalidate every paid cached detection).

jonfroehlich:
Merged main in (12c54f7) — this had gone `CONFLICTING` after the #87 and #106 merges landed.

One conflict, in `scripts/analysis/export_model_cache.py`: main's 9e87290 and this branch both changed `export()` and `verify()`. The two changes are orthogonal and **both are kept**:

- main's `overrides`, which patches the `compare.py`-defaults namespace so an export runs at the producing run's signature knobs (`--tiling`, `--yolo-imgsz`);
- this branch's `allow_partial`, which refuses to publish a split whose cache is incomplete, plus the `vacuous`/`unpublished` bookkeeping that stops `--verify` reporting a pass when it compared nothing.

`verify()` takes `overrides` now as well, which it needs for exactly the reason `export()` does: verifying at the wrong signature finds no cache and compares nothing — which is the state the `vacuous` check exists to refuse, so the two features are complementary rather than competing.

555 passed on the merge result, and `--help` confirms both flag sets are present.

Issue #122, opened 2026-08-15
Make adding a challenger cheap: one registry, and stop it moving the #46 human-pass item list


jonfroehlich:
Adding a challenger to the benchmark should be cheap. Right now it is not, and the cost falls in a place that is easy to miss: a completed human pass.

The roster is expected to keep growing — gemini-3.7-flash is published and off-roster today (#120), and a Claude leg is next. So this is worth fixing before the next addition rather than after.

## What adding a model to `CHALLENGERS` currently disturbs

`CHALLENGERS` is a 7-entry tuple at `scripts/analysis/fp_taxonomy.py:69`, read by `export_model_cache.py` (`--models` default), `fp_taxonomy.py` (`--models` default), and `silent_witness.py` (iterated directly — **no `--models` flag at all**).

1. **It moves the #46 human-pass item list.** `silent_witness` computes the RampNet "silent" misses that *no other model witnessed*. A ninth witness can only shrink that set — more models, more chances something saw it. The set is currently **59 unwitnessed → 50 crops** after the 30-source-pixel floor, and those 50 crops are the item list for the #46 tagging pass, which is **done, with committed per-rater verdicts**. Adding a challenger silently invalidates the correspondence between the verdicts and the list they were made against.
2. **`SPARSE` / `DENSE` are a second hardcoded roster** (`silent_witness.py:60`) that has to be kept in sync by hand, and the output prints "union of the 5 sparse models" / "union of the 2 dense detectors" — the counts are baked into the strings.
3. **Committed derived artifacts change**: `analysis_out/fp_taxonomy.json`, `analysis_out/silent_witness.json`.
4. **Doc prose quotes roster counts**: `docs/model_comparison.md`'s "all 8" coverage matrix, and `docs/curb_ramp_data_sourcing.md`'s 0.009-vs-0.022 gap, which is built on the 59.

None of that is wrong as a record of what was run. It is just far too much ceremony for "we have a new model to try", and the failure mode is silent.

## Proposal

- **One registry, one source of truth.** A single table declaring each challenger's spec, display label, density class, and the date it entered the benchmark. `CHALLENGERS`, `SPARSE` and `DENSE` all derive from it instead of being three tuples kept in sync by hand.
- **Freeze the witness pool the #46 pass was made under**, as a named constant with a comment saying so. The tagging verdicts stay valid because the pool that generated their item list stops moving. New challengers change the comparison tables; they do not retroactively change a finished human pass.
[...]

PR #123, opened 2026-08-15
Wire Claude into the model-comparison harness (#122): Opus 5 at low effort takes the top challenger slot on annapolis

jonfroehlich:
Adds Claude as a challenger in the model-comparison harness (`--models claude:<model-id>`) and runs the first real leg: **annapolis, both models × both effort levels, 125 panos each**.

**`claude-opus-5` at `low` effort is the strongest general model measured on this split, F1 0.588** — the first to displace `gemini-3.1-pro-preview` (0.567) from that slot. RampNet still leads it by 0.251 (0.839 vs 0.588).

Groundwork for #122. (Originally stacked on #119, which has since merged; `main` is merged in, so this bases on `main` now.)

## Results (annapolis, 125 panos, 294 GT ramps)

Shared `DETECTION_PROMPT` — deliberately **not** Claude-tuned, so these stay comparable to the other legs — `auto` tool choice, identical rig.

| model | effort | P | R | F1 | tp/fp/fn | thinking tok | cost |
| :--- | :--- | ---: | ---: | ---: | :--- | ---: | ---: |
| claude-sonnet-5 | low | 0.589 | 0.381 | 0.463 | 112/78/182 | 57 | $3.60 |
| claude-sonnet-5 | high | 0.506 | 0.415 | 0.456 | 122/119/172 | 17,820 | $3.82 |
| **claude-opus-5** | **low** | 0.572 | 0.605 | **0.588** | 178/133/116 | 523 | $8.94 |
| claude-opus-5 | high | 0.430 | 0.656 | 0.520 | 193/256/101 | 127,227 | $12.46 |

**Every figure here is re-derivable from committed files.** The per-panorama detections are published under `benchmark/model_detections/claude-*-effort-*__annapolis.json`, and `tests/test_claude_annapolis_leg.py` recomputes the whole table from them — no `.model_cache`, no credentials, no GPU. Editing a number in the doc without re-running anything fails CI.

**Effort is an operating-point dial, never a quality lever.** Both models move the same direction — thinking makes them fire more, so recall rises and precision falls — and F1 never improves. Sonnet nets slightly negative (0.463 → 0.456); Opus nets clearly negative (0.588 → 0.520), because precision falls harder than recall rises. Two models, same direction, so this is a property of the task rather than of one model — the same shape as this benchmark's Qwen 8B→32B finding, where scaling flipped the failure mode instead of fixing it. Note that the expensive setting is the worse one: 127k thinking tokens to lose 0.068 F1.

Opus's gain over Sonnet at equal effort is **+0.224 recall at essentially unchanged precision** (0.589 → 0.572), so that one is a capability difference, not a threshold shift.

## Four caveats that travel with the numbers

- **`claude-opus-5` at `high` has the highest recall of any challenger on this split — 0.656 against RampNet's 0.738**, closer than any general model has come here. That matters against this project's recall-first framing (a false negative is a permanent loss, a false positive is cheap). F1 ranks it below `low`; a recall-weighted objective would not. Recorded rather than decided.
- **The prompt is fixed and Gemini-derived.** Claude is documented to underperform on prompts carried over from another model, so these bound *Claude-on-our-prompt*, not Claude. Holding it fixed is the right call for comparability; a prompt variant is a separate, separately-labelled experiment.
- **The two paid legs are not fed identical pixels.** Claude gets JPEG q90; the Gemini leg gets lossless PNG, because `google-genai`'s `pil_to_blob` only picks JPEG for an image loaded from a jpeg *file* and a reprojected view is an in-memory `Image.fromarray`. `--claude-image-format png` closes it at zero token cost, but it is a cache-key change, so the published numbers are the JPEG ones. Direction of any bias unmeasured.
- **Decoding is not pinned.** `GeminiDetector` sets `temperature=0.0`; the Claude legs took the provider default. `--claude-temperature 0.0` matches them, at the same re-run cost.

## Two Vertex constraints worth knowing before wiring this up

Both are written into `docs/model_comparison.md` and guarded by tests.

**1. Structured outputs are blocked by org policy.** The GCP org sets `constraints/vertexai.allowedPartnerModelFeatures`, which allow-lists *features* of partner models rather than the models themselves, and `structured_outputs` is not on it. `output_config.format` returns 400 FAILED_PRECONDITION — and so does a tool marked `strict: True`, because that is structured outputs underneath. A plain, un-`strict` tool yields the same schema-shaped `{"boxes": [...]}` in a `tool_use` block, so that is what the detector uses. `test_claude_box_tool_is_not_strict` guards the regression, because adding `strict` back 400s every call. `output_config.effort` is allowed, so the cost lever survives.

**2. Vertex 404s intermittently after enablement.** Each Claude model is enabled separately in Model Garden, and for hours afterwards a valid request intermittently gets `404 Publisher model ... was not found`. Measured: 12/12 identical calls succeeded in one burst, then 3 of 5 panos 404'd minutes later. The SDK does not retry 404 (a 4xx is normally permanent), so the detector retries it explicitly with backoff — without that, a leg silently loses panos to a transient lie.

## Design notes and bugs found

- **Forcing `tool_choice` silently disabled effort.** Forced tool-calling suppresses thinking entirely, so `--claude-effort` did nothing at all — identical output tokens at forced/low and forced/high was the tell. `auto` is now the default and forced is a `--claude-tool-choice` option. Both settings are in the cache signature, since they change what the model returns.
- **Boxes are pixels, named `x1/y1/x2/y2`**, never a 4-element array. Gemini's `[ymin, xmin, ymax, xmax]` ordering is silently transposable and this repo has shipped that bug before (Molmo, 61c52d0). **Box mapping was verified with `dump_detections.py` for both models before reading anything into the numbers** — predictions sit tight on the ramps, no offset. That gate exists because a wrong convention does not crash, it produces a plausible wrong number.
- **Anthropic's `output_tokens` already includes thinking**, unlike `google-genai`'s `candidates_token_count`. Adding them the way the Gemini path does would double-count the dominant cost term; a test pins that, and the detector warns if the relationship ever inverts.
- `pricing.py` gains verified `claude-sonnet-5` and `claude-opus-5` entries, read from the Google Cloud rate card for the `global` endpoint on 2026-08-15 — the page that actually governs this billing path. Regional endpoints cost 10% more, and the detector now warns when pointed off `global`, because that would make every cost figure ~9% low with nothing else to show for it.

## Review fixes (deep review, [comment](https://github.com/ProjectSidewalk/RampNet/pull/123#issuecomment-5329402404))

Fourteen findings; the substantive ones:

- **A malformed *response* still cost a whole panorama.** `claude_boxes_to_points` was hardened but `boxes_from_claude_response` was not: prose raised `JSONDecodeError`, a bare array raised `AttributeError`, `{"boxes": null}` raised `TypeError`. More likely under `auto`, which is what makes a text-only turn possible at all. It now reuses `_first_json_blob`, the scanner the Qwen path already had.
- **A `max_tokens` truncation scored as "no curb ramps here" — and was cached.** Thinking bills against `max_tokens`, so a cut-off turn has no `tool_use` and no text, every GT ramp on that pano becomes a false negative, and `score_model` writes it to the cache. `_check_stop_reason` now raises instead, and tallies every stop reason; refusals still score as "found nothing" but are counted and reported.
- **The sonnet/low row was scored on 290 GT ramps, not 294.** One panorama was lost to exactly the parse crash above, so that row used 124 panos while the others used 125 — and the original "Sonnet nets *exactly* flat, 0.456 → 0.456" compared two different pano sets. Re-run under the fix (6 calls, $0.03); the table above is the whole split. The finding survives and is slightly strengthened.
- **The detections are now published.** Four files, all verifying identical to the cache. This exposed a general gap: one model id can be several legs (effort is in the signature), and both would write `claude-sonnet-5__annapolis.json`. `--publish-as` names the published file without touching the cache label, and the exporter now refuses to overwrite a file whose recorded signature differs.
- **The spend record has an unrecoverable gap, and it is stated.** The four legs' token counts were never written to `analysis_out/usage_log.jsonl`; the $28.82 comes from console output. A re-run cannot back-fill them, because cached detections make zero API calls. Only the 2026-08-18 re-run ($0.03) is in the log. `report_usage` now warns loudly when a paid leg finishes with no log destination.
- Smaller: `claude` was missing from three provider rosters (now generated from one `PROVIDERS` tuple); `pricing.py` claimed `claude-opus-5` was unavailable on the day two Opus legs ran; `dump_detections` could crash on the malformed item the detector tolerates; cache-read/write tokens are recorded and priced; the published-corpus count in `docs/replication.md` is now a test rather than a promise, after drifting 61 → 68 → 78 → 108 unremarked.

## Spend

**$28.82** for the four legs, plus **$0.03** for the single-panorama re-run.

## What this does not do

- **The model is deliberately not added to `CHALLENGERS`**, so nothing downstream moves and no other split's numbers change.
- **Only annapolis has been run.** The other nine splits are not measured — about $57 at `low` effort.
- **The complementarity analysis is not here** — how many of RampNet's 77 annapolis misses `claude-opus-5` recovers is the obvious next question and is still open. It is now answerable from the committed detections without re-paying.

Suite: 1,067 passed, 1 skipped.


jonfroehlich:
## Deep code review

Reviewed at `e2c4eca`. Verified first, so the findings below are read against a known-good baseline:

- Merge with `main` is clean; suite is **575 passed** on the branch and **1013 passed / 1 skipped** merged with `main`.
- The cost table reproduces from its own token counts (2,229 in × 745 calls × $2/M + output ≈ $3.61), and the ≈$61 ten-split extrapolation checks out once `manual_gold`'s 1,000 panos are counted.
- `output_config`, `usage.output_tokens_details.thinking_tokens`, and `region="global"` all exist in the installed `anthropic` 0.122.0, so the API surface is real rather than assumed.

---

### High

**1. Three parse paths still cost a whole panorama — the bug this PR set out to fix.** `claude_boxes_to_points` was hardened; `boxes_from_claude_response` (`detectors.py:790-806`) was not. Measured on the branch:

```
prose text ("I can see two curb ramps...")   -> JSONDecodeError
JSON array instead of object                 -> AttributeError: 'list' has no 'get'
tool_use with {"boxes": null}                -> TypeError: 'NoneType' is not iterable
tool_use with {"boxes": 3}                   -> TypeError
```

This matters *more* because of 28d65d9: under `forced` a text-only turn was impossible; under the new `auto` default it is the ordinary shape of a refusal. The docstring already promises the fix — "we read a JSON object out of the text if there is one" — and the helper already exists: `_first_json_blob()` at `detectors.py:102`, used by the Qwen path. `boxes_from_gemini_response` (`detectors.py:86`) also already handles list-vs-dict. Four lines, reusing what is there.

**2. A truncation or a thinking-only turn is silently scored as "no ramps here", then cached.** `_call` fixes `max_tokens=4096` (`detectors.py:926`) while `--claude-effort` exposes `xhigh` and `max`; thinking counts against that budget. Nothing inspects `resp.stop_reason`. A truncated response has no `tool_use` and no text → `boxes_from_claude_response` returns `[]` → `score_pano` turns every GT ramp on that pano into an FN → `compare.py:283` does `cache.put(key, [])`, making it permanent. This is the worse of the two failure modes: #1 at least lands in `failures` and is visible.

**3. The three artifacts this repo requires for a paid run are all absent.**

- **`analysis_out/usage_log.jsonl` has never existed in any commit on any branch** (`git log --all -- analysis_out/usage_log.jsonl` is empty). The PR body says "$28.82 … all recorded in `analysis_out/usage_log.jsonl`"; `.gitignore:84` un-ignores it precisely so it can be committed; `docs/model_comparison.md:1137,1164,1506` calls it "committed" and "checkable from a clean clone". The claim is currently unverifiable. (Shared with #119, but this is the PR that spends the money.)
- **No detections exported.** `benchmark/model_detections/` has 108 files, zero for Claude. The precedent is exact: the off-roster `gemini-3.7-flash` leg (#120) exported all ten splits, and `export_model_cache.py`'s own docstring warns that off-roster models are "neither exported nor verified … the default command will silently skip those files forever otherwise."
- **No `docs/replication.md` entry**, where #120's commands sit at lines 61-69.

Net effect: the table's tp/fp/fn cannot be re-derived by anyone else, and the complementarity follow-up this PR names as the next question costs a second $28.82. Same for "Box mapping was verified with `dump_detections.py` for both models" — the gate ran, but no artifact of it is in the repo.

---

### Medium

**4. The Sonnet-`low` row is scored on a smaller ground truth than the other three.** tp+fn = 108+182 = **290** for sonnet/low, but 294 for the other three rows — and `docs/model_comparison.md:148` states annapolis is 294 GT ramps over 125 panos. So one pano was dropped by exactly the malformed-item bug this PR describes, while `docs:1081` says "Full 125-pano annapolis split". The headline "Sonnet nets exactly flat (0.456 → 0.456)" is therefore across two different pano sets. Recall arithmetic confirms it: 108/290 = 0.372, 122/294 = 0.415, 178/294 = 0.605, 193/294 = 0.656. Re-running that one pano is six calls; the alternative is a footnote naming the dropped pano.

**5. The two paid legs are not fed the same pixels.** `ClaudeDetector._raw_detect` JPEG-encodes each view at quality 90 (`detectors.py:905`). `google-genai`'s `pil_to_blob` sends **PNG** for these views — its JPEG branch requires `image.format == 'JPEG'` *and* a filename, and reprojected views come from `Image.fromarray`, so both are absent. Undocumented, and it sits directly under the comment at `detectors.py:819-822` saying `max_edge` is "matched to Gemini's so the two legs see the same pixels there." On a benchmark whose miss story is faint far-field ramps this is not nothing. PNG costs zero extra tokens (image tokens are dimensional).

**6. Decoding is not pinned.** `GeminiDetector` sets `temperature=0.0` (`detectors.py:664`); `ClaudeDetector` sets nothing, so it runs at the server default. `temperature` is available on `messages.create` in 0.122.0. Greedy-vs-sampled between the two paid legs belongs in the caveats at minimum.

**7. The doc section describes the design this PR replaced.** `docs/model_comparison.md:1114` — "forced tool-calling is the way in" — and `:1121` — "which is what `ClaudeDetector` uses" — but `auto` is the default and the results at `:1142` were run on `auto`. Same staleness in the `CLAUDE_BOX_TOOL` header comment ("Delivered as a FORCED TOOL CALL", `detectors.py:709`) and in `test_claude_box_tool_is_not_strict`'s comment. `--claude-tool-choice` is not mentioned in the docs at all.

---

### Low

**8.** `pricing.py:65` says `claude-opus-5` is "NOT enabled on the project as of this date — price recorded, model unavailable until enabled", `as_of` 2026-08-15 — the same date two full Opus legs ran.

**9.** `claude` is missing from three provider rosters: `compare.py:16`, `compare.py:506` (the `--models` help), `detectors.py:1370` (`parse_model_spec` docstring). Only the `ValueError` at `detectors.py:1432` was updated.

**10.** Nothing warns when `location != "global"`. `pricing.py:56` says the numbers "are wrong if the rig is ever pointed off `global`", and `GOOGLE_CLOUD_LOCATION` is shared with the Gemini leg — someone setting a region for Gemini makes every Claude cost silently 10% low.

**11.** `docs/model_comparison.md:925` "Validating the box mapping" is the section that catalogs per-provider box conventions (it has a Qwen subsection); this PR adds a fourth convention and documents it 200 lines away.

**12.** The coverage matrix at `docs/model_comparison.md:19` still reads "all 8" for annapolis with no note that a Claude result exists for that split alone — the ambiguity `CLAUDE.md` says the matrix exists to remove.

**13.** `dump_detections.detections_to_view_shapes` has no `isinstance(it, dict)` guard, so the verification tool can still crash on input the detector now tolerates.

**14.** `Usage` carries `cache_creation_input_tokens` / `cache_read_input_tokens`, billed separately and excluded from `input_tokens`. Zero today, but `pricing.py:52-54` documents the cache multipliers as if they apply, so the first person to add `cache_control` gets a silently low number.

---

### What is solid

Effort, tool choice, and the tool schema all in the cache key (`signature()`, `detectors.py:851`) is the right call and is tested. The named-field box schema with the Molmo transposition bug cited as the reason is good defensive design, as is `test_claude_usage_does_not_double_count_thinking` plus the inversion warning — the Anthropic / `google-genai` semantic difference is a genuine trap and it is pinned. The explicit 404 retry with the measurement behind it (12/12, then 3 of 5) is exactly the kind of thing that otherwise eats panos invisibly. Keeping the model out of `CHALLENGERS` so no downstream number moves is the correct scoping.

---

**Note on 5 and 6:** `cache_key` is a SHA1 over the whole signature dict, so adding `image_format` or `temperature` to `signature()` orphans every already-paid Claude entry and makes the table cost $28.82 to regenerate. Flagging that explicitly rather than assuming — it is a spend decision, not a code decision.

jonfroehlich:
## Review resolution — all 14 findings

Two commits: `c1552d7` (code hardening) and `2bdcbd3` (artifacts + export safeguards + docs). Suite **1,067 passed, 1 skipped** (was 1,013), **+54 tests**.

| # | Finding | Status |
| ---: | :--- | :--- |
| 1 | Text fallback / non-list `boxes` crashed a whole panorama | **Fixed** — reuses `_first_json_blob`; hostile-input corpus asserts the whole `response → boxes → points` path is total |
| 2 | `max_tokens` truncation cached as "no ramps here" | **Fixed** — `_check_stop_reason` raises (nothing cached); every stop reason tallied and reported |
| 3 | Detections / usage log / replication entry all missing | **Fixed, except the token counts** — 4 legs published + verified, `docs/replication.md` entry added, `usage_log.jsonl` created. The four legs' token counts are **unrecoverable** and the gap is stated in both docs |
| 4 | sonnet/low scored on 290 GT, not 294 | **Fixed** — pano re-run under the fix ($0.03); row is now 0.589 / 0.381 / 0.463 on the full split |
| 5 | JPEG q90 vs Gemini's PNG | **Documented + switchable** (`--claude-image-format`), per your call — kept as-run, caveat travels with the numbers |
| 6 | Temperature not pinned | **Documented + switchable** (`--claude-temperature`), same call |
| 7 | Docs described the superseded forced-tool design | **Fixed** — section rewritten around `auto`, and why the text path now carries weight |
| 8 | `pricing.py` said Opus was unavailable on the day it ran | **Fixed** |
| 9 | `claude` missing from three provider rosters | **Fixed** — generated from one `PROVIDERS` tuple; a test checks the prose that cannot be generated |
| 10 | No warning when `location != "global"` | **Fixed** — `location_warning()`, printed at client build |
| 11 | Box convention not filed under "Validating the box mapping" | **Fixed** — new subsection alongside the Qwen one |
| 12 | Coverage matrix silent about the Claude legs | **Fixed** — matrix row + an off-roster note mirroring `gemini-3.7-flash`'s |
| 13 | `dump_detections` could crash on a malformed item | **Fixed** |
| 14 | Cache tokens unrecorded | **Fixed** — recorded and priced from verified rate-card entries |

### Three things the review did not ask for but the fixes exposed

- **One model id can be several legs.** Effort is in the cache signature, so `claude-sonnet-5` at `low` and `high` are different detections that both wanted `claude-sonnet-5__annapolis.json` — the second silently overwriting the first, leaving a file that still looks complete and still passes `--verify` against whichever was written last. `--publish-as` names the published file without touching the cache label, and `export_model_cache.py` now **refuses** to overwrite a file whose recorded signature differs, rather than trusting whoever runs it to remember the flag.
- **The partial-export guard caught finding 4 independently**, refusing the sonnet/low export at 124/125 before I had wired up any of the scoring. That guard was already in the repo and it earned its keep.
- **`docs/replication.md`'s corpus count had drifted 61 → 68 → 78 → 108** unremarked, in the one document whose job is keeping the repo honest about what a stranger can obtain. It is now asserted against the directory, along with "every published leg is named in a doc a reader would find".

### Worth knowing

- **No cache was orphaned.** `test_claude_cache_key_is_frozen` pins all four leg keys, and the encoding/temperature settings enter `signature()` only when they deviate from what the published legs ran — so documenting the gap cost $0 rather than $28.82.
- **The three YOLO pano arms (`y11l_pano`, `y11x_pano_h200`, `y26_pano`, 30 published files) are named in `scripts/model_comparison/yolo_baseline/` but in neither `docs/`.** Pre-existing and out of scope here, so I left it — the new test accepts the `yolo_baseline` README as documentation. Flagging it because it is the same class of gap this PR just closed for Claude.

Issue #11, opened 2026-07-02
Temporal-consistency filter bugs: sentinel dates bypass the check; month comparison ignores year


jonfroehlich:
The temporal-consistency filter (only keep curb ramps installed before the Street View capture date) has two defects. Only relevant if the dataset is ever regenerated — the released dataset is what it is — but both should be fixed before any regeneration:

1. **Silent placeholder dates defeat the filter.** `stage_one/dataset_generation/combine_location_data.py:9-27` — `convert_date` returns `"2000-01-01"` for any missing/unparseable install date, silently. Downstream, every dateless record trivially passes the "installed before pano" check regardless of reality.
2. **Month/year coupling bug.** `stage_one/dataset_generation/generate_dataset_meta.py:41-44` — the check tests `curb_year > pano_year` and, independently, `curb_month >= pano_month` with the month comparison ignoring the year. Example: curb installed 2019-05, pano captured 2020-03 → the curb clearly predates the pano, but `5 >= 3` triggers the reject path. This both wrongly discards valid panos and can wrongly accept invalid ones.

Fix: compare `(year, month)` tuples; treat missing dates explicitly (either exclude the record or flag it in the metadata) instead of substituting a sentinel that always passes.

From the post-release code review (July 2026).

PR #124, opened 2026-08-17
Run A epoch curve (#84): val loss bottoms at epoch 5, replicating the paper run

### LINKED ISSUE #84: Stage 2 trains for 1 epoch at constant LR (~9.4k steps): measure the epoch curve, and with it the auto-label ceiling

jonfroehlich:
Run A of the #84 epoch curve is trained. This carries the launch record, the runbook, and now the
result — the **auto-label half only**. The `manual_gold` half, which is the question #84 exists to
answer, is **not scored**, and the doc says so next to the empty columns rather than in a footnote.

## The result

Job `38566413` completed all 8 epochs on 2026-08-16 19:50 PDT. Validation loss bottoms at **epoch 5,
exactly as the paper run does** — the pre-registered expectation held.

| epoch | auto-label val loss | paper run | delta | vs. Run A min |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 0.00052194 | .000520 | +0.37% | +13.5% |
| 2 | 0.00048067 | .000478 | +0.56% | +4.5% |
| 3 | 0.00046341 | .000463 | +0.09% | +0.8% |
| 4 | 0.00046485 | .000466 | −0.25% | +1.1% |
| **5** | **0.00045976** | **.000458** | **+0.38%** | **min** |
| 6 | 0.00046855 | .000468 | +0.12% | +1.9% |
| 7 | 0.00047676 | .000470 | +1.44% | +3.7% |
| 8 | 0.00048100 | .000473 | +1.69% | +4.6% |

The whole curve replicates, not just the minimum: largest per-epoch delta 1.69%, mean absolute
0.61%, and the same shape turn for turn including the non-monotone bump at epoch 4.

## Three caveats, all recorded beside the numbers

- **These are not eight independent confirmations.** Both runs use seed 42 on the same data, so they
  are near-identical draws. What the agreement legitimately buys is a bound: whatever differs in the
  unrecoverable June-2025 code did not move the curve by more than ~1.7% at any epoch.
- **The basin is shallow.** Epochs 3–6 sit within 1.9% of each other, so epoch 5 replicates as a
  *rank* but val loss does not determine the selection sharply. Epoch 1 — the epoch the released
  checkpoint was taken at — is the unambiguous part at **13.5% above the minimum**, which measures
  the "free ~12%" this issue predicted.
- **A requeue handed us a noise floor.** Epoch 5 was computed twice, on two nodes via two resume
  paths: 0.00045980 vs 0.00045976, **0.009% apart**. That bounds resume-path nondeterminism and
  evaluation together, and it is ~91× below the basin's smallest gap — so the ordering inside the
  basin is real structure, even though it is too tight to select on confidently.

## Reproducible from a clean clone

The six TensorBoard event files the run wrote are committed (4.0 MB) — the only part of a 28 GPU-h
run small enough to live in the repo, and it carries the entire auto-label result.
`scripts/analysis/stage2_epoch_curve.py` reads them with the **standard library only**: it parses
the TFRecord framing and the two protobuf messages directly rather than importing tensorboard, so
nobody needs a 100 MB install to read eight floats. Its output was cross-checked against
`EventAccumulator` on klone — both readers give the same eight values to full float32 precision.

```bash
python scripts/analysis/stage2_epoch_curve.py --out-csv docs/data/stage2_epoch_curve_84.csv
python scripts/analysis/plot_epoch_curve.py      # -> docs/figures/stage2_epoch_curve_84.png
```

CSV line endings are pinned to LF so the committed table is byte-stable across platforms and a
regenerated copy can be *proven* identical rather than assumed to be.

**Stated gap:** the paper run's column is still not re-derivable from this repo. Those values were
transcribed in #104 from the June-2025 run's surviving events, so the column carries only 3
significant figures. Committing those rescued event files the same way would close it.

## What is still open

- **`manual_gold` F1 across the 8 checkpoints is not scored.** The checkpoints are at
  `/gscratch/makelab/jonf/rampnet_run_a_84/checkpoints/` and the commands are in the runbook under
  *Scoring the curve*.
- **The Run B gate is therefore unevaluated** — it reads on the calibration-free `manual_gold`
  curve, not on validation loss. Run B has not been submitted.

## Also worth having written down

`sacct -X` reports only the **final** incarnation of a requeued job. It shows `38541865` as a single
2h47m FAILED run and silently hides the two earlier incarnations that actually produced epochs 1–5.
Read requeue history from the run's own logs.

Timing came in under budget: a steady 3.75 h per clean epoch against the 3.49 h/epoch projection
(+7%), and 8 epochs in ~34 h calendar against the ~47 h estimate — the ×1.67 preemption factor was
pessimistic here.

Note this does **not** close #84; the issue's central question is still open. (#84 was auto-closed
on 2026-08-15 by the merge of #104, three hours before Run A logged epoch 1.)


jonfroehlich:
## Code review

Deep review of this branch. Eleven findings, ordered by severity. Two of them are load-bearing on the doc's headline claim, so I re-derived every number below from the committed events rather than taking the diff at its word.

### What I checked first, and what holds

- `scripts/analysis/stage2_epoch_curve.py` **reproduces `docs/data/stage2_epoch_curve_84.csv` byte-for-byte.** The parser is correct: I hand-decoded all 8 `Loss/val_epoch` points, confirmed the Event/Summary field numbers, confirmed the last-writer-wins dedupe on resumed steps, and traced the `STEPS_PER_EPOCH = 9378` resume arithmetic against `train.py`'s `global_step` bookkeeping. It does not drift.
- `plot_epoch_curve.py` runs clean and its printed numbers match the CSV.
- The `--no-tta` / `--threshold 0.0` / cache-keying claims in the runbook check out against `stage_two/evaluate.py`.

So the curve itself is right, and the epoch-5 result stands. The problems are in what surrounds it.

---

### 1. The stated replication gap is false — the paper run's events are already committed here

`docs/stage2_epoch_curve_84.md` (and the PR body) says the paper run's column is not re-derivable from this repo, that the raw events are not committed, and that committing the rescued event files is what would unblock it.

They are committed, and have been since #104: `docs/data/rampnet1_stage2_run/`, 18 tfevents files plus a `SHA256SUMS`. `scripts/analysis/stage2_train_cost.py` already reads them.

This branch's own script, pointed at them:

```
python scripts/analysis/stage2_epoch_curve.py --events-dir docs/data/rampnet1_stage2_run
```

gives the whole paper curve at full float32 precision, epochs 1–11, minimum at epoch 5 (`0.00045824519474990666`). A wrong "what would unblock it" line in a results doc is worse than no line — it sends the next person to re-upload data we already have. The fix is a one-line default change, not a data publication.

### 2. `PAPER_RUN_VAL_LOSS` is a 3-s.f. hand transcription, and it makes the published delta column wrong — including the headline bound

Because of #1, these can be read at full precision instead of typed in. Recomputed:

| ep | Run A | paper (3 s.f., as committed) | paper (true) | delta as published | true delta |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0.00052194 | 0.000520 | 0.00052044 | +0.37% | **+0.288%** |
| 2 | 0.00048067 | 0.000478 | 0.00047826 | +0.56% | +0.503% |
| 3 | 0.00046341 | 0.000463 | 0.00046331 | +0.09% | **+0.023%** |
| 4 | 0.00046485 | 0.000466 | 0.00046562 | −0.25% | **−0.166%** |
| 5 | 0.00045976 | 0.000458 | 0.00045825 | +0.38% | +0.331% |
| 6 | 0.00046855 | 0.000468 | 0.00046813 | +0.12% | **+0.090%** |
| 7 | 0.00047676 | 0.000470 | 0.00046985 | +1.44% | +1.471% |
| 8 | 0.00048100 | 0.000473 | 0.00047271 | +1.69% | **+1.755%** |

Mean |delta| is **0.579%**, not 0.61%.

The one that matters: the doc's claim that the auto-label swap *"did not move the curve by more than ~1.7% at any epoch"*, and the accompanying "largest per-epoch delta is 1.69%", are **understated by the real data — the true maximum is 1.755%**. The bound as written is violated by the very events we already have. It is a small miss, but it is a stated bound in a results doc, and it should be derived rather than asserted.

Separately, 3 s.f. at this magnitude carries ±0.11% quantization, so the published +0.09% (ep 3) and +0.12% (ep 6) are not distinguishable from zero as printed.

### 3. `stage_two/run_build_env.slurm:49` — the skip-create guard turns a preemption into a permanently poisoned env

`if [ -x "$ENV_PREFIX/bin/python" ]` is the only completeness test, and this is a `--requeue` job on `ckpt-all`. If preemption lands after conda has linked python but before torch is installed, the requeued incarnation sees an executable python, **skips `conda env create`**, runs `pip install -e` and the verification block against a half-built prefix, and dies under `set -e`. Every later submit skips creation the same way, so the prefix can never repair itself and the failure looks like a torch problem rather than a build problem.

Guard on a completion sentinel written *after* verification passes, and remove the prefix on retry.

### 4. Runbook step 1b cannot work as written, and breaks the doc's own klone rule

Step 1 `sbatch`es the env build — 15–50 min per the launch record — and step 1b immediately runs `"$RAMPNET_ENV/bin/python" -c "import timm ..."`. There is no `--wait` and no "wait for the job to finish" note, so anyone following the commands in order gets `No such file or directory`. The same step also pulls a ~350 MB backbone **on a login node**, which this repo's own guidance says kills the SSH control master.

### 5. `scripts/analysis/stage2_epoch_curve.py:65` — second TFRecord parser in a directory that already has one

`scripts/analysis/stage2_train_cost.py` already implements `_varint` / `_fields` / `_records` / `read_scalars` for exactly this format, with a `--verify` sha256 mode, and it is covered by `tests/test_stage2_train_cost.py`. The new module's stated rationale for not importing tensorboard is already satisfied by the existing reader.

Two independent parsers of one binary format in one directory will drift, and only one of them is under test. Import the existing one, or lift it to a shared helper.

### 6. The new events directory has no `SHA256SUMS` and no `.gitattributes` rule

The doc says integrity is covered because "the script prints a sha256 for each event file it reads so a regenerated copy can be proven identical." A printed digest with no committed reference digest proves nothing — there is nothing to compare against. `docs/data/rampnet1_stage2_run/` ships a `SHA256SUMS` for precisely this reason, and `.gitattributes` marks its event files `binary`. `stage_two/run_a_84_events/` gets neither, so it is relying on git's NUL heuristic under `core.autocrlf=true`.

### 7. `scripts/analysis/plot_epoch_curve.py:132` — positional list indexed by epoch label

`excess` is built over `epochs = sorted(curve)` but indexed `excess[e - 1]` at lines 132/135/138 and `excess[0]` at 174. That is only correct when the epochs are exactly `1..N` contiguous. Point `EVENTS_DIR` at a partial copy holding only the later event files and line 132 raises `IndexError`; a curve merely missing epoch 1 is worse, because it silently reads the wrong epoch's value into the basin annotation and the suptitle. Key it by epoch.

### 8. `plot_epoch_curve.py:68` — crashes on any run longer than 8 epochs

`paper = [PAPER_RUN_VAL_LOSS.get(e) for e in epochs]` yields `None` past epoch 8, and line 84's `[v * 1e4 for v in paper]` then raises `TypeError`. Run B is planned in this same doc at 30–60 epochs. It should skip absent paper points, not fault.

### 9. `stage2_epoch_curve.py:195` — `STEPS_PER_EPOCH` is never validated against the data

`round(step / STEPS_PER_EPOCH)` silently mislabels epochs when the hardcoded 9378 does not match the run — a different train-set size, or a world size other than 16, gives plausible-looking wrong epoch numbers instead of an error. `stage2_train_cost.py` explicitly leans on validation scalars landing on exact multiples of 9,378; assert it here too.

Relatedly, `main()` emits one row per *step* while `plot_epoch_curve.collect()` dedupes per *epoch*, so two steps mapping to one epoch would produce a 9-row CSV and an 8-point plot from the same data.

### 10. The scoring command is missing the arguments it needs on the host the doc names

`python evaluate.py --checkpoint ... --dataset manual` relies on the defaults `--data-root ../dataset` and `--manual-labels ../manual_labels`. `--dataset manual` needs both, since it pairs `manual_labels/` with the panos under `<data-root>/test` — and the doc says scoring runs on makelab2 while the dataset is staged on klone's `/gscratch/scrubbed`. As written it raises `FileNotFoundError("No image/label pairs found.")`.

### 11. The noise floor is hardcoded although both values are in the committed events

`EPOCH5_REPEATS = (0.00045980, 0.00045976)` is a rounded transcription. The events carry them exactly:

| file | step | value |
| --- | ---: | ---: |
| `events.out.tfevents.1786874180.g3047.1405499.0` | 46890 | 0.00045980451977811754 |
| `events.out.tfevents.1786888765.g3045.1660933.0` | 46890 | 0.0004597632505465299 |

So the figure's "measurement floor 0.009%" and "91× below" annotations are not derived from the committed data. The true values are **0.00897%** and **88×**. Deriving them makes the requeue table in the doc self-checking as a side effect.

---

Fixes for all eleven to follow on this branch. #1 and #2 collapse into one change — default the script at the committed paper events, drop the hardcoded table, regenerate the CSV and the figure — which also dissolves #11.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m], code-review effort: high)

jonfroehlich:
## All eleven fixed — `32df63e`

Also merged `main` into this branch first (`a5d398d`). That turned out to be load-bearing rather than housekeeping: the branch was cut at `911a957`, **before** the rescued paper-run events landed on main in #104, which is exactly why finding 1 read as plausible from inside the branch and false from the merge target.

### The two headline findings collapse into one change

`PAPER_RUN_VAL_LOSS` is gone. The paper column is now read from `docs/data/rampnet1_stage2_run/` at full float32 precision on every run. Corrected numbers, all re-derived:

| | as published | corrected |
| :--- | ---: | ---: |
| largest per-epoch delta | 1.69% | **1.755%** (epoch 8) |
| mean absolute delta | 0.61% | **0.579%** |
| stated bound in the doc | "no more than ~1.7% at any epoch" | **1.76%** |

The doc now carries an explicit note that an earlier draft claimed the paper column was not re-derivable, and that the claim was wrong — the omission is content, and so is the correction.

One thing the full-precision read surfaces that the 3-s.f. table had hidden: **the deltas are not uniform.** Epochs 1–6 agree to within 0.51%, then 7 and 8 diverge to 1.47% and 1.76%. That reads like accumulated floating-point drift rather than a recipe difference, but it does mean the tail of the curve is the weakest part of the replication claim, and the doc now says so.

### The rest

| # | finding | fix |
| ---: | :--- | :--- |
| 3 | `run_build_env.slurm` skip-guard poisons the env after a preemption | guard on a stamp written *after* verification; clear a partial prefix before rebuilding |
| 4 | runbook step 1b runs before the env exists, on a login node | backbone pre-warm moved inside the build job (non-fatal); step 2 gated with `--dependency=afterok` |
| 5 | second TFRecord parser beside a tested one | imports `stage2_train_cost.read_scalars` — one parser, one set of tests |
| 6 | no `SHA256SUMS`, no `.gitattributes` | manifest committed, events pinned `binary`, and **every read is now checked against it** rather than merely printing a digest |
| 7 | `excess[e - 1]` indexed by epoch label | keyed by epoch throughout |
| 8 | crashes past epoch 8 | missing paper points skipped; Run B at 30–60 epochs will plot |
| 9 | `STEPS_PER_EPOCH` never validated | `epoch_of()` refuses a step off the grid; `--steps-per-epoch` settable |
| 10 | scoring command missing `--data-root` | spelled out, with a note that makelab2 needs its own copy of the test split |
| 11 | noise floor hardcoded | derived per file from the events — which drops the figure's "91×" to the true **88×** |

### Verification

`tests/test_stage2_epoch_curve.py` is new and covers the three places a silent wrong answer was possible: step→epoch labelling, resume dedupe (later file wins), and manifest integrity. Full suite green — **1,003 passed, 1 skipped**.

I also drove the figure script through the cases findings 7 and 8 describe: a curve starting at epoch 5, a curve with epoch 4 missing, a 14-epoch curve, no paper events at all, and no requeue (so no noise floor to draw). All five render; before, the first three did not.

Two regressions of my own, caught and fixed while doing this: the reworked footnote initially grew to a fourth line and collided with the legend, and the suptitle still assumed epoch 1 was present after I had guarded the annotation for it.

PR #136, opened 2026-08-18
Actually fix the two collapsed runbook snippets from #129, and guard the class

jonfroehlich:
Review finding 14 on #129 was reported fixed and was not. Correcting the record along with the code.

**What happened.** The patch script that fixed it was a quoted bash heredoc, which strips one level of backslash — so the replacement text's trailing `\` + newline became a *Python line continuation* and the newline was removed. The output was the same collapsed line it was meant to repair. The `--op-threshold` correction in the same replacement did apply, so the diff looked right.

Both snippets now carry real continuations (written via `chr(92)`, so no layer can eat them):

```
python scripts/model_comparison/compare.py benchmark/richmond \
    --models rampnet,vistas:curb-cut,vistas:curb-cut+curb
```

**The guard matters more than the two lines.** A collapsed continuation still executes, still reads as fine, and its only symptom is that a reader copying the wrapped form gets a broken command — so nothing catches it, including a careful review. `test_no_runbook_snippet_has_a_collapsed_line_continuation` walks every shell block in the runbook docs and rejects run-on spaces mid-command, excluding the two forms that legitimately use them: aligned trailing comments, and lines that still carry their backslash. It found nothing beyond these two.

Suite 1,238 → 1,239.

### DIFF
--- a/docs/model_comparison.md
+++ b/docs/model_comparison.md
@@ -1155,7 +1155,8 @@ driveway aprons should appear as a characteristic false-positive mode — which
 recall hides on the other side of it is measurable:
 
 ```bash
-python scripts/model_comparison/compare.py benchmark/richmond     --models rampnet,vistas:curb-cut,vistas:curb-cut+curb
+python scripts/model_comparison/compare.py benchmark/richmond \
+    --models rampnet,vistas:curb-cut,vistas:curb-cut+curb
 ```
 
 The spec's `model_id` slot carries the **class set**, not a model id — the checkpoint comes
@@ -1312,7 +1313,8 @@ No new launcher; the arm needs nothing beyond the `transformers` + `torchvision`
 models already use, and Mask2Former is in-library (no `trust_remote_code`).
 
 ```bash
-PYTHON=$ENVPY MODELS=rampnet,vistas:curb-cut BUNDLE=benchmark/richmond     sbatch -A <account> scripts/model_comparison/run_open_models.slurm
+PYTHON=$ENVPY MODELS=rampnet,vistas:curb-cut BUNDLE=benchmark/richmond \
+    sbatch -A <account> scripts/model_comparison/run_open_models.slurm
 ```
 
 ## Status
--- a/tests/test_roster.py
+++ b/tests/test_roster.py
@@ -6,6 +6,8 @@ pass, and the failure they prevent is silent — the numbers simply change.
 """
 import json
 import pathlib
+import re
+import subprocess
 
 import pytest
 
@@ -85,6 +87,72 @@ def _table_lines(text):
         yield line
 
 
+def _blank_quoted(line):
+    """Replace the contents of "..." and '...' spans with x, keeping the width.
+
+    Quoted text is data, not command structure: a run of spaces or a ``#``
+    inside it says nothing about a lost continuation, and left alone the first
+    produces a false positive and the second hides everything after it."""
+    out = []
+    quote = None
+    for ch in line:
+        if quote is None:
+            out.append(ch)
+            if ch in "\"'":
+                quote = ch
+        else:
+            out.append(ch if ch == quote else "x")
+            if ch == quote:
+                quote = None
+    return "".join(out)
+
+
+def _collapsed_continuations(text):
+    """Yield (line_no, line) for every shell line that looks like a lost backslash.
+
+    Three spaces is the threshold because these runbooks indent continuations by
+    four, so a swallowed trailing backslash leaves at least that indent inside
+    one line; a flush-left or two-space continuation would collapse below three
+    and is not used here.
+    """
+    in_shell = False
+    for n, line in enumerate(text.splitlines(), 1):
+        stripped = line.lstrip()
+        if stripped.startswith("```"):
+            # lstrip, so a fence nested in a list item is entered and left.
+            in_shell = stripped.startswith("```bash") or stripped.startswith("```sh")
+            continue
+        if not in_shell or stripped.startswith("#"):
+            continue
+        scan = _blank_quoted(line)
+        # An aligned trailing comment legitimately uses run-on spaces.
+        scan = scan.split(" #", 1)[0].rstrip()
+        # A line that still HAS its continuation is correct by definition,
+        # whatever spacing it uses to align its arguments.
+        if scan.endswith("\\"):
+            continue
+        # Three or more spaces mid-command is what a swallowed trailing
+        # backslash plus newline leaves behind.
+        if re.search(r"\S {3,}\S", scan):
+            yield n, line
+
+
+def test_the_collapsed_continuation_scanner_catches_the_shape_it_claims():
+    """The scanner is the guard; these fixtures are the guard on the guard.
+
+    Planting a defect in a committed doc is the only other way to check that it
+    still catches the class, and that is not a thing to leave lying around."""
+    collapsed = "```bash\npython x.py --a b     --c d\n```\n"
+    assert list(_collapsed_continuations(collapsed)) == [
+        (2, "python x.py --a b     --c d")]
+
+    aligned_comment = "```bash\npython x.py --a b        # what b is for\n```\n"
+    assert list(_collapsed_continuations(aligned_comment)) == []
+
+    kept_continuation = "```bash\npython x.py --a b   \\\n    --c d\n```\n"
+    assert list(_collapsed_continuations(kept_continuation)) == []
+
+
 def test_no_runbook_snippet_has_a_collapsed_line_continuation():
     """A shell snippet whose trailing backslash was lost still *looks* fine -- the
     wrap becomes a run of spaces inside one long line -- and it still runs, so
@@ -92,31 +160,25 @@ def test_no_runbook_snippet_has_a_collapsed_line_continuation():
     exist to satisfy, and a reader copying the wrapped form gets a broken command.
 
     Two of them shipped in #126 because a patch script ate the backslashes.
+
+    Every tracked .md is walked, not the four runbooks the two defects happened
+    to be in: a collapsed snippet in README.md would be just as invisible.
     """
-    import re
-    for name in ("model_comparison.md", "replication.md", "operating_point.md",
-                 "adding_a_benchmark_city.md"):
-        path = REPO / "docs" / name
-        if not path.exists():
-            continue
-        in_shell = False
-        for n, line in enumerate(path.read_text("utf-8").splitlines(), 1):
-            if line.startswith("```"):
-                in_shell = line.startswith("```bash") or line.startswith("```sh")
-                continue
-            if not in_shell or line.lstrip().startswith("#"):
-                continue
-            # An aligned trailing comment legitimately uses run-on spaces.
-            line = line.split(" #", 1)[0].rstrip()
-            # A line that still HAS its continuation is correct by definition,
-            # whatever spacing it uses to align its arguments.
-            if line.endswith("\\"):
-                continue
-            # Three or more spaces mid-command is what a swallowed trailing
-            # backslash plus newline leaves behind.
-            hit = re.search(r"\S {3,}\S", line)
-            assert hit is None, (
-                f"docs/{name}:{n} looks like a collapsed line continuation: {line.strip()!r}")
+    hits = []
+    for path in _tracked_docs():
+        rel = path.relative_to(REPO).as_posix()
+        hits += [f"{rel}:{n} {line.strip()!r}"
+                 for n, line in _collapsed_continuations(path.read_text("utf-8"))]
+    assert hits == [], (
+        "these lines look like collapsed line continuations:\n" + "\n".join(hits))
+
+
+def _tracked_docs():
+    """Tracked .md paths. git ls-files, not a walk: the checkout can contain
+    nested worktrees and virtualenvs that a walk would descend into."""
+    listing = subprocess.run(["git", "-C", str(REPO), "ls-files", "-z", "*.md"],
+                             capture_output=True, text=True, check=True)
+    return [REPO / p for p in listing.stdout.split("\0") if p]
 
 
 def test_no_doc_still_hardcodes_the_old_roster_count():
--- a/tests/test_roster.py
+++ b/tests/test_roster.py
@@ -85,6 +85,40 @@ def _table_lines(text):
         yield line
 
 
+def test_no_runbook_snippet_has_a_collapsed_line_continuation():
+    """A shell snippet whose trailing backslash was lost still *looks* fine -- the
+    wrap becomes a run of spaces inside one long line -- and it still runs, so
+    nothing catches it. But the exact-commands-in-order rule is what these blocks
+    exist to satisfy, and a reader copying the wrapped form gets a broken command.
+
+    Two of them shipped in #126 because a patch script ate the backslashes.
+    """
+    import re
+    for name in ("model_comparison.md", "replication.md", "operating_point.md",
+                 "adding_a_benchmark_city.md"):
+        path = REPO / "docs" / name
+        if not path.exists():
+            continue
+        in_shell = False
+        for n, line in enumerate(path.read_text("utf-8").splitlines(), 1):
+            if line.startswith("```"):
+                in_shell = line.startswith("```bash") or line.startswith("```sh")
+                continue
+            if not in_shell or line.lstrip().startswith("#"):
+                continue
+            # An aligned trailing comment legitimately uses run-on spaces.
+            line = line.split(" #", 1)[0].rstrip()
+            # A line that still HAS its continuation is correct by definition,
+            # whatever spacing it uses to align its arguments.
+            if line.endswith("\\"):
+                continue
+            # Three or more spaces mid-command is what a swallowed trailing
+            # backslash plus newline leaves behind.
+            hit = re.search(r"\S {3,}\S", line)
+            assert hit is None, (
+                f"docs/{name}:{n} looks like a collapsed line continuation: {line.strip()!r}")
+
+
 def test_no_doc_still_hardcodes_the_old_roster_count():
     """These exact phrases were the drift. Catch them coming back.
 


jonfroehlich:
## Deep review

Re-ran from the branch head (`41facb3`) in a clean detached worktree. **The two snippets are fixed, the guard catches exactly the class it claims to, and the PR is still needed**: `main` is 23 commits ahead of this branch and both collapsed lines are still there (`docs/model_comparison.md:1192` and `:1349` on `main`). Checks that came back clean:

- Full suite: **1239 passed, 1 skipped**, as the body says.
- `git merge-tree` against `origin/main`: clean. Against `origin/exp/vistas-parity-126` (#137, which rewrites 270 lines of the same doc): clean. The merged tree passes the guard.
- The guard's exceptions are exercised, not decorative: across the four walked docs it scans 150 shell lines, 25 of which carry an aligned trailing comment and 32 of which end in a backslash.
- Run against `main`'s copy of the doc, the test fails on line 1192 (the first collapsed snippet); run against the branch's copy, it passes. Synthetic fixtures: a 3-space and a 5-space collapse are both caught, a collapse followed by an aligned comment is caught, a kept continuation with aligned arguments passes, an aligned trailing comment passes.
- The fixed `--models rampnet,vistas:curb-cut,vistas:curb-cut+curb` parses on both the branch and `main`: `parse_model_spec` is `partition(":")`, `VISTAS_CLASS_SETS` has `curb-cut+curb`, and `roster.label_for` resolves both specs on both refs. The slurm snippet's `PYTHON` / `MODELS` / `BUNDLE` are the variables `run_open_models.slurm` reads on both refs.
- "It found nothing beyond these two" is true for the four docs the test walks.

Everything below is about how far the guard reaches, not about what it does on the lines it reaches.

### 1. Medium-low — the guard walks four docs; the repo has eighteen with shell blocks

The test names `model_comparison.md`, `replication.md`, `operating_point.md` and `adding_a_benchmark_city.md`. On `main` those hold 25 of the 53 `bash` fences in the repo. Not walked: `README.md` (5 blocks, the front door), `docs/stage2_epoch_curve_84.md` (4), `data/inventories/README.md` (4), `scripts/model_comparison/yolo_baseline/README.md` (2), `docs/tillicum.md` (2), `docs/stage2_training_cost.md` (2), `docs/data_provenance.md` (2), the three HF card templates, `scripts/analysis/README.md`, `docs/stage1_generation_cost.md`, `docs/seam.md`, `docs/model_scoreboard.md`. A collapsed snippet in any of those is as invisible as the two this PR fixes were.

Running the same logic over every `.md` on `main` flags exactly one line outside the walked set, and it is a false positive: `data/inventories/README.md:45` is `echo "OK   $(basename "$f")"`, padding inside a quoted string. So broadening the walk costs one carve-out. Blanking quoted spans before the regex (`"…"` and `'…'`) removes that hit, and it also closes the hole in finding 2. The sweep then stays at zero on `main`, on the branch, and on the merged tree.

**Fix:** walk every tracked `*.md` instead of four names; blank quoted spans before matching; keep the sweep at zero.

### 2. Low — `split(" #", 1)` truncates at a `#` inside quotes

A quoted argument containing ` #` hides everything after it, so `python x.py --tag 'v #1'     --b c` passes. No line in the walked docs has that shape today. Blanking quoted spans first (finding 1) fixes it without a second exception.

### 3. Low — an indented fence is never entered

Fence detection is `line.startswith("```")`, so a fenced block inside a list item (two-space indent) is not scanned at all, opening or closing. `docs/replication.md:683` on `main` is one such block, a single line with an aligned comment, currently fine. **Fix:** `line.lstrip().startswith("```")` for both the opening and closing fence.

### 4. Low — the scanner is not itself testable

The scan is inline in the test body, so the only way to check that it still catches the class is to plant a defect in a committed doc. **Fix:** move the scan into a helper that takes text and returns `(line_no, line)` hits, and pin it with three inline fixtures: a 5-space collapse reports the right line number; an aligned trailing comment reports nothing; a kept backslash with aligned arguments reports nothing.

### Known limits, not fixes

- The 3-space threshold is right for the repo's 4-space continuation indent (a swallowed `\` leaves one trailing space plus the indent). A continuation that was flush-left collapses to one space and is indistinguishable from an ordinary line; a 2-space indent collapses to two. Neither is used in these docs. Worth one sentence in the docstring so the number is not read as arbitrary.
- `console` fences are not scanned. `shell` fences are, because `startswith("```sh")` matches them. Neither appears in the repo.
- A heredoc body with aligned columns inside a shell block would fire. None exists in any `.md` on `main`.

- **Placement.** `tests/test_roster.py` already holds two doc-consistency tests (`test_the_roster_table_in_the_docs_matches_the_registry`, `test_no_doc_still_hardcodes_the_old_roster_count`), so this one is not out of place. If the walk broadens to every `.md` (finding 1), a `tests/test_docs.py` is the natural home, and the two existing doc tests could move with it. Not worth doing on its own.

### What holds up

- Both snippets carry real continuations. The diff is exactly the two lines plus the test; no prose changed.
- The guard catches the defect on the real artifact (`main`'s doc) and on synthetic 3- and 5-space collapses, and its two exceptions are the right two: aligned comments and lines that still carry their backslash.
- Merge health is clean against `main` and against #137, which is the branch most likely to conflict.
- The fixed command runs as written on both refs; no argument-parser drift since the branch was cut.

## Fix list
- [F1] Medium-low — Walk every tracked `*.md` (via `git ls-files` or an rglob that skips `.venv`, `.claude`, `node_modules`) instead of the four names; blank out `"…"` and `'…'` spans before the regex so `data/inventories/README.md:45` does not fire and a collapse after a quoted `#` is not hidden; assert the sweep is zero. — files: tests/test_roster.py
- [F2] Low — Detect fences with `line.lstrip().startswith("```")` for both opening and closing, so list-nested blocks (e.g. `docs/replication.md:683` on `main`) are scanned. — files: tests/test_roster.py
- [F3] Low — Extract the scan into a helper returning `(line_no, line)` hits and pin it with three inline fixtures (5-space collapse hits the right line; aligned comment and kept backslash with aligned arguments do not). Add one docstring sentence deriving the 3-space threshold from the 4-space continuation indent. — files: tests/test_roster.py

jonfroehlich:
## Review fixes

All three in `tests/test_roster.py`, commit 3c16513.

- **F1 — fixed.** The guard walked four runbook names; it now walks every tracked `*.md` (`git ls-files`, not a filesystem walk, because the checkout can contain nested worktrees and virtualenvs). Quoted `"…"` and `'…'` spans are blanked before the regex, which drops the one false positive broadening would otherwise add (`data/inventories/README.md:45`, padding inside a quoted string) and closes F2 at the same time.
- **F2 — fixed.** `split(" #", 1)` no longer truncates at a `#` inside quotes, since the split now runs on the quote-blanked line. Fences are detected with `lstrip()` on both the opening and closing line, so blocks nested in a list item are scanned; there are 3 such fences in the merged tree.
- **F3 — fixed.** The scan is a module-level helper `_collapsed_continuations(text)` yielding `(line_no, line)`, pinned by `test_the_collapsed_continuation_scanner_catches_the_shape_it_claims` with three inline fixtures: a 5-space collapse reports line 2, an aligned trailing comment reports nothing, a kept backslash with aligned arguments reports nothing. Its docstring derives the 3-space threshold from the 4-space continuation indent.

Not applied: the placement question under "Decisions for Jon" — the test stays in `tests/test_roster.py`.

Sweep results, with the broadened walk:

- **This branch: zero** (25 tracked docs).
- **Merge with `origin/main`: zero.** Checked in a throwaway detached worktree of `origin/main` merged with this branch's head — 27 docs, 404 shell lines, merge clean.
- Run against `origin/main` alone it reports exactly two lines, `docs/model_comparison.md:1192` and `:1349` — the two this PR fixes, and nothing else across all of main's docs.

Full suite: **1240 passed, 1 skipped** (one more than before, the new scanner test).

jonfroehlich:

1 major / 4 minor / 2 nit confirmed, 1 rejected. The two doc fixes are correct and the guard does catch the class it was written for. The one thing to fix before merging: `main` has since acquired eight more collapsed continuations in `docs/replication.md` (via [#152](https://github.com/ProjectSidewalk/RampNet/issues/152)), this branch does not touch that file, so the PR's own new test fails on the merge result. Merge `origin/main` in, restore those eight backslashes, and the PR becomes its own best evidence that the guard is needed.

### Confirmed

#### S1 · major · replicability / CI — the new test fails on the merge with `main`
- **Where:** `tests/test_roster.py` — `def test_no_runbook_snippet_has_a_collapsed_line_continuation():`; `docs/replication.md:139–151` on `origin/main`, e.g. line 139 `python scripts/model_comparison/compare.py benchmark/laurens_mapillary     --models rampnet,owlv2,gdino --sweep`
- **What:** The branch fixes the two `model_comparison.md` lines and reports "it found nothing beyond these two". That was true on 2026-09-03. Commit `5111834` (2026-09-01) reached `main` through [#152](https://github.com/ProjectSidewalk/RampNet/issues/152) on 2026-09-04 and added eight lines with the same 5-space signature to the Laurens replication runbook. The Laurens runbook — exactly the exact-commands-in-order artifact the test protects — is shipped collapsed on `main` today.
- **Evidence:** Ran the branch's `_collapsed_continuations` over every tracked `.md` at each ref: `origin/main` 10 hits (`model_comparison.md:1329`, `:1486`, `replication.md:139, 140, 141, 142, 143, 146, 150, 151`); `HEAD` 0 hits. `git merge-tree --write-tree origin/main HEAD` produces a tree whose `docs/replication.md` blob is byte-identical to `main`'s (`eff8a9c`), so the merged tree carries all 8. `pytest -q tests/test_roster.py` on the branch alone: 31 passed.
- **Fix:** Merge `origin/main`, restore the eight continuations in `docs/replication.md:139–151` (written via `chr(92)`, per this PR's own lesson), confirm the sweep is zero on the merged tree, and update the PR body: 10 lines fixed, not 2, and two tests added, not one ("Suite 1,238 → 1,239" is stale either way). The existing `test_no_runbook_snippet_has_a_collapsed_line_continuation` is the test that pins it.
- **Found by:** Opus 5

#### S2 · minor · test coverage — untyped fences are not scanned, and five of them are shell runbook blocks
- **Where:** `tests/test_roster.py` — `in_shell = stripped.startswith("```bash") or stripped.startswith("```sh")`
- **What:** Only fences tagged `bash`/`sh*` are entered. Five untyped ```` ``` ```` blocks in the tree hold shell commands, two of them with backslash continuations; a collapse there is invisible to the guard, and the docstring's "Every tracked .md is walked" overstates the coverage.
- **Evidence:** Fence census over the 25 tracked `.md`: 54 `bash`, 8 `python`, 6 `bibtex`, 1 `py`, 1 `powershell`, 17 untyped. Untyped blocks with shell commands: `benchmark/README.md:42`, `benchmark/README.md:466`, `docs/crop_window_eval.md:9`, `docs/curb_ramp_data_sourcing.md:315` (one `\` continuation, `miss_gallery.py --bucket silent \`), `docs/curb_ramp_data_sourcing.md:844` (three, `temporal_gap.py --city "Washington, DC" \`). Probe: `_collapsed_continuations("```\npython x.py --a b     --c d\n```\n")` → `[]`; same text under ```` ```bash ```` → one hit.
- **Fix:** Retag those five blocks as ```` ```bash ```` in this PR, and add a fixture to `test_the_collapsed_continuation_scanner_catches_the_shape_it_claims` that pins the untyped-fence policy (scanned, or deliberately not) so the boundary is stated. Qualify the docstring: every file is walked, but only tagged shell blocks within each.
- **Found by:** Opus 5

#### S3 · minor · guard gap — backslash followed by trailing whitespace is accepted as a continuation
- **Where:** `tests/test_roster.py` — `scan = scan.split(" #", 1)[0].rstrip()` then `if scan.endswith("\\"): continue`
- **What:** `rstrip()` runs before the backslash check, so `cmd \␠⏎` passes as a kept continuation. In bash that is an escaped space, not a continuation; the next line runs as a separate command. Same "still executes, only the copier notices" class this PR fixes.
- **Evidence:** Probe: `_collapsed_continuations("```bash\npython x.py --a b \\ \n    --c d\n```\n")` → `[]`. Bash: a file containing `echo a \␠` / `echo b` prints `a␠␠` then `b` (two commands). `grep -E '\\[ \t]+$'` over tracked `.md` on both refs: zero instances, so no doc is wrong today — this is a gap in the guard, not a live defect.
- **Fix:** Before the `rstrip`, `if re.search(r"\\[ \t]+$", line): yield n, line`, plus a fourth fixture asserting that shape is reported.
- **Found by:** Opus 5, Gemini 3.1 Pro

#### S4 · minor · guard gap — a line that still ends in `\` is skipped even when it has a collapse earlier in it
- **Where:** `tests/test_roster.py` — `if scan.endswith("\\"):`
- **What:** Losing the first of two continuations in a three-line command yields `a     b \` / `    c`; line 1 ends with `\` and is skipped unconditionally, so the collapse is missed.
- **Evidence:** Probe: `_collapsed_continuations("```bash\npython tool.py --foo bar     --baz qux \\\n    --c d\n```\n")` → `[]`. The obvious fix — strip the trailing backslash, then apply the regex — flags exactly one existing line: `docs/replication.md:504` `python scripts/export_benchmark.py push   --out dist/rampnet-benchmark \`, where the three spaces align the `build`/`verify`/`push` subcommands (blame `a109ce7`, 2026-08-04). So the fix costs one deliberate alignment.
- **Fix:** Either single-space `replication.md:504` and scan the pre-backslash part of continued lines (with a fixture), or leave the skip and record the gap in the docstring. The former is one keystroke and closes a real hole.
- **Found by:** Gemini 3.8 Flash

#### S5 · minor · scanner false positives — quote masking is per-line and escape-unaware
- **Where:** `tests/test_roster.py` — `if ch == quote:` in `_blank_quoted`; `scan = _blank_quoted(line)`
- **What:** An escaped quote (`\"`) closes the mask, so spaces after it are scanned as command structure; and quote state resets each line, so the inside of a multi-line quoted string is scanned as command. Both directions produce false positives only — the test fails loudly with the line named — so the blast radius is a spurious red run, not a hidden defect.
- **Evidence:** Probes: `echo "Starting \"   Phase 2\""` → flagged (mask output `echo "xxxxxxxxxx"   Phase 2\""`); a `curl -d '{` / `    "key":   "value",` / `}'` block → line 2 flagged. In the tree today: zero escaped quotes inside bash blocks; one multi-line quoted string, `docs/model_comparison.md:1686–1689` (`$ENVPY -c 'from huggingface_hub import ...` spanning four lines), whose inner lines happen not to contain a 3-space run, so nothing fires now.
- **Fix:** Handle `\` inside `_blank_quoted` (skip the next character) and carry the quote state across lines within a block, or document both as known false-positive shapes in the docstring so the first person who hits one does not weaken the guard. Two fixtures either way.
- **Found by:** Gemini 3.1 Pro (both), Gemini 3.8 Flash (escaped quote)

#### S6 · nit · docstring — the threshold justification is wrong as stated
- **Where:** `tests/test_roster.py` — `a flush-left or two-space continuation would collapse below three and is not used here.`
- **What:** Both are used, and the two-space case is caught for a different reason than given.
- **Evidence:** Continuation-indent census over bash blocks: 4 spaces × 42, 2 × 6, 0 × 2 (`docs/tillicum.md:489–494`, `:500–501`), plus 3, 6, 7, 9, 17, 19. A two-space continuation collapses to `␠\⏎` + 2 = 3 spaces and *is* caught (probe: `foo --a b   bar` → flagged). A flush-left collapse leaves 1 space and is missed (probe: `PYTHON=x YOLO=y sbatch` → not flagged) — harmless there, since the run-on env-prefix line executes identically.
- **Fix:** Reword: ≥3 because a swallowed `␠\⏎` plus the indent leaves ≥3 for every indent ≥2 in these docs; the two flush-left env-prefix continuations in `docs/tillicum.md` would collapse to one space and are not covered.
- **Found by:** Opus 5

#### S7 · nit · references / placement
- **Where:** `tests/test_roster.py` — `Two of them shipped in #126 because a patch script ate the backslashes.`; module docstring `The challenger registry (#122): ...`
- **What:** The two lines shipped in [#129](https://github.com/ProjectSidewalk/RampNet/issues/129) (the Vistas PR); [#126](https://github.com/ProjectSidewalk/RampNet/issues/126) is the issue it closed — the PR body itself says "Review finding 14 on #129". And a repo-wide Markdown lint that shells out to `git ls-files` now lives in the roster-registry module, whose docstring says the file is about `WITNESS_POOL_46` and roster/doc consistency. The suite also now needs `git` on PATH (fine under `actions/checkout`; outside a checkout `_tracked_docs` raises `CalledProcessError` rather than skipping).
- **Fix:** `#126` → `#129` (or "#129, for #126"). Consider `tests/test_docs.py` for `_blank_quoted`, `_collapsed_continuations`, `_tracked_docs`, the two new tests and the existing `test_no_doc_still_hardcodes_the_old_roster_count`. Not blocking.
- **Found by:** Opus 5

### Rejected or downgraded
- **Gemini 3.8 Flash: `git ls-files "*.md"` matches only the repository root, so `docs/*.md` is never scanned** — rejected. Git pathspecs glob across directory separators by default. `_tracked_docs()` on the branch returns 25 paths, 23 of them in subdirectories (`benchmark/README.md`, `docs/model_comparison.md`, ...), and the same scanner run over `origin/main` found the 10 hits above, all under `docs/`.

### Coverage
- Files excluded from the Gemini reviews: none (both logs: 2 files, 138 diff lines, 0 untracked, tree clean). Both runs exited 0 with the end marker verified (Pro 244 s, Flash 149 s; 3 findings each).
- Tests run: `pytest -q tests/test_roster.py` on the branch → 31 passed. Opus additionally ran the full suite on the branch → 1,240 passed, 1 skipped. The branch scanner over every tracked `.md` at `origin/main` (10 hits) and `HEAD` (0). `git merge-tree` of `origin/main` + `HEAD` to confirm the merge carries `main`'s `replication.md`. Nine inline scanner probes (untyped/bash/shell fence, `\`+space, collapse-before-`\`, escaped quote, multi-line string, 2-space and flush-left collapse). A two-line bash script for the `\␠` semantics. Fence-language, escaped-quote, odd-quote-count, and continuation-indent censuses over bash blocks.
- Not checked: the runtime behaviour of the collapsed commands themselves (the one-line forms execute; the defect is the wrapped-copy one the PR describes); `_blank_quoted` against an apostrophe in an unquoted word (`docs/adding_a_benchmark_city.md:139` has one, but it sits after ` #` so the comment split still fires — not made certain in general).

---

jonfroehlich:

`origin/main` merged in (4e08ce6, plain merge, no rebase), then five commits, 3c16513..33b48da.

- S1 — fixed in 33a1f27: the eight collapsed lines in `docs/replication.md:139–151` (ten continuations, from [#152](https://github.com/ProjectSidewalk/RampNet/pull/152)) restored via `chr(92)`; the guard passes on the merged tree. The scanner over `origin/main` today reports 11 lines (the 10 collapsed plus the S4 alignment), over this branch 0. PR body updated: 10 lines fixed, not 2. Test: `test_no_runbook_snippet_has_a_collapsed_line_continuation`.
- S2 — fixed in 622e332 + 33b48da: the five untyped shell blocks (`benchmark/README.md` ×2, `docs/crop_window_eval.md`, `docs/curb_ramp_data_sourcing.md` ×2) are tagged `bash`; the fence policy is a named constant `_SHELL_FENCES` (bash/sh/shell/zsh; untyped deliberately not scanned, because in this tree they hold tree listings and program output); the sweep docstring says "every file, tagged shell blocks only". Test: per-tag fixtures in `test_the_collapsed_continuation_scanner_catches_the_shape_it_claims`.
- S3 — fixed in 33b48da: `\` followed by trailing whitespace is reported as its own shape, before the `rstrip`. Test: `escaped_space` fixture in the same test.
- S4 — fixed in 33b48da: a line that keeps its last `\` is scanned up to the backslash; the one line that flagged, the build/verify/push trio in `docs/replication.md`, is single-spaced. Test: `collapse_before_backslash` fixture.
- S5 — fixed in 33b48da: the mask runs over the whole block (multi-line strings stay masked) and follows bash's backslash rules (an escaped quote does not close the span). Doing it across lines exposed one more shape — an apostrophe in a comment (`# don't`) would have masked every line after it — so a `#` that starts a word outside quotes is copied through as a comment. Tests: `escaped_quote`, `multi_line_string`, `comment_apostrophe` fixtures, plus `test_blank_quoted_keeps_width_and_newlines`.
- S6 — fixed in 33b48da: docstring rewritten — the threshold holds for every indent ≥ 2, and the two flush-left env-prefix continuations in `docs/tillicum.md` are named as not covered. Test: `two_space_indent` fixture.
- S7 — fixed in 6f8ea2e: `#126` → `#129 (for #126)`; the scanner, `_tracked_docs` and `test_no_doc_still_hardcodes_the_old_roster_count` moved verbatim to a new `tests/test_docs.py`, whose docstring states the `git` dependency. One thing found in the move: the roster-count test's second pattern held a literal backspace byte (0x08) where `\b` should be, so it could never match — the guard was inert on `main`. Written as `\b` it passes on today's docs (`all 83 far-field silent misses` is not a hit precisely because of the boundary).

Not changed: a fence inside a blockquote (`> ```bash`, one in `docs/replication.md`) is not entered by the scanner; recorded in `_shell_blocks`'s docstring, not widened here.

Suite: 1,484 passed, 1 skipped (`main`: 1,481 passed, 1 skipped; this branch adds three tests).

jonfroehlich:

0 major / 1 minor / 2 nit confirmed, 2 rejected or downgraded. All seven round-1 items (S1–S7) are verified as properly fixed on `33b48da`: the sweep over `origin/main` reports 11 lines (the 10 collapsed plus the S4 alignment) and 0 on this branch, the ten restored continuations are real backslashes (`cat -A` shows `\$`), and a mutation check that reverts each scanner fix in turn makes the fixture test fail for S2, S3, S4 and S5 (both halves). The one round-1 fix that is not pinned is the S7 `\b` repair: putting the backspace byte back leaves every test green, which is the minor below. The single most important thing to fix: add a self-test for `stale[1]` and decide whether the now-live "all 8" guard is meant to reject every non-roster axis or only the two it names.

### Confirmed

#### S1 · minor · test infrastructure — the `\b` repair is unpinned and armed a guard broader than its own comment
- **Where:** `tests/test_docs.py` — `stale = (r"all 8 (?:model|challenger|zero-shot)", r"all 8\b(?! splits| cities)",`
- **What:** On `main` the second pattern held a literal 0x08 and could never match, so this branch is the first time it is live. Nothing asserts what it matches. And the comment above it says the count "has to be bound to the roster, or the guard fires on perfectly good prose", but the pattern is bound to nothing: it rejects every "all 8 <anything>" except `splits` and `cities`, and this repo already writes about other eights.
- **Evidence:** Probe of the committed pattern: `all 8 of them` → hit; `all 83 far-field` → no hit (the `\b` working); `all 8 splits` / `all 8 cities` → no hit; `all 8 seeds`, `all 8 epochs`, `all 8, and` → hit. `docs/stage2_run_b_power_135.md:460` contains `all 8 epochs` today and passes only because that file is not in the scanned set. Mutation: replacing `\b` with `\x08` in a copy of the test file → 4 passed, 0 failed. The only record of the repair is the comment at lines 252–255, the same evidence class that hid the byte on `main`.
- **Fix:** A small `test_stale_roster_patterns_match_what_they_claim` asserting `stale[1]` matches `"all 8 of them"` and rejects `"all 83 far-field silent misses"`, `"all 8 splits"`, `"all 8 cities"`; then either extend the lookahead with the known non-roster axes (`epochs|seeds|checkpoints`) or restate the comment so it says the guard deliberately rejects every "all 8" that is not splits/cities.
- **Found by:** Opus 5

#### S2 · nit · scanner — an inline comment introduced by a tab is not cut before the scan
- **Where:** `tests/test_docs.py` — `scan = scan.split(" #", 1)[0].rstrip()`
- **What:** bash starts a comment at a `#` preceded by any blank, and `_blank_quoted` already agrees (`text[i - 1] in " \t\n"`), but the cut in `_collapsed_continuations` looks for `" #"` only. A `cmd<TAB>#   why` line keeps its comment, and the spaces inside the comment fire the collapse regex. False positive only; the test fails loudly with the line named.
- **Evidence:** bash: `echo A<TAB>#   comment here` prints `A`. Scanner: `_collapsed_continuations("```bash\ncmd --a b\t#   why this\n```\n")` → `[(2, 'cmd --a b\t#   why this')]`. Census: no `\t#` in any shell block at HEAD.
- **Fix:** `re.split(r"[ \t]#", scan, 1)[0]`, plus a one-line fixture, so the two comment definitions in the file agree.
- **Found by:** Gemini 3.1 Pro (Opus 5 probed the same shape and did not raise it)

#### S3 · nit · consistency — the blockquoted `adopt`/`build`/`verify`/`push` trio still carries the alignment S4 removed
- **Where:** `docs/replication.md:328` — `> python scripts/export_benchmark.py push   --out dist/rampnet-benchmark --message "..."`
- **What:** S4 single-spaced the trio at lines 561–564 so the scanner accepts it, but the same commands inside the blockquote at 325–328 keep their column alignment. The doc now shows the command two ways, and one of them is exactly the shape the guard rejects; it passes only because `> ```bash` fences are not entered (the gap recorded in `_shell_blocks`).
- **Evidence:** Feeding lines 325–328 to the scanner under a plain ```` ```bash ```` fence → `(5, '...push   --out ...')`; as written (blockquoted) → `[]`.
- **Fix:** Single-space the four blockquoted lines now, so the blockquote gap stays a coverage gap and not a latent failure the day someone strips the `> ` prefix in `_shell_blocks`.
- **Found by:** Opus 5

### Rejected or downgraded
- **Gemini 3.8 Flash: `git ls-files "*.md"` matches only the repository root, so `docs/` is never scanned** — rejected, for the second round running. Git pathspec globs match across `/`. `_tracked_docs()` on this branch returns 32 paths, 30 of them in subdirectories (`benchmark/README.md`, `docs/model_comparison.md`, ...), and the same scanner run over `origin/main` reports the 11 hits listed above, all under `docs/`.
- **Gemini 3.1 Pro: a `#` after an escaped space (`echo \ #foo`) is wrongly treated as a comment, hiding a collapse later on the line** — partly valid, downgraded to a docstring note. The semantics are right: bash prints `B  #foo` for `echo B \ #foo`, so `#foo` is a word, and both `_blank_quoted` (checks `text[i - 1]`, which is the escaped space) and the ` #` cut treat it as a comment. But the shape does not occur in any shell block at HEAD, it is a false negative only when the same line also has a collapse after it, and the fix is not a one-liner — the mask has to keep the backslash visible for the `\␠` end-of-line check while hiding the escaped space from the comment cut. Record it as a known limit next to the heredoc and `$'...'` ones; not worth code.

### Coverage
- Files excluded from the Gemini reviews: none. Both logs: 7 files, 458 diff lines, 0 untracked, tree clean, base `origin/main` (2916ba5). Pro exited 0 (265 s, 116,992 tokens, 2 findings, end marker verified). Flash exited 0 on its second run (166 s, 84,790 tokens, 1 finding, end marker verified); the first Flash attempt exited 4 — no usable review, a headless tool-permission denial — and was re-run; the files on disk are from the successful run.
- Tests run: `pytest -q tests/test_docs.py tests/test_roster.py` → 32 passed. Full suite from Git Bash → 1,484 passed, 1 skipped (114 s), matching the PR body. (From PowerShell the same suite shows 25 failures, all in `tests/test_slurm_scripts.py`, because `bash` resolves to the WSL launcher there and gets Windows paths; pre-existing, not this PR's.) Scanner sweep over every tracked `.md` at `origin/main` (11 hits) and HEAD (0). Six mutants of the scanner against the fixture tests (S2, S3, S4, S5×2 caught; S7 `\b` not caught). Bash semantics for tab-before-`#` and `\ #`. Probes of `stale[1]` against seven phrases and a grep of `all 8\b` across `docs/`.
- Not checked: whether the ten restored Laurens commands run (they are the legs already executed under [#152](https://github.com/ProjectSidewalk/RampNet/pull/152); the defect class is the wrapped-copy one); `git` availability under CI beyond the `actions/checkout` assumption in the module docstring; the `main` test count of 1,481 quoted in the fixes comment (three new tests on this branch, one moved, consistent with 1,484).

---

**Merge readiness:** ready — all seven round-1 fixes are confirmed, the branch sweep is zero, and the three items above are a missing self-test and two nits; none changes a committed number or a runbook, so they can land here or in a follow-up.

jonfroehlich:

Three commits, 33b48da..4583450.

- S1 — fixed in 93fcba2: the stale-roster patterns are a module constant `STALE_ROSTER_PHRASES`, and `test_stale_roster_patterns_match_what_they_claim` asserts no pattern carries a 0x08 byte, that `all 8 of them` is a hit, and that `all 83 far-field silent misses`, `all 8 splits`, `all 8 cities`, `all 8 epochs`, `all 8 seeds`, `all 8 checkpoints` are not; putting the backspace byte back fails it (checked by mutation). The exclusion is widened to the axes these docs count (`splits|cities|epochs|seeds|checkpoints`) and the comment now says the guard is bound by exclusion on purpose: a false hit costs one edit to the tuple, a miss is a stale number in a published doc. Test: `test_stale_roster_patterns_match_what_they_claim`.
- S2 — fixed in 7827638: the comment cut is `re.split(r"[ \t]#", ...)`, so a tab-introduced trailing comment is dropped before the scan, matching `_blank_quoted` and bash. Test: `tab_comment` fixture in `test_the_collapsed_continuation_scanner_catches_the_shape_it_claims`.
- S3 — fixed in 4583450: the blockquoted `adopt`/`build`/`push` lines in `docs/replication.md` are single-spaced to match the trio at the bottom of the file; the blockquote fence gap stays a coverage gap, not a latent failure.

Not changed, per the downgrade: `echo \ #foo` (a `#` after an escaped space) is read as a comment by both the mask and the cut; no shell block has the shape, and it is a false negative only when the same line also holds a collapse after it.

Suite: 1,485 passed, 1 skipped (round 1 was 1,484; one test added).

Issue #130, opened 2026-08-18
360 seam, in the scoring code: 11 double-marked ramps in manual_gold, and a matcher that does not wrap


jonfroehlich:
Found while reviewing #99, whose Phase 1 window wraps at the seam and the matcher's does not. Chasing that divergence turned up a second, live defect underneath it. Two separate things, same geometry:

## 1. Eleven ramps are double-marked across the seam in `manual_gold` — live, measurable

`manual_gold`'s GT comes from the independent 1,000-pano YOLO labeling pass (`manual_labels/*.txt` → `rampnet/detection_eval.py::yolo_ground_truth`), which is a **different annotation path from the nine verdict-reviewed splits**. #43's reviewer rule ("model hit either side → one TP, do not add a missed mark on the other side") was written for the GT viewer and the verdict schema; it never applied here, and `yolo_ground_truth()` does no dedup of its own — every `class cx cy w h` line becomes a point.

Result: one physical ramp straddling x≈0 / x≈1 is labeled **twice**, once per half. The model emits one detection, claims one mark, and the other mark is an unclaimable false negative.

Scanned across every committed `analysis_out/op_cache` split — pairs that are distinct without the wrap and the same ramp with it (wrapped separation < the 22.53 px match radius):

| split | GT | wrapped-duplicate GT pairs |
| :--- | ---: | ---: |
| richmond, bend, clovis, morgantown, annapolis, paterson, gainesville, budapest_district5, sao_paulo | 2,641 | **0** |
| `manual_gold` | 3,919 | **11** |

Zero in the nine reviewed splits, which is the expected result and a nice confirmation that the #43 rule worked where it was applied. All eleven are in `manual_gold`:

| `manual_labels/<pano>.txt` | x of mark A | x of mark B | wrapped separation |
| :--- | ---: | ---: | ---: |
| `2NgKmkIoU9nUwvjk6K5wtw.txt` | 0.0021 | 0.9897 | 12.8 px |
| `JrR9wwG_ynJrSP9ov3Rxrw.txt` | 0.0067 | 0.9971 | 9.9 px |
| `MJbCTDJV5advy8eYbL0V5Q.txt` | 0.0023 | 0.9854 | 17.3 px |
| `XCbhS0jxrSrOLyndJXtQvA.txt` | 0.9942 | 0.0004 | 6.3 px |
| `XFxls0YjTNlFLpZ6eNUuwg.txt` | 0.0081 | 0.9992 | 9.2 px |
[...]

PR #137, opened 2026-08-18
Vistas at resolution parity: the handicap was real, and it was recall (#126)

jonfroehlich:
Closes the open caveat on the Mapillary Vistas arm from #129: the published richmond numbers were
measured at 384×384 while every other tiled leg sees 1024×1024, so the 0.338 F1 gap to RampNet was
an upper bound, not a measurement. `--vistas-input-size` landed in #129 for exactly this run.

**Docs only.** No code, no detection artifacts, no roster change.

### Pre-registered read

Posted to #126 **before** the scored output was read (comment 5334506091): gap closes by
**< 0.05 F1** ⇒ *"transfers but does not compete"* stands, one split, stop. More ⇒ revise and cost
3–4 further splits.

### It stands, and not narrowly

RampNet re-ran alongside as the comparability check and reproduced its committed richmond row to
every digit (0.964 / 0.768 / 0.855, AP 0.763, 238/9/72).

| arm | model input | P | R | F1 | AP | tp/fp/fn |
|---|---|---:|---:|---:|---:|---|
| rampnet (committed, reproduced) | — | 0.964 | 0.768 | **0.855** | 0.763 | 238/9/72 |
| vistas curb-cut — published (RTX 3070) | 384×384 | 0.411 | 0.697 | 0.517 | 0.513 | 216/309/94 |
| vistas curb-cut — same-env control (A40) | 384×384 | 0.411 | 0.694 | 0.516 | 0.510 | 215/308/95 |
| vistas curb-cut — **parity** (A40) | **1024×1024** | 0.383 | **0.884** | **0.534** | **0.649** | 274/442/36 |

F1 **0.516 → 0.534, +0.018** — about a third of the bar. RampNet's lead goes 0.339 → **0.321**.

### The mechanism is the finding, and it is not what the caveat predicted

The handicap was real and large. It was almost entirely a **recall** handicap:

- **Recall 0.694 → 0.884 (+0.190)**; misses fall 95 → **36**, a 62% reduction. At parity this arm
  **out-recalls RampNet** (0.884 vs 0.768) while never having seen one of our labels.
- **AP 0.510 → 0.649 (+0.139)**, 27% relative.
- **Precision 0.411 → 0.383 (−0.028)** — slightly *worse*. FPs rise faster than TPs.

F1 stayed flat only because precision is the binding constraint and resolution does nothing for
it. That **sharpens** the conclusion #126 was built to test: *the concept is findable, the
discrimination is not* now holds against the supervised arm at equal input, no longer confounded
with how many pixels the model was given.

### An env control that wasn't in the plan, and earned its place

The published run was on an RTX 3070 on an older `transformers`; makelab2 is a major version on,
and the `transformers` version is **not** in the detection signature — a hazard the doc already
flags. So parity-vs-published would have differed in *two* things.

Re-running 384 in the parity env separates them: it lands within **one detection out of 523** of
the published run (215/308/95 vs 216/309/94, F1 0.516 vs 0.517). The 4.x→5.15 jump is benign for
this checkpoint, the residual is fp16 kernel nondeterminism, and the whole parity delta is
attributable to input size. That also retires *"an upgrade could have changed every mask under an
unchanged cache key"* as a measurement rather than an assurance.

### Two claims at 384 that parity corrects

- The YOLO AP comparison said richmond AP 0.748 / 0.724 / 0.536 are *"all above 0.513"*. At parity
  the arm is **0.649**, so **`y26_pano` (0.536) no longer clears it**. The sentence's conclusion
  holds — the two stronger YOLO arms still lead — its arithmetic does not.
- **`min_area_px=16` is inert at 384 but not at 1024.** At 384 the mask upsamples 10.67× so the
  smallest blob is ~114 px; at 1024 it upsamples 4× so the smallest blob is ~16 px, exactly the
  floor. The two rows share the value, not its meaning. Left untouched so as not to confound the
  one variable.

### Deliberate omissions

- **Parity detections are not published.** `--vistas-input-size` does not change the arm's label,
  so exporting would overwrite `mask2former-vistas-curb-cut__richmond.json`. Publishing them needs
  a distinct name (the `--publish-as` pattern from #123). Out of scope here.
- **richmond only**, per the decision rule.
- **richmond only** for the complementarity read below too.

### Verification

- Ran on makelab2 (A40, fp16, transformers 5.15.0 / torch 2.13.0+cu130) in a scratch worktree with
  a private `--cache-dir`, so nothing shared a cache directory with the published detections.
- **Instrument checked before the run, not assumed:** the override reaches the model —
  `pixel_values` (1, 3, 384, 384) → (1, 3, 1024, 1024), mask logits (1, 100, 96, 96) →
  (1, 100, 256, 256). A silently no-opping `processor.size` on a new major version would have made
  "parity" a second 384 run wearing a different cache key. Also re-confirmed the 65-class head
  (class 9 `Curb Cut`, class 2 `Curb`).
- **Cost measured, not extrapolated** — the estimate going in was 3–4× and was wrong in the cheap
  direction: full 124-pano run **3m38s** on one A40; GPU forward 0.078 → 0.092 s per view
  (**1.17×**), because Swin's windowed attention scales far better than pixel count. Peak GPU
  memory 1.64 GB.
- `pytest -q`: **1238 passed, 1 skipped**, including #136's runbook line-continuation test against
  the new shell block.

---

## Follow-on in this PR: complementarity (the #35 gate)

The parity result raised the recall-first question, so it is answered here rather than filed.
Scoring-side only, on the same cached detections.

| | vistas @384 | **vistas @1024 (parity)** |
|---|---:|---:|
| **challenger ONLY** (rampnet-miss ∩ hit) | 21 | **54** |
| found by NEITHER | 51 | **18** |
| of rampnet's 72 misses, recovered | 21 (29%) | **54 (75%)** |
| null on that subset | 0.090 | 0.143 |
| **attributable after the null** | **~15** | **~44** |
| oracle-union recall | 0.835 | **0.942** |
| boxes/pano · above chance | 4.5 · 0.657 | 6.2 · **0.864** |

**Discounted for chance, a free zero-training model finds ~44 of the 72 ramps RampNet misses —
61%.** The null is measured *on the miss subset*, not extrapolated from the split-wide one, because
RampNet's misses are a biased sample and that is where density differs; it landed at 0.143 against
0.145, so the extrapolation would have been fair — a result, not an assumption.

**The 384 control is the interesting half.** Parity moved F1 by +0.018 and was correctly judged not
to change the ranking — but it nearly **tripled** the attributable complementary gain (~15 → ~44)
and shrank the found-by-nobody core from 51 to **18**. A flat headline metric hid a large change in
the structure underneath it.

**A naive union is still dead:** F1 0.555 vs RampNet's 0.855. The 54 ramps arrive with 442 false
positives — ~8.2 FPs per recovered ramp, against the 9 FPs RampNet pays for 238 TPs. The oracle
0.942 is recorded as the ceiling it is, with the real union arithmetic beside it.

**The cascade was then run — see below.**

### Tooling changes

- **`complementarity.py` generalized past its Gemini-only form** (it called
  `build_detector("gemini", ...)`, so this arm could not go through the gate at all). Takes a model
  spec now; the documented positional form still works. Verified against the committed paterson
  numbers — 36 of 124, 88 found by no model, oracle-union 0.777, reproduced exactly.
- **`--vistas-*` args added to `null_recall.py`.** `build_detector` reads them via `getattr`, so
  without them the null there was silently the **384** arm even when analysing parity.
- **Provider defaults now come from `roster.PROVIDER_DEFAULTS`** via the same shim shape
  `fp_taxonomy` uses, so the reconstructed cache key cannot drift from the one `compare.py` wrote
  under — a wrong default does not crash, it silently misses every entry and reports zero.
- **The oracle-union number now prints its counterweight**: the FP bill a real union pays and the
  resulting P/R/F1. On paterson/gemini-pro the union genuinely wins (0.821 vs 0.805), which is the
  check that the line is not hardcoded to one verdict.

`pytest -q`: **1238 passed, 1 skipped** after every step.

---

### The complementarity numbers above are against a RampNet we do not recommend

The committed bundle detections are the **shipped** operating point (≥ 0.5519 on richmond); this
document has recommended **0.30** since #54/#55 (PR #79). Verified `op_cache` is the same source
before re-basing: filtered at ≥ 0.5519 it reproduces the published row exactly (238/9/72). At 0.30
it gives **257/28/53, P 0.9018 / R 0.8290 / F1 0.8639**, matching committed `corrected_at_0.3.csv`.

| | rampnet @0.55 | **rampnet @0.30 (recommended)** |
|---|---:|---:|
| rampnet recall / F1 | 0.768 / 0.855 | **0.829 / 0.864** |
| rampnet misses | 72 | **53** |
| challenger recovers | 54 (75%) | **38 (72%)** |
| **attributable after null** | ~44 | **~30** |
| naive union F1 | 0.555 | 0.549 |

**~14 of the recoveries are ramps RampNet already has at the recommended point.** Deployable gain
is **~30, not ~44** — the correction is applied in the doc, next to the original figure rather
than replacing it. The recovery *rate* barely moves (75% → 72%): fewer misses to find, not a
different kind of miss.

### `cascade_gate.py` (new): the cascade is live, ceiling ~19 ramps

Partitions all 310 GT ramps into the four cells and reads RampNet's heatmap at each, reusing #46
Phase 1's instrument verbatim. Read pre-registered on #126 before running. At rampnet@0.30, of the
38 genuinely-complementary ramps:

| what RampNet has there | n | meaning |
|---|---:|---|
| floor peak in radius, **0.05–0.30** | **19** | **promotable** — the cascade's real target |
| floor peak ≥0.30 but unmatched | 4 | greedy matcher gave it to an adjacent GT — matcher/σ (#130), not threshold |
| no floor peak in radius | 15 | nothing to promote; `act` 0.215 is unpeaked mass |

**Ceiling ≈ 19 ramps, +6.1 recall points (0.829 → 0.890), FP cost unmeasured.** Far below the 54
the raw complementarity suggested.

**Negative worth keeping:** RampNet's activation does *not* discriminate — `challenger_only` null
percentile 0.88 vs the hard core's 0.925, so the ramps nobody finds look *stronger* on raw heatmap
mass. No self-gating shortcut exists.

### An instrument bug I caught mid-run

`nearest_peak` expects floor peaks (≥0.05); I first fed it the shipped ≥0.5519 detections, so
"is there a peak near this missed ramp?" was really "is there a 0.55+ detection near it?" — close
to definitionally no for a miss. It made every miss look peakless (nearest peak a median 104 px
away) and I nearly wrote up "the cascade is dead". Fixed in `dcb9c32`, artifacts regenerated. The
greedy match that *defines* the miss set still uses the bundle records, since that is what
produced the published 238/9/72 — two inputs, two questions, now documented.

New committed artifacts: `analysis_out/cascade_gate.json`, `analysis_out/cascade_gate_op030.json`
(both whitelisted in `.gitignore`, written with `newline=""` so a Windows re-run stays
byte-comparable).


jonfroehlich:
## Deep review

Re-ran everything from the branch head (`54a98c6`) in a clean detached worktree, against the committed inputs only. **The headline result stands and the arithmetic in the doc is right almost everywhere it can be checked.** What reproduced:

- Full suite: **1238 passed, 1 skipped**, as claimed.
- RampNet's committed richmond row from `records.jsonl` + `verdicts.json`: **238/9/72, P 0.9636 / R 0.7677 / F1 0.8546, AP 0.7635**. `analysis_out/op_cache/richmond.json` filtered at ≥ 0.5519 reproduces it exactly; at ≥ 0.30 it gives **257/28/53, 0.9018 / 0.8290 / 0.8639**, matching the doc and `corrected_at_0.3.csv`.
- The committed 384 detections (`benchmark/model_detections/mask2former-vistas-curb-cut__richmond.json`) reproduce the "published" row **216/309/94, 0.411 / 0.697 / 0.517, AP 0.513** and the ≥ 0.30 row **0.419 / 0.697 / 0.524**.
- `complementarity.py gemini-3.1-pro-preview paterson` (legacy positional form) reproduces the committed paterson numbers exactly: 36 of 124, 88 found by no model, oracle-union 0.777, naive union 0.821 vs 0.805.
- `cascade_gate_op030.json`: 38 `challenger_only` sites split **19 / 4 / 15** as the table says; null percentile 0.88 vs 0.925; 6 of 15 `neither` sites carry a ≥ 0.30 peak; 6 seam sites, 5 in `both` and 1 in `challenger_only`. All match.
- Every number in the union tables re-derives from the cell counts (0.393 / 0.942 / 0.555; 0.549 at 0.30; 5.8%; 8.2 FP per ramp; 3.6 vs 0.07 FP/pano; +6.1 recall points).
- Both pre-registrations are on #126 with timestamps before the corresponding results.
- `git merge-tree` against `origin/main`, #136, #146 and #147: **no conflicts** on any of the four, despite the shared `docs/model_comparison.md`.
- Every symbol `cascade_gate.py` imports from `silent_activation`, `farfield_forensics`, `operating_point_curve`, `miss_gallery` and `threshold_sweep` exists on this branch **and** on `origin/main`, so the merge does not break its imports.
- No licensing or terms-of-service text in the diff. The new shell block carries real line continuations (#136's class).

Everything below is about what a clean clone cannot re-derive, one column that quietly comes from a different run than it says, and a docstring quote that lost its second half.

---

### 1. High — the parity run and everything downstream of it rest on detections that exist on one machine

The parity detections (1024×1024) are, per the doc, "deliberately not published", and the same-env 384 control is not published either. Both live in a private `--cache-dir` on makelab2. That single sentence is true, but its consequence is not stated where the replication rule requires it — beside every number it affects. Those are:

- the parity row (274/442/36, AP 0.649) and the control row (215/308/95) in the first table;
- the entire *Complementarity* table at parity (54 / 18 / 0.143 / ~44 / 0.942 / 6.2 · 0.864);
- the entire *operating-point correction* table (38 / ~30 / 15 / 0.952 / 0.549);
- both committed `cascade_gate*.json` — the four-cell partition inside them was computed against those detections, so the artifacts are committed but cannot be regenerated;
- the seam-exposure count (1 in 38).

The only side of any of those a clean clone can rebuild is RampNet's. The doc's own "Reproducing it" block gives three `compare.py` commands that re-run the model, which is a different thing from re-deriving the committed numbers from committed inputs. The mechanism to close this exists on `main` (this branch is 23 commits behind it): `export_model_cache.py --publish-as` from #123 gives the parity arm a distinct published name so it does not overwrite `mask2former-vistas-curb-cut__richmond.json`.

**Fix:** a one-line statement next to each of the five items above: *"the 1024 and same-env 384 detections are not published; these numbers are re-derivable only from the makelab2 cache. Publishing them needs `--publish-as` (#123) after this branch takes `main`."* Publishing is a decision for Jon (below).

### 2. Medium — the "vistas @384" column is the unpublished control run, not the published arm, and it is not labelled

Re-running `complementarity.py vistas:curb-cut richmond` against the **published** 384 detections gives:

| | doc says | published 384 detections give |
| :--- | ---: | ---: |
| challenger ONLY | 21 (29%) | **22 (31%)** |
| found by NEITHER | 51 | **50** |
| oracle-union recall | 0.835 | **0.839** |
| naive union F1 | 0.585 | **0.586** |
| above chance (`null_recall.py`) | 0.657 | **0.661** |
| attributable after the null | ~15 | ~15 |

The doc's column matches the same-env control (215/308/95), which differs from the published run by one detection. The text says "on the same cached detections" without saying which of the two 384 runs that is. Nothing turns on one ramp, but as written the column is not reproducible from committed data and the reader has no way to know why.

**Fix:** either replace the column with the published-arm values above (then it reproduces from a clean clone with the command in the doc), or label it *"same-env control, unpublished"*. Replacing is better. The rate becomes 22 (31%).

### 3. Medium — the seam paragraph quotes half of `score_pano`'s docstring, and the half it drops is the one that applies

The doc cites `score_pano`'s docstring as saying wrapping *"moves no metric on any committed split"*. The sentence on `main` is: *"Wrapping moves no RampNet or YOLO metric on any committed split — **but it does move the challengers.**"* The challenger is the side being partitioned here. The paragraph's conclusion may still be right, but the evidence offered for it is a misquote.

Measured on this branch with a wrapping copy of `matched_gt`: **0 cell flips** on RampNet's side at ≥ 0.5519, ≥ 0.30 and ≥ 0.05, and **0 flips** on the published 384 arm. The parity arm cannot be checked (finding 1), so the doc's own count — 1 seam site in 38 — remains the bound, and it holds.

One more consequence for the merge. On `main`, `score_pano` wraps by default, so after this branch takes `main`, the `r_fp` / `c_fp` and union P/R/F1 in `complementarity.py` will come from a wrapping matcher while the four cells come from the non-wrapping `matched_gt` in the same script. Two matchers in one output.

**Fix:** quote the docstring in full and record the measured zero-flip result for RampNet and the 384 arm. Merge `origin/main` into the branch (verified clean), then route `matched_gt` through `rampnet.geometry.dist_sq(..., wrap_x=True)` so the cells and the FP counts use one matcher, and state beside the committed JSONs that they were computed before the wrap with the 1-in-38 bound.

### 4. Medium-low — `act` 0.215 is the median of the wrong subset

The cascade table's third row, *"no floor peak in radius | 15 | `act` is 0.215 median"*: 0.2152 is the median `act` of the whole `challenger_only` cell (n = 38, `cells[].act_median` in the JSON). The 15 no-peak sites have median `act` **0.272** (mean 0.369). The 19 promotable sites have median 0.153. The point the row makes — unpeaked mass with nothing to promote — is unaffected, but the number attached to it is the cell's, not the row's.

**Fix:** 0.215 → 0.272.

### 5. Medium-low — three different numbers for how many recoveries were RampNet's own: 19, 16, ~14

At 0.55 → 0.30 RampNet gains 19 hits (72 → 53 misses). Of those 19, **16** came out of `challenger_only` (54 → 38) and 3 out of `neither` (18 → 15). So:

- `complementarity.py` docstring: *"19 of the 'misses' a challenger gets credit for recovering are ramps RampNet already has"* — 19 is the miss reduction; the challenger's credited recoveries fall by 16.
- doc: *"~14 of the ramps the challenger got credit for recovering are ramps RampNet already has"* — ~14 is the null-discounted difference (~44 → ~30); the raw count is 16.
- `cascade_gate.py` help text: *"19 of the shipped point's misses are ramps RampNet already has"* — correct as phrased.

**Fix:** docstring → 16 (of the 19); doc → "16 of the 54, ~14 after the null".

### 6. Medium-low — none of the new code has a test

`complementarity.py` (spec parsing, `compare_args`, `complementary_null`, the union arithmetic), `cascade_gate.py` (`cell_of`, `summarize`) and the new `null_recall.py` arguments add zero tests; the suite count is unchanged at 1238. One concrete hole: `model_spec("foo:bar")` — an unknown provider *with* a colon — is silently read as the Gemini model id `"foo:bar"`, and the script then reports zero cached detections rather than a bad spec.

**Fix:** tests for `model_spec` (provider, `provider:id`, legacy bare Gemini id, and unknown-provider-with-colon should raise), `compare_args` producing a `vistas` signature equal to `compare.py`'s with and without `--vistas-input-size`, `complementary_null` on a synthetic three-pano case, `cell_of`/`summarize` on synthetic rows, and the 384 regression (194 / 44 / 22 / 50) from the committed `model_detections` file written into a temporary `DetectionCache`.

### 7. Low — `analysis_out/cascade_gate.json` predates its own schema

It was last written at `b7342dc`; `4c192ca` then added `rampnet_op_threshold` to the payload, and `cascade_gate_op030.json` (written at `730baff`) carries it while the shipped-point file does not. A regeneration would change bytes (`"rampnet_op_threshold": null`) even with identical results — which is exactly the comparison `newline=""` was added to protect. It cannot be regenerated without the parity cache (finding 1).

**Fix:** state it beside the artifact list in the doc; regenerate when the parity detections are published.

### 8. Low — the run instructions for the complementarity and cascade sections are not in the doc, and the README has no row for the new script

"Reproducing it" lists three `compare.py` commands and nothing else. The exact invocations for the complementarity table, the 0.30 re-base, `null_recall.py` at parity and both `cascade_gate.py` runs exist only in docstrings and the PR body. `scripts/analysis/README.md` gained a `complementarity.py` row but no `cascade_gate.py` row; every other script in the directory has one.

**Fix:** add the five commands to the block; add the README row.

### 9. Low — cost is wall-clock for one of three runs

3m38s is given for the parity run; the same-env control and the RampNet re-run alongside are not timed; there is no GPU-hour or dollar figure. Three runs on one lab A40 is about 0.2 GPU-hours and $0, and the rule is to write both units down.

**Fix:** one line: three runs, total wall-clock, ~0.2 GPU-h, $0 (lab hardware).

### 10. Low — two silent paths in `cascade_gate.py`

- Under `--rampnet-op-threshold`, a pano missing from `op_cache` is scored as RampNet-blank (`floor_peaks.get(pid, [])`), so every GT on it becomes a miss with no warning. All 124 richmond panos are present, so nothing here is affected; another split would not be told.
- `preds = floor_peaks.get(pid) or [bundle records]` substitutes the bundle per-pano whenever `op_cache` lists a pano with zero floor peaks, while the header still prints `floor peaks from: op_cache`. Harmless (the bundle is a subset of the floor) but the two-input story in the docstring is no longer exactly true.

**Fix:** warn and count panos absent from `op_cache`; drop the `or` fallback when `floor_src` is `op_cache`.

### 11. Low — register

"a matching bug wearing a threshold costume", "wearing a different cache key", "dead" (three times), "the economics are the whole story", "earns its place", "the honest way to read it". These are published under Jon's name; each has a plain statement behind it.

---

- **Publish the parity detections.** `--publish-as` (e.g. `mask2former-vistas-curb-cut-1024`) once the branch takes `main`, assuming the private cache directory on makelab2 still exists. Without it, finding 1 stays open and the three tables and two artifacts remain results that live on one host.
- **Merge `main` into the branch before it lands**, so `matched_gt` can use `rampnet.geometry` and the two matchers in `complementarity.py` agree. Verified conflict-free; measured zero effect on RampNet's and the 384 arm's cells.

### What holds up

- The decision rule was read fixed on #126 before the result, and the result is not close: +0.018 F1 against a 0.05 bar.
- The env control is the right design and it did its job — one detection in 523 between transformers 4.x and 5.15.
- The mechanism read is correct and the doc says it plainly: resolution bought recall and AP, not precision.
- The null measured on the miss subset rather than extrapolated (0.143 vs 0.145) is the right instrument and the doc reports that it landed where the extrapolation would have.
- `cascade_gate.py` asks the one question that decides whether a cascade is possible, uses #46 Phase 1's instrument by import rather than copy, and the `peak_in_radius` column — not the class — is correctly identified as the one that decides it.
- The op-0.30 re-base is the most important correction in the PR and it is applied next to the original number rather than over it.
- The mid-run instrument bug (feeding ≥ 0.5519 detections to `nearest_peak`) was caught, recorded, and the artifacts regenerated.
- `compare_args` reads provider defaults from `roster.PROVIDER_DEFAULTS`; every `vistas_*` key it needs is present, and `null_recall.py`'s new arguments follow the deviation-only convention exactly.
- The two JSONs are whitelisted correctly (`git check-ignore` returns nothing) and written LF-pinned.

## Fix list

- [F1] High — Next to the parity/control table, the complementarity table, the op-0.30 table, the cascade tables and the artifact list, state that the 1024 and same-env 384 detections are not published and these numbers re-derive only from the makelab2 cache; name `export_model_cache.py --publish-as` (#123, on `main`) as the way to close it — files: docs/model_comparison.md
- [F2] Medium — Replace the "vistas @384" column with the published-arm values (194 / 44 / 22 (31%) / 50, oracle-union 0.839, naive-union F1 0.586, above-chance 0.661) and add one sentence that the same-env control differs by one detection — files: docs/model_comparison.md
- [F3] Medium — Merge `origin/main` into the branch (clean); quote `score_pano`'s docstring in full; route `matched_gt` through `rampnet.geometry.dist_sq(wrap_x=True)` so cells and FP counts use one matcher; record the measured zero-flip result for RampNet (all three thresholds) and the 384 arm, and note the committed JSONs predate the wrap with the 1-in-38 bound; re-run the suite after the merge — files: scripts/analysis/complementarity.py, scripts/analysis/cascade_gate.py, docs/model_comparison.md
- [F4] Medium-low — Cascade table row 3: `act` 0.215 → 0.272 (median of the 15 no-peak sites; 0.215 is the whole cell) — files: docs/model_comparison.md
- [F5] Medium-low — "19 of the misses a challenger gets credit for recovering" → 16 of the 54 (3 of the 19 were in `neither`); doc "~14" → "16, ~14 after the null" — files: scripts/analysis/complementarity.py, docs/model_comparison.md
- [F6] Medium-low — Add tests: `model_spec` (four cases, unknown-provider-with-colon raises), `compare_args` vistas signature with/without input size equals `compare.py`'s, `complementary_null` synthetic, `cell_of`/`summarize` synthetic, and the 384 regression 194/44/22/50 from the committed model_detections file via a temp `DetectionCache` — files: tests/test_complementarity.py (new), tests/test_cascade_gate.py (new), scripts/analysis/complementarity.py
- [F7] Low — State beside the artifact list that `cascade_gate.json` was written before `rampnet_op_threshold` entered the payload and will differ by that key when regenerated — files: docs/model_comparison.md
- [F8] Low — Add the exact `complementarity.py` (parity, and `--rampnet-op-threshold 0.30`), `null_recall.py --vistas-input-size 1024 1024` and both `cascade_gate.py` commands to "Reproducing it"; add a `cascade_gate.py` row to the README table — files: docs/model_comparison.md, scripts/analysis/README.md
- [F9] Low — Record cost in both units: three runs on one A40, total wall-clock, ~0.2 GPU-h, $0 — files: docs/model_comparison.md
- [F10] Low — `cascade_gate.py`: warn and count panos absent from `op_cache` under `--rampnet-op-threshold`; drop the per-pano `or` bundle fallback when floor peaks come from `op_cache` — files: scripts/analysis/cascade_gate.py
- [F11] Low — Replace "wearing a threshold costume", "wearing a different cache key", "dead", "the economics are the whole story", "earns its place", "honest way to read it" with plain statements — files: docs/model_comparison.md, scripts/analysis/cascade_gate.py

jonfroehlich:
## Review fixes

All eleven items applied. Four commits: `1b634e2` (merge `main`), `a3ebe37` (the matcher),
`7e8c9d0` (tests + the two silent paths), `70ce265` (the write-up).

**Merge with `main` and the post-merge suite.** `git merge origin/main` was conflict-free as the
review found, and the suite went **1238 passed / 1 skipped → 1322 passed / 1 skipped** on the
merge alone (`main` brought 84 tests, mostly the #132 seam work and the scoreboard).

**The zero-flip measurement.** Cells are both / rampnet-only / challenger-only / neither, wrapping
against not wrapping, on the committed detections:

| arm | RampNet's side | cells | wrapped |
|---|---|---|---|
| vistas 384 (published) | bundle, ≥ 0.5519 | 194 / 44 / 22 / 50 | identical |
| vistas 384 (published) | op_cache ≥ 0.30 | 202 / 55 / 14 / 39 | identical |
| vistas 384 (published) | op_cache ≥ 0.05 | 213 / 66 / 3 / 28 | identical |
| gemini-3.1-pro-preview, paterson | bundle | 188 / 83 / 36 / 88 | identical |
| gemini-3.1-pro-preview, paterson | op_cache ≥ 0.30 | 194 / 90 / 30 / 81 | identical |
| gemini-3.1-pro-preview, paterson | op_cache ≥ 0.05 | 201 / 98 / 23 / 73 | identical |

Zero cell flips everywhere, so the published 384 column and the committed #35 gate numbers are
unchanged by the wrap. The paterson gate reproduces exactly as the review reported it: 36 of 124,
88 found by no model, oracle-union 0.777, naive union 0.821 vs 0.805. The parity arm cannot be
re-checked — its detections are not published — so its bound stays the doc's own 1 seam ramp in 38.

**Per item**

- **F1** — `70ce265`. Statements next to all five: the parity/control rows, the 1024
  complementarity column, the operating-point table, the cascade artifacts, and the seam count.
  `export_model_cache.py --publish-as` (#123, now on this branch via the merge) named as the fix.
- **F2** — `70ce265`. Column replaced with the published arm, re-run here rather than copied:
  challenger-only **22 (31%)**, found-by-nobody **50**, oracle-union **0.839**, naive-union F1
  **0.586**, above-chance **0.661** (`null_recall.py` gives 4.5 boxes/pano · 0.661). Every value
  matched the review's. `found by BOTH` 194 and `rampnet ONLY` 44 were already right — the two
  arms agree on those cells — and `attributable after the null` stays ~15 (null 0.090).
- **F3** — merge `1b634e2`, matcher `a3ebe37`, docstring and measurement `70ce265`. `matched_gt`
  now calls `rampnet.metrics.greedy_match(..., wrap_x=True)`, the same core `score_pano` uses, so
  the four cells and the FP counts printed in one table come from one matcher.
- **F4** — `70ce265`. 0.215 → **0.272** (0.369 mean). Checked against
  `cascade_gate_op030.json`: 0.2152 is `cells[].act_median` over the 38-ramp cell, the 15 no-peak
  rows are at 0.2723, the 19 promotable rows at 0.1525. All three are now named.
- **F5** — `7e8c9d0` (docstring, `--rampnet-op-threshold` help) and `70ce265` (doc). 19 is the
  miss reduction; 16 of the 19 come out of `challenger_only` and 3 out of `neither`, so the
  headline falls by 16, and ~14 is the null-discounted figure. Verified from the two JSONs.
- **F6** — `7e8c9d0`. `tests/test_complementarity.py` (21) and `tests/test_cascade_gate.py` (18).
  Committed files only — `benchmark/richmond`, `benchmark/model_detections`, the two
  `cascade_gate` JSONs — no `.model_cache`, no network, no GPU. Covers `model_spec` (including
  unknown-provider-with-colon, which now exits rather than being read as a Gemini model id),
  `compare_args` against the signature recorded *inside* the published export with and without
  `--vistas-input-size`, `complementary_null`, `cell_of`, `summarize`, the seam case on
  `matched_gt`, and the 384 regression 194/44/22/50 through a `tmp_path` `DetectionCache`. Also
  `summarize(sites) == cells` on both artifacts, so a hand-copied cascade number now fails CI.
- **F7** — `70ce265`. Stated beside the artifact list.
- **F8** — `70ce265`. Five commands added; `cascade_gate.py` row added to
  `scripts/analysis/README.md`. One thing found while writing it: `complementarity.py` and
  `null_recall.py` read `.model_cache`, not the published export the way `fp_taxonomy.py` and
  `silent_witness.py` do, so a clean clone has to materialise the export into a cache directory
  first. That step is now written down, and `tests/test_complementarity.py` is the working example.
- **F9** — `70ce265`. $0 (lab-owned hardware) and 3m38s for the one leg of three that was timed.
  No session total is quoted: the 384 control was untimed and RampNet's row loads no model.
- **F10** — `7e8c9d0`. Panos absent from `op_cache` under `--rampnet-op-threshold` are counted and
  warned about; the per-pano `or <bundle records>` fallback is gone when the floor came from
  `op_cache`. All 124 richmond panos are present with peaks, so no committed number moves. No test
  for the warning path — reaching it in `main()` requires the model load, so it would need a GPU
  or a refactor; `cell_of`/`summarize` are tested directly instead.
- **F11** — `70ce265`. Six phrases replaced with plain statements.

Nothing skipped. Nothing published: the parity detections stay unpublished, which is the decision
left for you.

Final suite: **1361 passed, 1 skipped**.

jonfroehlich:

2 major / 4 minor / 3 nit confirmed, from three independent reviews (Claude Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash) verified against the worktree at `70ce265`. Every number a clean clone can re-derive from this branch reproduces: the 384 cells at all three thresholds, the paterson gate, every `cells[]` figure in both `cascade_gate*.json`, the 19 / 4 / 15 split, the act medians, the 6-of-15 matcher-claimed peaks in `neither`, the seam count. The one thing to fix before merge is S1: the doc's interpretation of the 15 "no floor peak" sites is contradicted by the committed artifact's own columns.

### Confirmed

#### S1 · major · scientific integrity — the "no floor peak" row is mostly adjacent-peak tail, not unpeaked mass
- **Where:** `docs/model_comparison.md` cascade-gate table — `| no floor peak in radius | 15 | nothing to promote. \`act\` across these 15 is 0.272 median (0.369 mean) — unpeaked heatmap mass \`peak_local_max\` never called a maximum. |`, and *"Two-fifths of the recoverable set has no peak to raise"*; same wording in `scripts/analysis/cascade_gate.py` docstring and the printed footer.
- **What:** Re-derived from `analysis_out/cascade_gate_op030.json` joined to `analysis_out/op_cache/richmond.json` (R = 22.5 px): of the 15 `challenger_only` sites with no peak in radius, **11 have a floor peak 1.0–2.0 R away** (nearest-peak median 35.0 px), and for 10 of those 11 the in-window `argmax_off_px` is 22.2–22.5 px, i.e. the maximum sits on the window edge. That is a neighbouring mode's shoulder reaching in — [#46](https://github.com/ProjectSidewalk/RampNet/issues/46) Phase 1's `tail` definition — not mass `peak_local_max` never called. `class_of` puts the 15 at 12 `tail` / 3 `faint_local` / 0 `absent`. Of the 15 neighbouring peaks, 11 are ≥ 0.30 and already matched to another GT at 0.30 (7 of them within 2 R), which is the same matcher/σ category ([#130](https://github.com/ProjectSidewalk/RampNet/issues/130)) the table assigns only to the 4-row. Only 4 sites have no peak within 2 R. The pre-registration on [#126](https://github.com/ProjectSidewalk/RampNet/issues/126) promised the `class_of` decomposition "so the numbers are directly comparable" to Phase 1's 8 / 62 / 30; the artifact carries it (`cells[].classes`) but the doc never reports it.
- **Evidence:** Script re-run in this review; per-site table matches Opus's to every value. One site is different in kind: `723487737079243` at x = 0.0069 (the one `challenger_only` seam site) has `act` 0.946 at 7.4 px, `center` 0.78, and no op_cache peak within 117 px — the `f4c71c8` seam dropout the doc bounds abstractly, made concrete. A regenerated op_cache would almost certainly list that peak, which makes the site a RampNet **hit** at 0.30 and moves it out of `challenger_only` (38 → 37), not "nothing to promote".
- **Fix:** The count 15 and the ceiling of 19 under an exactly-R gate both stand; the gloss does not. Re-cut the 38 as 19 promotable in R / ~11 peak ≥ 0.30 at ≤ 2 R claimed by an adjacent GT (#130 mechanism, merging the 4-row) / 3 unmatched peaks at 1.0–1.45 R (two ≥ 0.30, one 0.242) / 1 seam-dropped / 4 no peak within 2 R. Report `nearest_peak_px` median and `argmax_off_px` median for the row, add the `class_of` shares, and delete "unpeaked heatmap mass" from the doc and both places in `cascade_gate.py`. Pin it with a test on the op030 artifact: the no-peak rows' `argmax_off_px` median == 22.4 and count with a floor peak within 2 R == 11.
- **Found by:** Opus 5 (counts of 9 / 6 / 2 in the original corrected here to 10 of 11 / 11 (7 within 2 R) / 3 — the mechanism claim is unchanged).

#### S2 · major · replicability — the only copy of both A40 runs is a private cache in a scratch worktree
- **Where:** `docs/model_comparison.md` — `in a scratch worktree with a private --cache-dir` and `Neither A40 run's detections are published`.
- **What:** The disclosure is complete and follows the standing rule, but the parity (1024) and same-env control (384) detections exist only in that cache dir on makelab2. Five doc sections, both committed `cascade_gate*.json`, and the S1 re-cut all re-derive from them; if the scratch worktree is pruned, every 1024 number becomes permanently unreproducible.
- **Evidence:** `benchmark/model_detections/` holds `mask2former-vistas-curb-cut__richmond.json` (384) only. `scripts/analysis/export_model_cache.py` already has `publish_as`, and `rampnet/roster.py` already carries the vistas leg, so publication is a roster pin plus one export. `tests/test_cascade_gate.py` can only check the artifacts against themselves (its docstring says so).
- **Fix:** Add a roster leg pinned at `vistas_input_size=[1024,1024]` with `published_as="mask2former-vistas-curb-cut-1024"`, export both A40 runs (the 384 control is the attribution arm, so it too), and extend `test_the_published_384_column_reproduces` with the 1024 column (220 / 18 / 54 / 18). At minimum, copy the cache dir to `/gscratch/makelab` today.
- **Found by:** Opus 5

#### S3 · minor · correctness — `--radius` reaches the cells but not the FP counts
- **Where:** `scripts/analysis/complementarity.py` — `r_fp += score_pano(rp, gt).fp` (and the `cp` line below it).
- **What:** `--radius` is new in this PR and is threaded to `matched_gt`, but `score_pano` is called without `radius_sq` and falls back to the default 0.022. With a custom radius the four cells and the FP bill are on two different matchers, in a script whose docstring says keeping them on one is the point.
- **Evidence:** Ran on the published 384 arm from a cache seeded from the committed export: `--radius 0.05` moves the cells 194/44/22/50 → 202/37/30/41 while `FP cost` stays 9 | 309 and "rampnet alone P 0.964" is printed against a recall computed at the other radius. Default runs (all committed numbers) are unaffected.
- **Fix:** Pass `radius_sq` to both `score_pano` calls. Test: `test_the_cells_and_the_fp_counts_come_from_one_matcher` extended with a non-default radius.
- **Found by:** Gemini 3.1 Pro, Gemini 3.8 Flash

#### S4 · minor · silent failure — a pano missing from the op_cache scores as RampNet-blank without a warning
- **Where:** `scripts/analysis/complementarity.py` — `rp = [p for p in floor_peaks.get(pid, [])`; `scripts/analysis/cascade_gate.py` — `preds = (floor_peaks.get(pid, []) if have_op_cache else`.
- **What:** In `complementarity.py` under `--rampnet-op-threshold`, and in `cascade_gate.py`'s default probe path, a pano the op_cache does not list gets `[]`: every GT on it becomes a miss (complementarity) or reads `peak_in_radius: false` / `nearest_peak_px: null` (cascade) while the header still says the floor came from the op_cache. `cascade_gate.py`'s F10 `no_floor` warning only fires inside the threshold branch.
- **Evidence:** Code paths above. `analysis_out/op_cache/richmond.json` lists all 124 richmond panos, so no committed number is affected; a split extracted with `--limit`, or a pano added after extraction, would inflate the no-peak row in the direction that under-states the cascade.
- **Fix:** Count and warn in both scripts whenever the op_cache is the source, not only under the threshold flag. Test: a fake op_cache missing one pano must produce the warning.
- **Found by:** Gemini 3.1 Pro (complementarity), Opus 5 (cascade_gate)

#### S5 · minor · correctness — one null RNG stream across sites, so a site's `null_pct` depends on which sites preceded it
- **Where:** `scripts/analysis/cascade_gate.py` — `rng = random.Random(NULL_SEED)`.
- **What:** The stream is consumed in pano order for miss-cell sites only. The miss set differs between the two artifacts (19 cell transitions), so the same site draws different azimuths in each.
- **Evidence:** Joined the two artifacts on `(pano, x, y)`: 53 sites carry a null in both; **43 differ**, max |Δ| 0.075; `act` and `nearest_peak_px` differ on 0. The doc's within-file comparison (0.88 vs 0.925) stands; any per-site comparison across runs moves by up to ±0.075 for reasons unrelated to the heatmap.
- **Fix:** Seed per site, e.g. `random.Random(f"{NULL_SEED}:{pid}:{x}:{y}")`. `silent_activation.py` (line 331) has the same shape. Test: the same site with different preceding sites returns the same `null_pct`.
- **Found by:** Opus 5

#### S6 · minor · test infrastructure — the partition loop is copied three times and the regression test bypasses the script
- **Where:** `tests/test_complementarity.py` `_published_cells`, `complementarity.py` main loop, `cascade_gate.py` lines 200–225.
- **What:** Line-for-line the same logic. `test_the_published_384_column_reproduces` re-implements the partition rather than calling the script, so a bug in either script's loop (the threshold filter, the `.get(pid, [])` fallback, the `cp is None: continue`) does not fail it.
- **Fix:** Extract `partition_cells(...)` into `complementarity.py`, call it from both `main`s and the test; then the test can also assert 202/55/14/39 at 0.30 and 213/66/3/28 at 0.05, both verified from committed inputs in this review.
- **Found by:** Opus 5

#### S7 · nit · robustness — `--rampnet-op-threshold` on a split without an op_cache is a raw traceback
- **Where:** `scripts/analysis/complementarity.py` — `cached, _ = read_cache(os.path.join(CACHE_DIR, f"{args.split}.json"))`.
- **What:** `FileNotFoundError` with the path, where `cascade_gate.py` already wraps the same call and exits with a message. Loud, not silent; every split in this worktree has an op_cache. Mirror the `cascade_gate.py` handling.
- **Found by:** Gemini 3.1 Pro, Gemini 3.8 Flash (both filed as correctness; downgraded)

#### S8 · nit · documentation — the subset null gives each box more targets than the real pairing did
- **Where:** `scripts/analysis/complementarity.py` — `hit = sum(len(matched_gt(preds[(i + k) % n], missed[i], radius_sq))`.
- **What:** The shifted pano's whole prediction set is matched against only the missed GT, whereas the real `c_only` was matched against all GT. The null is a slight over-estimate of chance, so the "attributable" numbers are conservative; `null_recall.py` matches against all GT, so 0.145 and 0.143 are not the same construction and their agreement is the evidence the bias is small. One docstring sentence.
- **Found by:** Opus 5

#### S9 · nit · documentation — PR body header is stale
- **What:** *"Docs only. No code, no detection artifacts, no roster change"* and *"1238 passed"*: the PR adds two scripts, two test files and two committed artifacts, and the suite is 1361 passed, 1 skipped. The body becomes the merge-commit message.
- **Found by:** Opus 5

### Rejected or downgraded
- **Gemini 3.1 Pro / 3.8 Flash: unhandled `read_cache` exception is a correctness defect** — partly valid: the crash is real (reproduced: `FileNotFoundError` naming the path) but it is loud, not silent, and cannot corrupt a number. Kept as S7, nit.
- **Opus 5, S1 sub-counts** — partly valid: "9 of them argmax 22.2–22.5" is 10 of the 11 within 2 R; "6 of 15 matched ≥ 0.30 to an adjacent GT" is 11 of 15 overall (7 within 2 R, 6 within 1.6 R); "2 sub-0.30 peaks at 24–32 px" is 1 sub-0.30 (0.242 @ 24.3 px) plus 2 unmatched ≥ 0.30 (0.643 @ 23.5, 0.358 @ 32.2). The mechanism claim and the 11-of-15 / 12-tail / 4-no-peak-within-2R figures are confirmed exactly.
- No finding from any reviewer was rejected outright.

### Coverage
- Files excluded from the Gemini reviews: `analysis_out/cascade_gate.json`, `analysis_out/cascade_gate_op030.json` (both runs; 98,819-char diff, 8 files sent). Gemini 3.1 Pro exit 0, 3 findings. Gemini 3.8 Flash exit 0, 2 findings, on its second attempt — the first exited 4 (it tried to run a shell command headlessly and produced no usable review).
- Tests run: full suite in the worktree with `PYTHONPATH=.` — 1361 passed, 1 skipped in 99 s. Re-derived in this review from committed files only: the S1 per-site table (15 rows, every value), `class_of` shares, both artifacts' `cells[]` summaries, the 43-of-53 null drift, op_cache coverage (124/124), `complementarity.py` at radius 0.022 and 0.05 on the published 384 arm, the paterson gate.
- Not checked: anything needing the unpublished 1024 / A40-control detections (parity and control rows, the 1024 complementarity column, `above chance 0.864`); the GPU half of `cascade_gate.py` (`ts.heatmap_for`) was read, not run; whether the makelab2 cache dir still exists.

---

jonfroehlich:

Three commits on `exp/vistas-parity-126`, `70ce265..5d0c05e`. The doc's reading of one table row changes (S1); no count or committed artifact changes.

- **S1 — fixed in `cc2299e`:** the "no floor peak in radius" row is re-cut from the artifact's own columns and "unpeaked heatmap mass" is withdrawn from the doc, `cascade_gate.py`'s docstring and its printed footer. Of the 15: nearest floor peak median **35.0 px** (R = 22.5), in-window maximum on the window edge for 11 (median `argmax_off_px` 22.4), `class_of` 12 tail / 3 faint_local / 0 absent; for 11 of 15 the nearest peak is ≥ 0.30 and already matched to another GT at 0.30 (7 within 2 R), the same [#130](https://github.com/ProjectSidewalk/RampNet/issues/130) mechanism as the 4-row (all 4 of which are claimed the same way); 4 unclaimed at 1–2 R; 4 with nothing within 2 R, one of them the seam site `723487737079243` (act 0.946 at 7.4 px, no op_cache peak within 117 px — a RampNet hit at 0.30 once the op_cache is regenerated, 38 → 37). The 38 are re-cut exhaustively in the doc as 19 / 11 / 4 / 4; the 19-ramp ceiling and the 15 count stand. `cascade_gate.py` gains `no_peak_profile` and `claimed_by_adjacent` and writes `nearest_peak_claimed` per site on new runs; tests: `test_the_no_peak_row_is_mostly_a_neighbours_shoulder_not_unpeaked_mass`, `test_the_no_peak_rows_nearest_peaks_are_mostly_claimed_by_an_adjacent_ramp`, `test_the_seam_site_in_the_recovered_cell_has_a_peak_the_op_cache_lacks`, plus unit tests for both helpers.
- **S2 — skipped, needs a decision (recorded in `5d0c05e`):** the private makelab2 cache cannot be found. Searched 2026-09-17: no worktree registered under `/homes/gws/jonf/RampNet`, no `.model_cache` shard written 2026-08-17..20, and the parity run's shard names (`compare.cache_key` over the 1024 signature) absent from `/homes/gws/jonf`, `/tmp`, `/var`, the root filesystem and the lab mounts. Nothing to copy or export. The doc now states that beside the numbers, with the re-run cost (3m38s, one A40) and the three things the roster's own rules leave open: a pinned 1024 leg forces the published 384 file to be renamed too (`test_every_leg_of_a_pinned_model_is_qualified`) and the pin-naming rule expects a scalar; the 384 control shares the published run's cache key by design, so it cannot be a pinned leg; `export_model_cache.py` does not take `--vistas-input-size`. Not guessed at.
- **S3 — fixed in `d1860e5`:** both `score_pano` calls take `radius_sq`, so the cells and the FP bill are on one matcher at any `--radius`; committed numbers unaffected (all at the default). Test: `test_a_non_default_radius_reaches_the_cells_and_the_fp_counts_alike`.
- **S4 — fixed in `d1860e5`:** a pano missing from the op_cache is counted and warned about in `complementarity.py` and on `cascade_gate.py`'s probe path (payload records `panos_without_floor_peaks`). Tests: `test_a_pano_missing_from_the_op_cache_is_counted_not_absorbed`, `test_a_pano_the_op_cache_does_not_list_is_named`.
- **S5 — fixed in `d1860e5`:** `cascade_gate.py` seeds the null per site (`site_rng`). Committed artifacts unchanged and the doc says, beside the 0.88 / 0.925 medians, that they were written from one stream (43 of 53 shared sites differ, max 0.075) and that a regeneration will move individual `null_pct` values by that much. `silent_activation.py` left as is — its committed Phase 1 artifacts were written from that stream. Tests: `test_a_sites_null_is_the_same_whatever_ran_before_it`, `test_one_stream_shared_across_sites_did_depend_on_order`, `test_the_committed_nulls_came_from_one_stream_and_say_so`.
- **S6 — fixed in `d1860e5`:** `partition_cells` in `complementarity.py`, called from both scripts and from the tests; the published 384 column is now also pinned at 0.30 (202/55/14/39) and 0.05 (213/66/3/28). Tests: `test_the_384_column_at_the_two_op_cache_thresholds_reproduces`, `test_the_threshold_filters_the_floor_peaks_and_nothing_else`.
- **S7 — fixed in `d1860e5`:** `--rampnet-op-threshold` without an op_cache exits with a message naming the split. Test: `test_a_split_without_an_op_cache_exits_with_a_message_not_a_traceback`.
- **S8 — fixed in `d1860e5`:** `complementary_null`'s docstring states the construction difference from `null_recall.py` and that it makes the attributable figures conservative.
- **S9 — fixed (PR body edited):** header now lists the scripts, tests and artifacts; test counts updated; the 384 column in the body's complementarity table now shows the published run (22/50) as the doc does, not the control (21/51); the cascade row carries the S1 re-cut; the parity-detections omission says the cache is gone.

`pytest -q` in the worktree with `PYTHONPATH=.`: **1385 passed, 1 skipped** in 97 s (1361 passed, 1 skipped before; 24 tests added, all CPU-only on committed files).

jonfroehlich:

0 major / 3 minor / 1 nit confirmed, from three independent reviews (Claude Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash) verified against the worktree at `5d0c05e`. Of the nine round-1 items, **8 are confirmed properly fixed and pinned** (S1, S3–S9); S2 is deferred with an accurate statement in the doc and the PR body, and stays an open decision. Every number in the S1 re-cut re-derives from `cascade_gate_op030.json` + `op_cache/richmond.json` (15 rows, nearest-peak median 35.0 px, `argmax_off_px` median 22.4, 11 on the edge, 11 within 2 R / 4 beyond, classes 12/3/0, seam 1, act mean 0.369, cell classes 35/3/0), and the full suite is 1385 passed / 1 skipped. All four new findings are in `docs/model_comparison.md` and none changes a number; the one to fix first is S1, where the prose attributes the 19-way re-cut of the 38 to the 15-row and the arithmetic does not close.

### Confirmed

#### S1 · minor · scientific integrity — the 15-row is 7 / 4 / 4, not 11 / 4 / 4
- **Where:** `docs/model_comparison.md` ~1611 — `So the row that read as "two-fifths of the recoverable set has no peak to raise" is 11 parts matching problem, 4 parts near-miss geometry and 4 parts absence.` and ~1607 — `the old "4" row and most of the old "15" row are one cause`
- **What:** 11 + 4 + 4 = 19, not 15. The table's own "claimed" row says the 11 are 4 inside R (the old 4-row) plus 7 at 1–2 R, so the 15-row decomposes as **7** matching / 4 near-miss / 4 absence. "Most of the old 15 row" is 7 of 15 by the table's own membership.
- **Evidence:** Re-derived from the artifact: of the 15 no-peak rows, `claimed_by_adjacent` is True for 11, but only 7 of those have `nearest_peak_px` ≤ 45 px; the other 4 claimed rows are the 51.3–117.4 px "absence" rows (scores 0.85 / 0.86 / 0.34 / 0.94, all matched to a neighbour). `test_the_no_peak_rows_nearest_peaks_are_mostly_claimed_by_an_adjacent_ramp` already pins exactly this split.
- **Fix:** "the 15 are 7 parts matching problem, 4 parts near-miss geometry and 4 parts absence; with the old 4-row folded in, the 38 are 19 / 11 / 4 / 4", and "7 of the old 15 row". Optionally note that the 4 "absence" sites also have a claimed nearest peak, just beyond 2 R. No test change.
- **Found by:** Opus 5

#### S2 · minor · documentation — the "regenerated copy" paragraph is stale after the round-1 fixes
- **Where:** `docs/model_comparison.md` ~1577–1581 — `a re-run would add "rampnet_op_threshold": null and change the bytes with identical results, which is exactly the comparison the newline="" pinning exists to make possible.`
- **What:** Since `d1860e5` / `cc2299e` a re-run no longer gives identical results with one extra key: the payload also adds `null_rng`, `panos_without_floor_peaks` and a per-site `nearest_peak_claimed`, and per-site seeding (round-1 S5) moves every miss-cell `null_pct` by up to 0.075. The paragraph whose purpose is "so a regenerated copy is not mistaken for a changed result" lists two of five differences.
- **Evidence:** `cascade_gate.py` payload carries `"null_rng": "per-site"` and `"panos_without_floor_peaks"`; row dict carries `"nearest_peak_claimed"`. Committed artifacts: neither has `null_rng`, and `nearest_peak_claimed` is absent from every `op030` site (checked). `test_only_the_op030_artifact_records_the_threshold_key` and `test_the_committed_nulls_came_from_one_stream_and_say_so` assert the *pre-fix* key set, so the doc's "Regenerate both when the parity detections are published" also means editing those two tests, which the doc does not say. The S5 caveat 60 lines later has the null half only.
- **Fix:** List all expected differences (three new keys, per-site nulls) in that paragraph, name the two tests that pin the old state, and cross-reference the S5 caveat. No test change now.
- **Found by:** Opus 5

#### S3 · minor · replicability — the same-env control command reads the cache instead of running the model
- **Where:** `docs/model_comparison.md` ~1712 — `python scripts/model_comparison/compare.py benchmark/richmond --models vistas:curb-cut` (labelled "same-env control (the attribution) -- default 384, no override")
- **What:** The doc says the control has the same signature and cache key as the published 384 run (~1429), and 30 lines later tells a clean clone to write the published detections into `.model_cache` first. With both followed, `compare.py` finds all 124 panos cached, prints `all 124 panos already cached; model load skipped`, and never runs the forward: the attribution arm is silently the published arm. The original run avoided this with a private `--cache-dir` that the command block omits.
- **Evidence:** `compare.py` ~270–288: `cached = {pid: cache.get(k) ...}`; the `else` branch skips `detector.prepare()` when nothing is `None`. `--no-cache` and `--cache-dir` exist (`compare.py` 683/686).
- **Fix:** Add `--no-cache` (or `--cache-dir <fresh dir>`) to the control command and one sentence saying why. No test.
- **Found by:** Opus 5

#### S4 · nit · documentation — two of the four "unclaimed" peaks are below 0.30
- **Where:** `docs/model_comparison.md` ~1608 — `| nearest floor peak at 1–2 R, unclaimed (23.5 px @0.643, 24.3 px @0.242, 32.2 px @0.358, 42.5 px @0.143) | 4 | a peak just outside the window; would need a wider radius, not a lower threshold |`
- **What:** 0.242 and 0.143 are under the 0.30 operating point, so a wider radius alone recovers two of the four; the other two need both. The row's own numbers say so.
- **Fix:** "would need a wider radius (and, for two of the four, a lower threshold as well)".
- **Found by:** Opus 5

### Rejected or downgraded
Both Gemini runs returned zero findings. Per the pipeline's rule, their reasoning was read for defects found and then dropped; each candidate below was checked against the code.

- **Gemini 3.1 Pro (reasoning): `no_peak_profile`'s `peak_beyond_2r = len(sel) - within_2r` counts rows with `nearest_peak_px: None` as "beyond 2 R"** — rejected: that is the intended reading (a site with no floor peak on the pano has nothing within 2 R), `within + beyond == n` by construction, and the doc's "no floor peak within 2 R" row means exactly that. The model withdrew it itself.
- **Gemini 3.1 Pro (reasoning): possible division by zero in `summarize`** — rejected: every `/ len(sel)` is behind `if not sel: return {"cell": cell, "n": 0}`.
- **Gemini 3.1 Pro (reasoning): `nearest_peak_index` tie-breaking on equidistant peaks** — rejected: it uses the same strict `d < best` first-wins rule as `silent_activation.nearest_peak`, so the index the claim flag refers to is the peak whose distance and score the row reports.
- **Gemini 3.8 Flash (reasoning): `ValueError` from `min()` over an empty generator in `claim_threshold`** — rejected: the call is `min((...), default=0.0)`; the reasoning transcript truncated the line before `default=`.
- **Gemini 3.8 Flash (reasoning): `KeyError` at the row dict if a `MISS_CELLS` site lacks null data** — rejected: every row is initialised with `null_pct/null_med/null_p95 = None` and every miss-cell row gets `null_percentile` unconditionally, so `summarize`'s `nulls` list is non-empty whenever the cell has rows and the `null_pct_median` print cannot miss its key.
- **Gemini 3.8 Flash (reasoning): `--limit` can lose data** — rejected: `p.error("--limit truncates the run; refusing to write it to --json-out")` guards the only persistent output; the header's site count is the smoke-test's own business.
- **Gemini 3.8 Flash (reasoning): the profile print fails when `nearest_peak_px_median` is `None`** — rejected: it is formatted with `{...}` not `:.1f`, so it prints `None px`.
- **Gemini 3.8 Flash (reasoning): `claimed` is only conditionally present and prints as `?`** — rejected as design: the docstring says the key exists only on rows that carry `nearest_peak_claimed`, which the committed artifacts predate (see S2 for the doc side of that).
- **Gemini 3.8 Flash (reasoning): `trials` is not passed to `null_percentile`** — rejected: the signature is `null_percentile(heat, x, y, rng, trials=NULL_TRIALS, radius_sq=None)`, and the payload records `null_trials: NULL_TRIALS`.
- No Opus finding was rejected or downgraded; S1–S4 above are its F1–F4 with the evidence re-derived here.

### Coverage
- Files excluded from the Gemini reviews: `analysis_out/cascade_gate.json`, `analysis_out/cascade_gate_op030.json` (both runs; 134,651-char diff, 8 files, base `origin/main` at `d964d5d`). Gemini 3.1 Pro exit 0, 434 s, 0 findings, end marker verified. Gemini 3.8 Flash exit 0, 199 s, 0 findings, end marker verified. Neither run failed; both are genuinely empty after the reasoning check above.
- Tests run: full suite in the worktree with `PYTHONPATH=.` — 1385 passed, 1 skipped in 91 s. Re-derived from committed files only: the 15-row per-site table (every value), the 7 / 4 / 4 vs 11 / 4 / 4 split behind S1, `no_peak_profile` for `challenger_only` and `neither`, `class_of` shares, both artifacts' top-level key sets and the absence of `nearest_peak_claimed`, the `f2acf1b1…` shard name for the 1024 signature. `compare.py`'s cache short-circuit read for S3, not run.
- Round-1 fixes verified: S1 (re-cut, wording withdrawn in doc, `cascade_gate.py` docstring and footer, three pinning tests), S3 (`radius_sq` on both `score_pano` calls; the radius test fails on the pre-fix code), S4 (`no_floor` counted in `partition_cells`, `floor_gap_warning`, `panos_without_floor` on the probe path, `panos_without_floor_peaks` in the payload), S5 (`site_rng`; the order-dependence test demonstrates the old failure), S6 (one `partition_cells`, called by both `main`s and the test), S7 (`sys.exit` with the split name, driven through `cx.main()`), S8 (docstring paragraph present), S9 (PR body). S2 deferred, stated accurately in the doc and PR body.
- Not checked: anything needing the unpublished 1024 / A40-control detections (parity P/R/F1/AP, the 1024 complementarity column, the partition inside both cascade artifacts); the GPU half of `cascade_gate.py`; whether the makelab2 cache exists (out of scope for this round); the suite on the merged tree with current `origin/main` (`git merge-tree` is clean, 10+ commits behind).

**Merge readiness:** ready after the listed fixes — four doc edits, no code, no artifact or number changes — with round-1 S2 still open as the owner's decision: the A40 parity and control detections cannot be found on makelab2, so either re-run (3m38s on one A40, both arms) and publish them, or merge with the gap stated as it now is beside every affected number.

---

jonfroehlich:

One commit on `exp/vistas-parity-126`, `5d0c05e..5af6c53`, doc only. No number, test or artifact changes.

- **S1 — fixed in `5af6c53`:** the sentence under the re-cut table now reads "the 15-row … is 7 parts matching problem, 4 parts near-miss geometry and 4 parts absence; with the old 4-row folded in, the 38 are 19 / 11 / 4 / 4", the table cell says "7 of the old 15 row", and a parenthesis notes the 4 absence sites have a claimed nearest peak too, beyond 2 R (51–117 px, scores 0.85 / 0.86 / 0.34 / 0.94). Test: unchanged — `test_the_no_peak_rows_nearest_peaks_are_mostly_claimed_by_an_adjacent_ramp` already pins 11 claimed with 7 within 2 R.
- **S2 — fixed in `5af6c53`:** the "regenerated copy" paragraph lists every expected difference on a re-run — `rampnet_op_threshold` in `cascade_gate.json`, the new `null_rng` and `panos_without_floor_peaks` payload keys, per-site `nearest_peak_claimed`, and every miss-cell null moving by up to 0.075 under per-site seeding — names `test_only_the_op030_artifact_records_the_threshold_key` and `test_the_committed_nulls_came_from_one_stream_and_say_so` as the two tests a regeneration must edit, and cross-references the null caveat.
- **S3 — fixed in `5af6c53`:** the same-env control command now carries `--no-cache`, with a comment saying why: it shares the published 384 run's signature and cache key, so on a clone that has seeded `.model_cache` from the published export it would re-score the published arm instead of running the model.
- **S4 — fixed in `5af6c53`:** the unclaimed-at-1–2 R row says "a wider radius (and, for the two below 0.30, a lower threshold as well)".

Round-1 S2 (the A40 parity and control detections cannot be found on makelab2) is untouched and stays the owner's decision.

`pytest -q` in the worktree with `PYTHONPATH=.`: **1385 passed, 1 skipped** in 94 s.

PR #146, opened 2026-08-19
claude-opus-5 on nine splits: the annapolis displacement does not generalise (#139)

jonfroehlich:
Closes #139.

`claude-opus-5` at `effort=low` ran the eight remaining city splits (984 panoramas). This
publishes the detections, scores them, and answers the question the issue was opened to
settle.

## The annapolis displacement does not generalise

| | annapolis only | pooled, 7 city splits |
|---|--:|--:|
| `claude-opus-5` (low) | 0.588 | **0.588** |
| `gemini-3.1-pro-preview` | 0.567 | **0.608** |
| **delta** | **+0.021** | **−0.021** |

The one split that made this worth running was the one split that flattered it. annapolis is
the survey-grade Trimble MX7 city — the least representative rig in the benchmark — and the
issue pre-registered that as the risk. **`gemini-3.1-pro` remains the best challenger and
RampNet's lead is unchanged at +0.219 F1.**

## The shape survives even though the ranking doesn't, and it is the more useful result

Opus trades **−0.083 precision for +0.041 recall**, which makes it the **highest-recall chat
VLM on the board** (0.614, above every Gemini leg). It loses on the aggregate and wins on the
axis `operating_point.md` says to optimise. The scoreboard's own third finding is *"precision
is not the differentiator; recall is"* — so quoting the ranking alone gets the direction right
and the reason wrong. It is written up as finding 4 under the headline table rather than as a
footnote, for that reason.

budapest replicates as Opus's collapse split (0.378, worst by a wide margin), consistent with
every other chat VLM. paterson is its only precision-strong split (0.752).

## `standing` stays `False`, deliberately

`roster.py` raises **at import** for any pinned leg that is standing — a scored entry has to be
what a bare `--models` spec reproduces, and both efforts of `claude:claude-opus-5` share one
spec. It is also unnecessary: `standing` gates the roster tables (`fp_taxonomy`, `null_recall`,
the frozen `WITNESS_POOL_46`), **not** the scoreboard, which keys off coverage. Opus entered the
headline table on regeneration by itself. `gemini-3.7-flash` and the three YOLO pano arms sit in
exactly the same position. Recorded in the doc's caveat list so the next reader doesn't try the
flip and hit the assertion.

## Cost: $70.41, and layer 1 didn't record it

The run wrote **no row** to `analysis_out/usage_log.jsonl`. Recovered from Cloud Monitoring the
next day (11,988,993 in / 418,503 out) and written up in `docs/model_comparison.md` § "Cost
[...]

PR #147, opened 2026-08-19
Record time as well as money for every non-free run (#143)

### LINKED ISSUE #143: Record time as well as money for every non-free run: the harness times nothing, and the free-GPU half of the roster logs nothing at all
We record **money** for paid APIs and nothing else. A paper's efficiency table wants
seconds-per-panorama next to F1, and we cannot produce one from any committed artifact.

## What already works (don't rebuild it)

#119 made paid-API spend a hard rule enforced in code, and that half is in good shape:

- `compare.py:report_usage` appends one JSONL row per leg to `analysis_out/usage_log.jsonl`
  (committed): token counts, `panos_scored`, detector `signature`, `model_versions` (#121),
  stop reasons, `est_cost_usd`.
- It runs in a `finally`, so a leg that dies or is Ctrl-C'd after paying still records.
- `--usage-log none` requires `--allow-unrecorded-spend`; otherwise `score_model` raises
  `UnrecordedSpend` before the first uncached pano.
- `pricing.py` is verified-only (`as_of` + source per rate); `scripts/analysis/vertex_usage.py`
  reconciles against Cloud Monitoring actuals.
- Written up in `docs/model_comparison.md` § "Cost accounting for paid models".

### 1. Nothing in the harness measures time

`grep -n 'perf_counter\|time.time()\|elapsed' scripts/model_comparison/*.py rampnet/*.py`
returns **zero hits**. `usage_log.jsonl` carries `ts` — a completion timestamp — but no
duration. So the log supports $/pano and cannot support **s/pano**.

The runtimes we do have are hand-transcribed prose in `docs/model_comparison.md` § "Cost and
runtime": *"OWLv2+GDINO pair 1h33m on one L40S; Qwen-8B 1h57m; Qwen-32B 3h46m (two A40s);
Molmo2-8B 3h49m; Gemini pro-preview ~4 h, flash ~9 h."* Those numbers are correct and they are
the only record — not attached to a run, not re-derivable, not checkable from a clean clone.

**This is as unrecoverable as a token count, for the same reason.** A re-run reads
`.model_cache`, makes zero calls, and finishes in seconds — so a leg's inference time can never
be re-measured without paying for the leg again.

### 2. Free-API legs log nothing at all

`report_usage` returns early when the detector has no `usage` dict:

```python
usage = getattr(detector, "usage", None)
if not usage or not usage.get("calls"):
    return
```

Every local GPU model — OWLv2, Grounding DINO, Qwen3-VL 8B/32B, Molmo2, the YOLO arms, RampNet
itself — takes that branch and produces **no row**. Those legs are free in API terms and are
exactly the ones whose cost *is* GPU time. Half the roster is invisible to the ledger that
exists.

### 3. Compute spend has no ledger

`docs/tillicum.md` records the rate model ($0.90/GPU-hour = elapsed x N GPUs), the monthly cap,
and per-job actuals for the jobs we ran (the smoke job: 4.67 GPU-h, $4.20). `docs/tillicum.md`
also records 496.5 GPU-hours burned on the #51 baseline, from `sacct`. The `.slurm` scripts
print "check `hyakusage`".

All of that is prose, written by hand, per job, only when someone remembers. There is no
`compute_log.jsonl` analogous to `usage_log.jsonl`. Unlike token counts this is **partly
back-fillable** — `sacct` retains job records — so the fix can seed itself from history, but
the retention window is finite and shrinking on the older runs.

### 4. The rule is scoped to model comparison, not to the repo

The standing rule is written in `docs/model_comparison.md`, which reads as a convention of that
harness. It is not in `CLAUDE.md` beside the replication and record-in-GitHub rules, and it says
nothing about time or about non-free compute.

## Why now

The precedent is already on the books: #123's four Claude legs spent **$28.82 with no token
record**, unrecoverable, documented in `docs/replication.md`. Time is in exactly the state money
was in before #119 — measured occasionally, by hand, into prose — and #135's Run B decision turns
on GPU-hours, so the compute side is live work, not hypothetical.

## Plan

1. **Time in the run record.** `perf_counter` around `score_model`; add `elapsed_s`,
   `panos_scored`, and a derived `s_per_pano` to the usage row. Split model-load
   (`detector.prepare()`) from inference so a cached-heavy re-run doesn't look fast for the
   wrong reason.
2. **Log free legs too.** Replace the early return with an always-write path: a row for every
   leg, with device / GPU model / world size for local models, `null` token fields where there
   is no API. Rename nothing in the existing schema — the file is committed and append-only, so
   old rows must still parse.
3. **`scripts/analysis/slurm_usage.py`** — scrape `sacct` into a committed
   `analysis_out/compute_log.jsonl`: JobID, JobName, cluster, partition/QoS, Elapsed,
   AllocGPUs, GPU-hours, and cost for Tillicum at the rate in `pricing.py`. Runs
   retroactively so the existing history gets captured before it ages out.
4. **Promote the rule into `CLAUDE.md`**: time *and* money, for any non-free model *or*
   non-free compute, recorded at run time.

### Acceptance

- A `compare.py` run over any bundle, paid or free, appends exactly one row per leg carrying
  both a duration and (where applicable) tokens and dollars.
- `analysis_out/compute_log.jsonl` reproduces the 496.5 GPU-hours in `docs/tillicum.md` and the
  smoke job's $4.20 from `sacct`, without hand entry.
- `pytest -q` stays CPU-only, network-free, and green; new tests cover the free-leg row, the
  duration fields, and back-compat parsing of the rows already committed.

### Non-goals

- Per-call latency distributions. One duration per leg is what a paper reports; per-call
  timing can come later if a latency claim ever needs it.
- Re-running anything to recover times we lost. Those legs are gone; say so in the doc next to
  the numbers rather than paying to re-measure.

jonfroehlich:
Closes #143 (except the Tillicum back-fill — see **Blocked** below).

We recorded **money for paid APIs** and nothing else. This adds the other three quadrants:
time for paid legs, time for free GPU legs, and money+time for cluster compute — and fixes
the defect the comment on #143 identified, which was actively losing data.

## The bug that was live

`DEFAULT_USAGE_LOG` derived from `REPO_ROOT`, which in a linked worktree *is the worktree*. A
leg run from a scratch worktree wrote its ledger there and lost it when the worktree was
removed. That is how the #139 `claude-opus-5` leg spent **$70.41 and left no row**.

#119's guard cannot see this, and the reason matters: it proves a log path was **accepted**,
never that the file it wrote **survives**. The operator followed the rule and lost the record
anyway, so promoting the rule into `CLAUDE.md` would not have caught it either.

Both ledgers now resolve through `git rev-parse --git-common-dir`, which every worktree of a
repo shares, so they all append to one canonical file in the main checkout. An explicit
`--usage-log` into a worktree warns; each logged leg prints the **absolute** path and the
running total; and it falls back to the local checkout when git can't answer (a tarball, an HF
clone) — bookkeeping must never be why a run refuses to start.

## What else changed

**Time is measured.** There was no `perf_counter` anywhere in the harness. `score_model` now
fills a caller-owned timing dict (model-load / inference / panos actually called) and `main`
stops the clock in the `finally`, so a leg that dies partway still reports what it spent.
`panos_called` counts *attempts* — a call that raised burnt wall-clock and, on a paid provider,
still billed.

**Free legs are no longer invisible.** `report_usage` returned early whenever a detector had no
`usage` dict — i.e. OWLv2, Grounding DINO, Qwen, Molmo, YOLO, the entire GPU half of the
roster. They now write the same row with `paid: false`. Two legs deliberately write nothing: a
fully cached re-score, and `--models rampnet`, which replays committed detections rather than
running a model.

**#139's recovered $70.41 is in the ledger**, as a row marked `kind: "recovered"` carrying the
billed total minus what the surviving rows already account for. Cost totals include it —
omitting it under-reported this benchmark's Claude spend by ~200x — while reconciliation never
counts it as a measurement.

**A compute ledger.** `scripts/analysis/slurm_usage.py` → `analysis_out/compute_log.jsonl`,
priced from a new verified-only `COMPUTE_PRICING` in `pricing.py`. **Back-filled from a real
`sacct` pull: 3,991 allocations, 2,684.4 GPU-hours on klone since 2026-07-02**, with the raw
dump committed so it re-derives without a cluster account.

**Reconciliation.** `vertex_usage.py --reconcile` compares the committed ledger against Cloud
Monitoring per model. This is the only check that catches a silent no-write. On the real #139
numbers, before the recovery row existed it reports `claude-opus-5` **UNDER with 11,940,249**
input tokens unexplained — the two richmond smoke legs (48,744 tokens) are logged — and with
that row committed it reports `ok (1 recovered)`. A recovered row is still never counted as a
measurement; it is totalled in its own column and subtracted before the verdict, so the one gap
this PR closes is not re-raised as an emergency on every run.

**The rule is now repo-level**, in `CLAUDE.md` beside replication and record-in-GitHub: both
units, both cost types, ledgers committed into the main checkout.

## Finding: `sacct -D` is worth 4.35x

96% of our klone allocations end in `PREEMPTED` (3,780 of 3,991), so the default `sacct` view —
last incarnation only — discards nearly everything:

| #51 YOLO baseline | rows | GPU-hours |
| :--- | ---: | ---: |
| `sacct -D` | 3,857 | **2,046.9** |
| without `-D` | 27 | 470.5 |

It **validates** the one figure the repo already had. `docs/tillicum.md` records 496.5
GPU-hours as of 2026-07-30; summing this ledger over jobs *ending* from 07-24, the running
total crosses 496.5 at **2026-07-29T21:17** — the evening before that line was written.

## Blocked

- **Tillicum is not back-filled.** Duo-gated and the control master was down. One command once
  it's up; the gap is stated in `docs/compute_cost.md` next to the numbers, not left implicit.

## Overlap

Touches the same section of `docs/model_comparison.md` as
`data/claude-opus-5-nine-splits-139`, in a different hunk (that branch narrates the incident;
this one documents the mechanism and the fix). Should merge clean.

## Tests

44 new, 1,340 passing (1 skipped), still CPU-only and network-free. The reconciliation tests
use the real #139 numbers; the compute tests are checked against `docs/tillicum.md`'s
independently-derived $4.20 / 4.67 GPU-h.


jonfroehlich:
## Deep review

Re-ran everything from the branch head (`37b7faa`) in a clean detached worktree on Windows. **The mechanism is right and the artifacts reproduce.** Three independent checks came back clean:

- Full suite: **1335 passed, 1 skipped** (the body says 1,334; the last commit added one).
- `analysis_out/compute_log.jsonl` **regenerates byte-for-byte** from the committed `sacct_klone_2026-08-19.txt` once `recorded_at` is re-stamped (3,991 rows, LF only, zero duplicate `(cluster, job id, start)` keys). Every number in `docs/compute_cost.md` re-derives: 2,684.4 GPU-h; 3,780/3,991 PREEMPTED; YOLO 3,857 rows / 2,046.9 GPU-h vs 27 / 470.5 without `-D` = **4.35x**; GPU types 2,194 / 1,285 / 456 / 23 / 6 / 1 by allocation; 28 `sal-*`/`arch_*` rows = 34.3 GPU-h = 1.28%; RampNet-only 2,650.0; the 496.5 GPU-h running total crosses at **2026-07-29T21:17:10**. All exact.
- `canonical_repo_root` traced from the main checkout root (`git` prints `.git`), a main subdirectory (`../.git`), a scratchpad worktree root and subdirectory, and a `.claude/worktrees/` nested worktree (all print the absolute common dir): every one resolves to `D:\Git\RampNet`. A non-git temp dir, a nonexistent path, and an empty `PATH` all fall back to `start` without raising.
- Merging this branch onto `main`, onto `data/claude-opus-5-nine-splits-139` (#146), and onto `main` + #146 in sequence: **all three clean**, no conflict in `docs/model_comparison.md`.

So what follows is about the half of the design the code does not yet implement, and about text that went stale between commits.

---

### 1. Medium — the compute ledger's idempotency is half-implemented: superseded rows are re-appended but never de-duplicated by any reader

`slurm_usage.py:38-40` and `new_rows` (line 173) promise: "A job first recorded while RUNNING is re-appended once it reaches a terminal state; readers take the last row per key." **No reader does.** `summarize` (189), `print_by_name` (206), `main`'s "Ledger now holds" (322) and `ledger.ledger_totals` (106) sum every row.

Reproduced with the test fixture's own shapes: a job parsed RUNNING at 3,600 s, then COMPLETED at 18,000 s → `new_rows` correctly re-appends → `summarize(existing + fresh)` reports **6.0 GPU-h for a 5.0 GPU-h job**. The committed ledger is unaffected today only because the 2026-08-19 pull happened to contain no RUNNING rows (one PENDING row, `38640313`, at `start=Unknown`, 0 GPU-h, which will get a *new* key when it starts and so never collides). The first re-pull after any job is caught mid-run double counts it in every total the script prints.

Fix: a `latest_rows(rows)` in `slurm_usage.py` that keeps the last row per `row_key`, applied in `summarize`, `print_by_name` and before the "Ledger now holds" summary. Do **not** put the dedupe in `ledger.ledger_totals` keyed on `row_key`: `usage_log.jsonl` rows have no `job_id`, so they would all collapse to one key. Add a test that asserts 5.0.

### 2. Medium — once a recovery row exists, `--reconcile` reports the recovered gap as an unexplained emergency, forever

`ledger_totals_by_model` skips RECOVERED rows by design (`vertex_usage.py:117`), and the docstring's reason is right: a recovered row must not make the bill agree with itself. But the *output* does not distinguish "billed more than we measured, and nobody has looked" from "billed more than we measured, and the gap is already in the ledger as a recovery." Run against the committed ledger on this branch with the real #139 bill:

```
claude-opus-5   billed 11,988,993   logged 48,744   rows 2   UNDER — ledger short
1 model(s) billed more than the ledger records (11,940,249 input tokens unaccounted for).
A missing layer-1 row is an emergency with a deadline ...
```

That 11,940,249 is exactly the recovered row's `input_tokens`. The check the PR adds will say "emergency" about the one gap this PR closed, every time it is run inside the window. An operator will learn to ignore it, which defeats the check.

Fix: have `ledger_totals_by_model` also total `input`/`output` of RECOVERED rows per model under a separate key (`recovered_input`), keep them out of `logged_input`, and have `reconcile` compute `unexplained = billed − logged − recovered`. Verdict `ok (N recovered)` when the remainder is within tolerance, `UNDER` only on the unexplained part; `print_reconciliation` prints the recovered column and totals the *unexplained* gap. The existing asymmetry and the "never compare the bill to itself" rule both survive: measured rows still have to account for the bill minus what was explicitly written down as unmeasured.

### 3. Medium-low — the read side of the worktree fix was not done: `.env` is still loaded from the worktree

Both entry points still read the repo-root `.env` from the *running file's* checkout: `compare.py:836` (`load_dotenv(str(REPO_ROOT))`) and `vertex_usage.py:54` (`load_dotenv(str(REPO))`). #146's write-up already documents the consequence: "Run `vertex_usage.py` from a worktree and it exits with 'no project'". So the tool that recovers a lost spend still cannot find the project id from the place a leg is most likely to have been run, and a Gemini/Claude leg from a worktree finds no key unless the operator copies `.env`. The PR body says "Both ledgers now resolve through `git rev-parse --git-common-dir`"; the credentials do not.

Fix: load `.env` from `canonical_repo_root()` as well as `REPO_ROOT` (the loader's `setdefault` semantics make loading both harmless; worktree first so a deliberate local override still wins). One line in each file, one test using the same `git worktree add` fixture as `test_the_ledger_resolves_to_the_main_checkout_from_a_worktree`.

### 4. Low — `REQUEUED` is a finished incarnation but is classed as non-terminal

`TERMINAL_STATES` (line 72) omits `REQUEUED`. The committed dump has **59 REQUEUED rows, all with an `End` time and elapsed > 0, carrying 848.5 GPU-h** (e.g. `37745358`, 29,072 s, 1 GPU). They are counted correctly today because they were first seen in that state. The path that goes wrong: a job recorded while RUNNING, then requeued rather than preempted, is never re-appended, so its row keeps the understated elapsed. Fix: add `REQUEUED` to `TERMINAL_STATES`; the `(job id, start)` key already separates it from the next incarnation.

### 5. Low — the PR body is stale relative to its own last commit, and one claim in it is wrong

- "**Blocked:** #139's recovered $70.41 is not in the ledger as a row" — `37b7faa` added it (`usage_log.jsonl` line 4, `kind: "recovered"`).
- "on the real #139 numbers it reports `claude-opus-5` **MISSING** with 11,988,993 input tokens unaccounted for" — against the committed ledger it reports **UNDER with 11,940,249**, because the two richmond smoke legs (48,744 tokens) were logged. MISSING is only what the test fixture with no opus rows produces.
- "39 new, 1,334 passing" — 1,335 on the head.

Fix: edit the body (`gh pr edit 147 --body-file`), moving the recovery row out of "Blocked" and correcting the verdict and the count.

### 6. Low — two doc lines aged between commits

- `docs/replication.md:20`: "back-fill pending" — klone is back-filled on this branch (3,991 rows); only Tillicum is pending. Say which.
- `docs/compute_cost.md:52`: `rampnet_cosine_rung_135` "still in flight" — true of the 2026-08-19 pull, not of the ledger as a statement. Add "as of the 2026-08-19 pull", or the sentence will read as wrong the day the rung finishes (it has, on `analysis/run-b-power-135`).

### 7. Low — `CLAUDE.md` overstates the token pricing table

Line 66: rates carry "the date it was checked and the page it came from". The `COMPUTE_PRICING` entries do carry a `source`; the token `PRICING` rows carry `as_of` and a `note`, no source. Either add `source` to the token rows or say "date checked" only.

### 8. Low — one phrase not in the repo's register

`rampnet/ledger.py:34`: "the difference is load-bearing". Say what it is: "the difference matters for reconciliation".

---

- **The ledger now lands in whatever branch the main checkout has checked out.** That is the point of the fix, but it has a consequence the doc does not state: the row is safe from deletion yet is written into an uncommitted working tree on whatever unrelated branch the main checkout happens to be on. The experiment's own PR has to carry that row, so the finishing step is "copy the new line(s) into the experiment branch and commit", not "commit the ledger". Worth one sentence in `CLAUDE.md`'s "A ledger is not a commit" bullet.
- **The recovered row's `est_cost_usd` (70.095) is the token delta priced at list, not `billed_usd − already_logged_usd` (70.0975).** The ledger sums to $70.4375 against a $70.44 bill. Immaterial, but the row's note says "the BILLED total minus what the surviving rows already account for" and that is true of the tokens, not the dollars. Your call whether to leave it.
- **Whether the report_usage totals line should count free-leg hours.** `ledger now: … 0.0 h` will start climbing once GPU legs write rows; it mixes API wall-clock and GPU wall-clock in one number. Fine as a visibility line, not a cost figure.

### What holds up

- **The bug is real and the fix is the right shape.** `--git-common-dir` is the one thing every worktree shares; the fallback to `start` on any failure means bookkeeping cannot stop a run; the warning on an explicit `--usage-log` into a worktree fires only when the two roots differ.
- **Timing is measured where it has to be.** The clock stops in `main`'s `finally`, so a leg that raised after paying still records; `panos_called` increments in a `finally` inside the loop so `continue` and `break` are both counted; a fully cached leg and `--models rampnet` are the only silent cases and both are tested.
- **`sacct -D` is worth 4.35x, and the 496.5 cross-check is exact to the minute.** The generic-over-typed GPU count is correct for this dump (3,965 rows report both forms, 26 report neither), and no `(cluster, job id, start)` key collides.
- **Recovered rows are excluded from reconciliation for a correct reason** — the fix in finding 2 keeps that reason and changes only what the report says about a gap that is already written down.
- **The compute pricing is sourced** (Tillicum page + provisioning email + `sacctmgr`, with the `debug`/`hyakusage` disagreement carried in the row's note rather than dropped).
- Clean merge onto `main`, onto #146, and onto both in sequence.

## Fix list
- [F1] Medium — `slurm_usage.py`: add `latest_rows()` (last row per `row_key`), use it in `summarize`, `print_by_name` and before the "Ledger now holds" summary; test that RUNNING→COMPLETED sums to 5.0 not 6.0. Keep `ledger.ledger_totals` as is (usage rows have no `job_id`). — files: scripts/analysis/slurm_usage.py, tests/test_slurm_usage.py
- [F2] Medium — `vertex_usage.py`: total RECOVERED rows per model separately (`recovered_input`/`recovered_output`), keep them out of `logged_*`, compute `unexplained = billed − logged − recovered`, verdict `ok (N recovered)` when the remainder is within tolerance, and print the recovered column and the unexplained gap; update `test_a_recovered_row_counts_toward_cost_but_never_toward_reconciliation` and `test_vertex_usage.py` to assert the real committed-ledger outcome (UNDER 11,940,249 without the row; ok-with-recovery once it is written). Update the "Reconcile" paragraph in docs/model_comparison.md and the CLAUDE.md bullet to say the check distinguishes an unexplained gap from a recorded recovery. — files: scripts/analysis/vertex_usage.py, tests/test_vertex_usage.py, tests/test_model_comparison.py, docs/model_comparison.md, CLAUDE.md
- [F3] Medium-low — load `.env` from `canonical_repo_root()` after `REPO_ROOT` in `compare.py:836` and from `ledger.canonical_repo_root(REPO)` in `vertex_usage.py:_load_dotenv`; one test with the worktree fixture. Mention in docs/model_comparison.md "The ledger has to outlive the run" that credentials resolve the same way now. — files: scripts/model_comparison/compare.py, scripts/analysis/vertex_usage.py, tests/test_model_comparison.py, docs/model_comparison.md
- [F4] Low — add `REQUEUED` to `TERMINAL_STATES`; assert `is_terminal("REQUEUED")` in the terminal-states test. — files: scripts/analysis/slurm_usage.py, tests/test_slurm_usage.py
- [F5] Low — `gh pr edit 147 --body-file`: move the $70.41 row out of "Blocked" (it is in as `kind: recovered`, commit 37b7faa), change "MISSING with 11,988,993" to "UNDER with 11,940,249 (the two richmond smoke legs are logged)", and the test count to whatever the suite reports after F1–F4. — files: PR body only
- [F6] Low — docs/replication.md:20 → "klone back-filled 2026-08-19 (3,991 rows); Tillicum pending"; docs/compute_cost.md:52 → "still in flight as of the 2026-08-19 pull". — files: docs/replication.md, docs/compute_cost.md
- [F7] Low — CLAUDE.md:66: "the date it was checked and, for compute, the page it came from" (or add `source` to the token PRICING rows in pricing.py). — files: CLAUDE.md (or scripts/model_comparison/pricing.py)
- [F8] Low — rampnet/ledger.py:34: replace "the difference is load-bearing" with "the difference matters for reconciliation". — files: rampnet/ledger.py

jonfroehlich:
## Review fixes

- **F1** — fixed in b190da0. `latest_rows()` keeps the last row per `(cluster, job id, start)`, and `summarize()` and `print_by_name()` now read through it, so the "Ledger now holds" summary does too. Without it a job first recorded RUNNING and re-appended once finished was billed twice: 6.0 GPU-hours for a 5.0-hour job. `ledger.ledger_totals` is unchanged, as the review asked — `usage_log.jsonl` rows carry no `job_id`. Regenerating `analysis_out/compute_log.jsonl` from the committed `sacct` dump gives the same 3,991 rows, identical apart from the `recorded_at` stamp the script re-writes, so the committed ledger is untouched.
- **F2** — fixed in 749d345. `ledger_totals_by_model` totals RECOVERED rows under `recovered_input`/`recovered_output`/`recovered_rows`, still outside `logged_*`; `reconcile` verdicts on `billed − logged − recovered` and reports `ok (N recovered)` when the remainder is inside tolerance; `print_reconciliation` prints a recovered column and totals the unexplained gap. Against the committed ledger and the real #139 bill this now reads `ok (1 recovered)` instead of raising the same 11,940,249-token emergency on every run. The rule that a recovered row never makes the bill agree with itself is unchanged — it is not counted as a measurement. `docs/model_comparison.md` and the `CLAUDE.md` bullet say so, and the `rampnet/ledger.py` docstring was brought in step in aa4a142.
- **F3** — fixed in baf0b49. New `compare.load_dotenv_for_run()` loads `.env` from the running checkout and then from `canonical_repo_root()`; `compare.main` and `vertex_usage._load_dotenv` both use it. `setdefault` semantics mean the worktree's own file is read first and still wins. Tested with the same `git worktree add` fixture, now factored into a `_main_and_worktree` helper, and noted in `docs/model_comparison.md` under "The ledger has to outlive the run".
- **F4** — fixed in b190da0. `REQUEUED` added to `TERMINAL_STATES` with `is_terminal("REQUEUED")` asserted. The 59 REQUEUED rows in the committed dump (848.5 GPU-hours) were already counted correctly; the path this closes is a job recorded while RUNNING and then requeued, which was never re-appended and kept its understated elapsed.
- **F5** — done: PR body edited. The recovered row moved out of **Blocked**, the reconciliation verdict corrected, and the test count updated. The verdict sentence states both cases, because F2 changed what the check says: **UNDER with 11,940,249** before the recovery row existed (the two richmond smoke legs, 48,744 tokens, are logged), `ok (1 recovered)` with it committed.
- **F6** — fixed in fbba3b7. `docs/replication.md` now says klone back-filled 2026-08-19 (3,991 rows), Tillicum pending; `docs/compute_cost.md` dates the cosine rung's status to the 2026-08-19 pull.
- **F7** — fixed in fbba3b7. `CLAUDE.md` now says "the date it was checked and, for compute, the page it came from" — only `COMPUTE_PRICING` carries a `source`.
- **F8** — fixed in fbba3b7, with the surrounding paragraph reflowed in aa4a142.

Nothing was skipped; every finding held up when checked against the code. Full suite: **1,340 passed, 1 skipped** (was 1,335/1), CPU-only and network-free.

jonfroehlich:

0 major / 4 minor / 3 nit confirmed, from three independent reviews (Claude Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash) verified against the worktree at `aa4a142`. The code is sound: the committed compute ledger re-derives from the committed `sacct` dump (3,991 rows, 0 differences ignoring `recorded_at`), and every number in `docs/compute_cost.md` reproduces (3,857 rows / 2,046.9 GPU-h with `-D` vs 27 / 470.5 without, 4.35x; 3,780 of 3,991 PREEMPTED; 28 `sal-*`/`arch_*` rows = 34.3 h; RampNet-only 2,650.0). The most important fix is S1: the "496.5 GPU-h crosses at 2026-07-29T21:17, to the hour" validation is computed over every job on the account, while the figure it validates is baseline-only; on the baseline the same method crosses on 2026-07-30 at 11:03, hours *after* the line was committed.

**Context, not a finding:** this branch is CONFLICTING with `origin/main` on `analysis_out/usage_log.jsonl` (both sides appended at EOF: main added 3 Laurens `claude-opus-5` rows totalling $12.78, this branch added the `kind: "recovered"` row). The fix pass will merge main and resolve it by keeping both blocks; row order carries no meaning in either ledger.

### Confirmed

#### S1 · minor · scientific integrity — the 496.5 crossing is computed over the wrong population
- **Where:** `docs/compute_cost.md` — `Summing this ledger over jobs **ending** from 2026-07-24 onward, the running total crosses 496.5 GPU-hours at **2026-07-29T21:17** — i.e. the evening before that figure was written. It reproduces to the hour`
- **What:** `docs/tillicum.md` says "496.5 GPU-hours consumed **on the baseline** since 2026-07-24". The doc's by-end running sum uses all job names. Restricted to `yolo_curb_ramp_train`, as the sentence it validates says, the crossing moves 14 hours later.
- **Evidence:** Re-derived from the committed dump via `parse_sacct` (script in the review scratch dir, 40 lines). By-end running sum, all job names: crosses at 2026-07-29T21:17:10 (matches the doc). Baseline only: crosses at **2026-07-30T11:03:20**. `git log -S"496.5"` dates the tillicum line to `9430266`, authored 2026-07-30 07:07 -0700, committed 08:15. A snapshot that mimics `sacct -S 2026-07-24` run at 07:00 (elapsed of every incarnation alive after 07-24, truncated at query time) gives **497.5 GPU-h baseline-only** vs 553.2 all-jobs — so the tillicum figure does validate, and it validates `-D`, but under a different method than the doc states. At the doc's 21:17 instant the baseline snapshot is 450.2.
- **Fix:** Replace the crossing sentence with the snapshot computation (population `yolo_curb_ramp_train`, elapsed truncated at T ≈ 07:00 on 07-30, 497.5 vs 496.5) and commit that script, or soften to "the baseline-only total passes 496.5 during the morning of 2026-07-30, consistent with the date". Two related sentences while there: (a) say that the 2,684.4 / 2,046.9 headline includes **158.0 GPU-h of elapsed-so-far from 3 jobs RUNNING at pull time** (`38304087` alone is 157.5 h) and will change on the next pull; (b) "The window starts 2026-07-02, which is where this account's retained records begin" is not supported — the dump was pulled with `-S 2026-07-01`, so it only shows nothing was alive on 07-01. Test: a `test_slurm_usage.py` case that runs the committed script over the committed dump and asserts the quoted number.
- **Found by:** Opus 5

#### S2 · minor · correctness — `reconcile` reports `ok` for a model the bill never saw
- **Where:** `scripts/analysis/vertex_usage.py` — `elif b_in and abs(unexplained) / b_in > tolerance:`
- **What:** The `b_in and` guard against division by zero also skips the over/under branch entirely when billed input is 0, so logged-but-unbilled tokens fall through to `"ok"`.
- **Evidence:** `reconcile({'m': {'input': 0}}, {'m': {'input': 12594, 'rows': 1}})` → `unexplained_input -12594, verdict 'ok'`. Same result when the model is absent from `billed` altogether. `test_a_short_ledger_is_flagged_under_and_an_over_one_is_not_alarming` only exercises `b_in > 0`.
- **Fix:** Treat `b_in == 0 and (l_in or r_in)` as `over — ledger exceeds billed` (or a distinct `logged, not billed`, which is what a leg run against the wrong GCP project or a mistyped `model_id` looks like). Add a test asserting the verdict for `billed={'m': {'input': 0}}`, logged 12,594.
- **Found by:** Opus 5 / Gemini 3.1 Pro / Gemini 3.8 Flash

#### S3 · minor · replicability — `--from-file` stamps the local login as the cluster user
- **Where:** `scripts/analysis/slurm_usage.py` — `ap.add_argument("--user", default=os.environ.get("USER") or os.environ.get("USERNAME"),`
- **What:** With `--from-file` or `--print-command`, `user` in every ledger row and the `-u` in the printed regeneration command come from the machine running the script, not from the dump. Nothing checks the dump.
- **Evidence:** `python scripts/analysis/slurm_usage.py --print-command --from-file x` on this box prints `sacct -X -D -P -n -u jonf -S 2026-07-01 ...`; the klone account is `jfroehli`. `docs/compute_cost.md` tells the reader to regenerate the dump with `--print-command`. The committed ledger is right (3,991 rows, all `jfroehli`) only because the documented command passes `--user jfroehli` explicitly.
- **Fix:** Add `User` to `SACCT_FIELDS` (sacct emits it) and take `user` from the row, using `--user` only for the live invocation; or make `--user` required when `--from-file` is given. Test: `parse_sacct` of a line carrying a `User` column yields that user regardless of `$USER`.
- **Found by:** Opus 5

#### S4 · minor · correctness — `ledger_totals` double-counts re-recorded jobs
- **Where:** `rampnet/ledger.py` — `usd = sum(r.get("est_cost_usd") or 0 for r in rows)`; `tests/test_slurm_usage.py` — `# Both ledgers name the duration the same, so one reader can total them.`
- **What:** The compute ledger's contract is last-row-per-key (`latest_rows`), because `new_rows` re-appends a job first seen RUNNING once it finishes. `ledger_totals` has no dedup, so the "one reader" the test advertises is wrong for the compute ledger the moment a second pull lands.
- **Evidence:** The committed ledger has 0 duplicate keys today, 3 RUNNING rows (158.0 GPU-h) and 1 PENDING row (`38640313`, start `Unknown`, 0 s). Simulated a second pull with those 3 finished: `new_rows` re-appends 3; `ledger_totals` then reports 3,994 rows and a naive per-row GPU-h sum reads 2,845.4 against `summarize`'s 2,687.4. The PENDING key `(klone, 38640313, Unknown)` is never superseded (later incarnations carry real starts), so it stays as a permanent 0-hour job in `summarize()["jobs"]`.
- **Fix:** Route `ledger_totals` through a `latest_rows`-style dedup when rows carry `job_id` (a no-op for `usage_log.jsonl`), or drop the "one reader" claim and pin the opposite in a test (RUNNING + COMPLETED pair → `ledger_totals` is not the number to quote). Skip `Unknown`-start rows in `parse_sacct`; they carry no elapsed.
- **Found by:** Opus 5

#### S5 · nit · robustness — `sacct_command(None, ...)` raises a TypeError inside `subprocess.run`
- **Where:** `scripts/analysis/slurm_usage.py` — `cmd = sacct_command(args.user, args.since, args.until)`
- **What:** If neither `USER` nor `USERNAME` is set and `--user` is omitted, `None` lands in the argv list. Confirmed: `TypeError: expected str, bytes or os.PathLike object, not NoneType`, and the `except (OSError, SubprocessError)` around it does not catch it. The `--print-command` and error-message paths already guard with `args.user or "$USER"`; the live path does not.
- **Fix:** `ap.error("--user is required when $USER is unset")` before the call, or omit `-u` entirely when user is None (sacct defaults to the caller). Rated nit: a login node without `$USER` is not a realistic state.
- **Found by:** Gemini 3.1 Pro / Gemini 3.8 Flash

#### S6 · nit · replicability — `replication.md` size is wrong and the derived ledger is not pinned to its input
- **Where:** `docs/replication.md` — `| analysis_out/compute_log.jsonl (cluster GPU-hours + $) | ~KB | **committed** ✅ |`
- **Evidence:** `compute_log.jsonl` is 1,927,952 bytes; the dump is 765,754 bytes. Rows carry `recorded_at`, so the ledger can never be byte-identical on regeneration, but the dump can be.
- **Fix:** Say "~2 MB, grows per pull". Record the dump's sha256 in `docs/compute_cost.md` beside the command and have `--from-file` print it.
- **Found by:** Opus 5

#### S7 · nit · scientific integrity — two figures in the standing rule have no pointer, and one ratio is undated
- **Where:** `CLAUDE.md` — `it came within 0.02% for a 984-panorama leg); **never estimate output**, which is thinking plus box count and ran 62% high when tried.` and `under-reported the benchmark's Claude spend by ~200x`
- **Evidence:** `grep -rn "0.02%\|62% high"` over `docs/` and `scripts/` hits only `CLAUDE.md`; the figures come from the body of [#146](https://github.com/ProjectSidewalk/RampNet/pull/146), not a committed doc. "~200x" is $70.41 against the $0.34 of measured Claude rows on 2026-08-19; after merging main's Laurens rows the measured total is $13.12 and the same sentence reads ~5x.
- **Fix:** Append "([#139](https://github.com/ProjectSidewalk/RampNet/issues/139), [#146](https://github.com/ProjectSidewalk/RampNet/pull/146))" or move the sentence to `docs/model_comparison.md` once that merges. Date the ratio: "at the time, measured Claude rows totalled $0.34".
- **Found by:** Opus 5

### Rejected or downgraded
- **Gemini 3.8 Flash: `report_usage` never populates `input_tokens` etc. in `rec`, so `test_a_paid_leg_records_time_beside_the_tokens` fails** — rejected: `rec` is built with `**usage` (compare.py, after `"stop_reasons"`), which splats every key of `detector.usage`; the test passes.
- **Gemini 3.8 Flash: `compare` is referenced but never imported in `tests/test_model_comparison.py`, four tests crash with NameError** — rejected: `import compare  # noqa: E402` is at line 34; all four named tests pass (`5 passed, 168 deselected`).
- **Gemini 3.1 Pro: `is_terminal` raises IndexError on a whitespace-only State** — partly valid: `is_terminal(" ")` does raise (`"".split()[0]`), but `parse_sacct` strips each line and `sacct -P` never emits a whitespace-only State, and `""`/`None` are already handled. Not reachable from real input; hardening is one `[:1]` if wanted.
- **Opus 5: merge conflict on `usage_log.jsonl`** — true (`git merge-tree --write-tree origin/main HEAD` → CONFLICT), but the resolution is decided (merge main, keep both blocks) so it is recorded above as context, not argued as a finding. The undated "~200x" part of it survives as S7.
- **Gemini 3.1 Pro / 3.8 Flash: `None` user crashes `subprocess.run`** — confirmed but downgraded from "Crash / Correctness" to nit (S5): needs `$USER` unset on a Slurm login node.

### Coverage
- Files excluded from both Gemini reviews (from the `.log` files): `analysis_out/compute_log.jsonl`, `docs/data/compute/sacct_klone_2026-08-19.txt`. Base `origin/main` (`0cba7b2`), 16 files / 2,200 diff lines on the wire, end markers verified, exit 0 for both. The Gemini 3.8 Flash run was re-run once: the first attempt exited 4 on agy's own output-token limit; the findings above are from the successful second run.
- Tests run: `pytest -q tests/test_slurm_usage.py tests/test_vertex_usage.py tests/test_model_comparison.py` → 199 passed (Opus 5 also ran the full suite: 1,340 passed, 1 skipped). Four-line probes for S2, S5 and the `is_terminal` claim; `--print-command --from-file` run for S3; re-derivation of every `compute_cost.md` figure from the committed ledger; the 496.5 crossing re-derived from the committed dump both ways (S1); a simulated second `sacct` pull for S4.
- Not checked: live `sacct` on a cluster; the Cloud Monitoring fetch (credentials); `hardware_note()` on a GPU host; the `--usage-log`-into-a-worktree warning end to end (read only, no test covers it).

---

jonfroehlich:

Pushed `aa4a142..7b5ef17`. The branch first merges `origin/main` (7a3474a): the `usage_log.jsonl` conflict is resolved by keeping both sides' rows in the order they were appended (this branch's recovered row, then main's three Laurens rows) — 7 rows, none duplicated.

- S1 — fixed in 2beb175 (+ 7b5ef17): the 496.5 check is now a baseline-only snapshot (each incarnation's elapsed truncated at the query instant), done by the committed `scripts/analysis/gpu_hours_as_of.py`: 497.5 GPU-h at 07:00 on 2026-07-30, passing 496.5 at about 06:50, seventeen minutes before the line was committed; 85.8 without `-D`, 553.2 account-wide. The doc says the by-end 21:17 figure was a coincidence of two mistakes. Also added: the 2,684.4 headline holds 158.0 GPU-h from 3 RUNNING jobs; 2026-07-02 is where the `-S 2026-07-01` pull begins, not the account's retention; and the PREEMPTED share paired with 3,780 rows is 95%, not 96% (96% is with the 59 REQUEUED). Tests: `test_a_snapshot_counts_elapsed_so_far_not_finished_jobs`, `test_the_tillicum_496_5_reproduces_from_the_committed_dump_as_a_baseline_snapshot`, `test_the_snapshot_script_prints_the_documented_figure`.
- S2 — fixed in d7bedf1: billed 0 with logged tokens is a new verdict `LOGGED, NOT BILLED`, reported on its own line (not the missing-row emergency). Tests: `test_tokens_logged_against_a_model_the_bill_never_saw_are_not_ok`, `test_the_report_names_the_unbilled_rows_without_crying_emergency`.
- S3 — fixed in a064d49: `--user` has no default; required with `--from-file`, live falls back to `$USER` on the login node; `--print-command` never substitutes the local login. Tests: `test_a_dump_must_name_the_account_it_came_from`, `test_print_command_never_substitutes_the_local_login`.
- S4 — fixed in dfaa8e2: `row_key`/`latest_rows` moved to `rampnet.ledger`, `ledger_totals` reads through them (rows without `job_id` pass through untouched); `parse_sacct` skips PENDING rows. **This changes a committed artifact by one row:** the PENDING `38640313` row is removed from `compute_log.jsonl`, counts move 3,991 → 3,990 (GPU-hours unchanged, 2,684.4), and the cosine-rung by-name row is 1 job. Tests: `test_the_shared_reader_counts_a_re_recorded_job_once`, `test_api_rows_are_never_collapsed_by_the_shared_reader`, `test_a_pending_row_is_not_a_job_allocation`, `test_the_committed_ledger_is_exactly_what_the_committed_dump_parses_to` (3,990 rows, same order, only `recorded_at` differs, every headline figure re-derived).
- S5 — fixed in a064d49: `sacct_command(None, ...)` omits `-u`. Test: `test_no_user_means_sacct_default_not_a_none_in_argv`.
- S6 — fixed in 2b3452b: dump size and sha256 `d6597d97…c69d` recorded in `compute_cost.md`; `--from-file` and `--save-raw` print the hash; `replication.md` says "~2 MB, grows per pull" and names the dump. Test: `test_from_file_prints_the_dump_hash_and_the_doc_pins_the_committed_one`.
- S7 — fixed in 59a31db: `CLAUDE.md` dates the ~200x ratio ($0.34 measured on 2026-08-19; $13.12 after the Laurens legs, ~5x) and says the 0.02% / 62% figures come from the [#139](https://github.com/ProjectSidewalk/RampNet/issues/139) recovery worked in [#146](https://github.com/ProjectSidewalk/RampNet/pull/146), not a committed script.

`docs/model_comparison.md` and `docs/replication.md` were touched only where the findings required (one table row in `replication.md`; nothing in `model_comparison.md`), so [#146](https://github.com/ProjectSidewalk/RampNet/pull/146) stays mergeable with this branch.

Full suite on the final tree: **1531 passed, 1 skipped** (baseline on the merged tree before fixes: 1518 passed, 1 skipped).

jonfroehlich:

0 major / 3 minor / 3 nit confirmed, from three independent reviews (Claude Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash) verified against the worktree at `7b5ef17`. **All 7 round-1 items are confirmed properly fixed and pinned by tests** — re-derived here, not taken from the fixes comment: the committed dump is 765,754 bytes / sha256 `d6597d97…c69d`, parses to 3,990 rows with 0 differences against the committed ledger ignoring `recorded_at` and 0 rows for `new_rows`; the baseline snapshot is 497.5 GPU-h / 400 incarnations at 07:00, first minute ≥ 496.5 is 06:50, 85.8 without `-D`, 553.2 account-wide, by-end baseline crossing 2026-07-30T11:03:20; 3,857 / 2,046.9 vs 27 / 470.5 = 4.35x; 3,780 PREEMPTED = 94.7%, 96.2% with the 59 REQUEUED; 3 RUNNING rows = 158.0 GPU-h; `reconcile` on billed 0 / logged 12,594 and on an absent model both give `LOGGED, NOT BILLED`, `(0, 0)` stays `ok`; `--from-file` without `--user` exits 2 and `--print-command` emits no `-u`; `sacct_command(None, …)` omits `-u`; `latest_rows` collapses RUNNING+COMPLETED to the 5.0-hour row and passes API rows through; the PENDING `38640313` line is the one dump line not parsed; `CLAUDE.md` dates the $0.34 / $13.12 / ~5x ratio (the 3 measured Claude rows sum to $0.3425, the 7-row ledger to $13.12, 70.41 / 13.12 = 5.4). The merge `7a3474a` brings main's changes in verbatim (same file set, identical hunks on every file but `usage_log.jsonl`, which is the exact 7-row union). The thing to fix before merge is S1 + S2: the branch made `--user` mandatory and sha256-pinned the dump, and one documented command and one `.gitattributes` line were not updated to match — each is a one-line change.

### Confirmed

#### S1 · minor · replicability — the `model_comparison.md` copy of the `--from-file` command now exits 2
- **Where:** `docs/model_comparison.md` line 1793 — `python scripts/analysis/slurm_usage.py --cluster klone --from-file sacct_klone.txt`
- **What:** The S3 fix (`a064d49`) made `--user` required with `--from-file`. `compute_cost.md` and the module docstring were updated; this copy was not.
- **Evidence:** `python scripts/analysis/slurm_usage.py --cluster klone --from-file docs/data/compute/sacct_klone_2026-08-19.txt --dry-run` → `error: --user is required with --from-file …`, exit 2. `grep -n "from-file" docs/*.md CLAUDE.md` shows this is the only stale copy.
- **Fix:** Add `--user <account>` to that line. The round-1 fixes comment left `model_comparison.md` untouched to keep [#146](https://github.com/ProjectSidewalk/RampNet/pull/146) mergeable; this is a different hunk from #146's, and a one-word change either way. No new test: `test_a_dump_must_name_the_account_it_came_from` already pins the behaviour.
- **Found by:** Opus 5

#### S2 · minor · replicability — the sha256-pinned dump has no line-ending pin, so a default Windows clone fails the hash test
- **Where:** `.gitattributes` — no rule for `docs/data/compute/`; `docs/compute_cost.md` line 20 — `(765,754 bytes, sha256 \`d6597d97…c69d\`)`
- **What:** With `core.autocrlf=true` (the Git-for-Windows installer default) the dump checks out as CRLF, the pinned hash does not match, and the test that pins it fails. The repo already handles exactly this case for `docs/data/rampnet1_stage1_run/* -text`, with a comment saying why.
- **Evidence:** `git -c core.autocrlf=true clone --shared` of the worktree, checkout `7b5ef17`: `git ls-files --eol` → `i/lf w/crlf` for `docs/data/compute/sacct_klone_2026-08-19.txt`, `analysis_out/compute_log.jsonl` and `analysis_out/usage_log.jsonl` (the stage-1 files under the existing `-text` rule stay `w/lf`). The dump is 769,745 bytes, sha256 `5ee22a8b…`. `pytest tests/test_slurm_usage.py` in that clone → `1 failed, 26 passed`: `test_from_file_prints_the_dump_hash_and_the_doc_pins_the_committed_one`, `5ee22a8b… != d6597d97…`. Every number still reproduces (the parser strips each line), so this is the hash check alone — which is the one check that exists to prove the input is byte-identical. Jon's box is `core.autocrlf=input` and CI is Linux, so neither sees it.
- **Fix:** `docs/data/compute/* -text` in `.gitattributes`, with the same comment as the stage-1 block. `analysis_out/*.jsonl -text` is worth adding in the same hunk: both ledgers check out CRLF under this setting and `append_rows` then appends LF rows to a CRLF working copy (git renormalises on commit, so the committed bytes stay LF, but the file on disk is mixed). The existing hash test is the guard once the attribute is there.
- **Found by:** Opus 5

#### S3 · minor · scientific integrity — the PR body still states the numbers and the validation that the branch retracted
- **Where:** PR #147 description — `3,991 allocations`; `96% of our klone allocations end in \`PREEMPTED\` (3,780 of 3,991)`; `the running total crosses 496.5 at **2026-07-29T21:17** — the evening before that line was written`; `44 new, 1,340 passing`
- **What:** The body is what the merge commit and the GitHub record present as the finding. `docs/compute_cost.md` on this branch says 3,990, 95% (96% only with REQUEUED), and that the 21:17 by-end figure "was a coincidence of two mistakes, not a validation"; the suite is 1,531 passing. The round-1 human review's item 5 asked for a body edit and part of it was done (the reconcile verdict now reads `UNDER with 11,940,249`); the round-1 fix pass did not touch the body again.
- **Evidence:** `gh pr view 147 --json body` at review time contains all four strings quoted above.
- **Fix:** `gh pr edit 147 --body-file`: 3,990 / 95% (96% with REQUEUED) / replace the 21:17 sentence with the 06:50 baseline snapshot from `gpu_hours_as_of.py` / 1,531 tests. No test.
- **Found by:** Opus 5

#### S4 · nit · scientific integrity — "seventeen minutes before the line was committed" is the author time, not the commit time
- **Where:** `docs/compute_cost.md` line 61 — `The baseline passes 496.5 at about 06:50 that morning, seventeen minutes before the line was committed.`; same wording in the `scripts/analysis/gpu_hours_as_of.py` docstring
- **Evidence:** `git show 9430266 --format='%ad %cd' --date=iso` → authored `2026-07-30 07:07:23 -0700`, committed `08:15:25 -0700`. `gpu_hours_as_of` at 08:15 on the baseline → 503.9 GPU-h / 416 incarnations. Line 48 of the same doc correctly says "written 2026-07-30 at 07:07".
- **Fix:** "seventeen minutes before the line was written (author time 07:07; the commit landed at 08:15, by which point the snapshot reads 503.9)". No test.
- **Found by:** Opus 5

#### S5 · nit · simplification — `compare.py` re-implements `ledger.append_rows`
- **Where:** `scripts/model_comparison/compare.py` — `with open(usage_log_path, "a", encoding="utf-8", newline="") as f:`
- **What:** `report_usage` does its own `os.makedirs` + open-append-write, with the same `newline=""` comment that `rampnet/ledger.py` carries. `ledger` is already imported there (`ledger.ledger_totals` two lines later). A future change to how a ledger line is written has to be made twice.
- **Fix:** `ledger.append_rows(usage_log_path, [rec])` inside the existing `try`; the `except OSError` still wraps it. `test_a_free_leg_still_gets_a_row` and `test_a_paid_leg_records_time_beside_the_tokens` cover it.

#### S6 · nit · robustness — `parse_sacct` drops a malformed record with no count and no message
- **Where:** `scripts/analysis/slurm_usage.py` — `if len(parts) != len(COLUMNS):` / `continue  # a header, a warning, a wrapped line: not a job record`
- **What:** Gemini's trigger is a job name containing `|`, which `sacct -P` would split into extra columns. None of the 3,991 lines in the committed dump has a column-count mismatch and no job name on this account has one, so nothing is lost today; I did not run `sacct` to confirm it leaves `|` unescaped. What is certain is that a dropped record leaves no trace: `parse_sacct` of a line with `yolo|baseline` as the name returns 0 rows and prints nothing.
- **Fix:** Count the non-blank lines skipped and print the number (with the first offending line) at the end of `parse_sacct`; a test with one such line asserting the count. Downgraded from Gemini's "Correctness".

### Rejected or downgraded
- **Gemini 3.8 Flash: `report_usage` writes `"input_tokens": usage.get("input_tokens")` and so puts `None` token keys on free rows, failing `test_a_free_leg_still_gets_a_row`** — rejected: the quoted line does not exist anywhere in the tree (`grep -rn 'usage.get("input_tokens")' scripts/ rampnet/ tests/` → nothing). `rec` is built with `**usage`, which adds no keys for a detector with no `usage` dict; the named test passes (212 passed in the three test files). This is the same non-existent line Flash reported in round 1, in a different form.
- **Gemini 3.1 Pro: `is_terminal` raises `IndexError` on a whitespace-only State** — rejected, repeat of a round-1 item already answered: `" "` does raise, but `sacct -P` never emits a whitespace-only State and `""`/`None` are handled. Not reachable from real input.
- **Gemini 3.1 Pro: the `except` handler in `report_usage` calls `json.dumps(rec)` again and re-raises when `rec` is unserializable** — rejected: the handler is `except OSError`, and a `TypeError` from `json.dumps` is not an `OSError` (four-line test: `except OSError` does not catch it), so the handler never runs for that case. Every field of `rec` is a str/int/float/dict/list by construction — `hardware_note()` returns the hostname and a list of device-name strings, `signature()` is already `json.dumps`-ed for the cache key — so there is no value that makes the first `json.dumps` raise either.
- **Gemini 3.1 Pro: `|` in a job name silently drops the record** — partly valid, kept as S6 at nit: the silent drop is real and certain; the trigger is hypothetical for this account and not verified against a live `sacct`.

### Coverage
- Files excluded from both Gemini reviews (from the `.log` files): `analysis_out/compute_log.jsonl`, `docs/data/compute/sacct_klone_2026-08-19.txt`. Base `origin/main` (`2916ba5`), 17 files / 2,642 diff lines, 163,086 chars on the wire, end markers verified, exit 0 for both. **Both Gemini runs were re-run once**: the first attempts exited 4 on agy-side transient errors (a stream interruption for one, the output-token limit for the other); the findings above are from the successful retries.
- Tests run: `pytest -q tests/test_slurm_usage.py tests/test_vertex_usage.py tests/test_model_comparison.py` → 212 passed; full suite → 1,531 passed, 1 skipped (107 s), matching the fixes comment; `tests/test_slurm_usage.py` in an `autocrlf=true` clone → 1 failed / 26 passed (S2). A 90-line script re-derived every round-1 figure listed in the summary from the committed dump and ledger. Four-line probes: `except OSError` vs `TypeError`; a `|` in the job name; the documented `--from-file` command without `--user`; `gpu_hours_as_of.py` as documented (prints 497.5 / 400). Merge `7a3474a` checked hunk-by-hunk against `origin/main`.
- Not checked: live `sacct` or Cloud Monitoring (no cluster session, no credentials); `hardware_note()` on a GPU host; whether `sacct -P` escapes `|` in `JobName`; sacct's `-S` selection semantics for an incarnation that ended before `-S` while a sibling ran after — modelled by `gpu_hours_as_of`, and the 497.5-vs-496.5 agreement is the only evidence the model is right.

**Merge readiness:** ready after the listed fixes — S1 and S2 are one-line changes that stop a new reader on a default Windows clone from hitting an argparse error and a red hash test respectively; S3 is a body edit; S4–S6 can follow or be skipped.

---

jonfroehlich:

Pushed `7b5ef17..410278e`.

- S1 — fixed in 4a40669: the `docs/model_comparison.md` `--from-file` example now passes `--user <cluster account>` (one line; the only edit to that file). Test: `test_a_dump_must_name_the_account_it_came_from` already pins the behaviour.
- S2 — fixed in b5c5b46: `.gitattributes` gains `docs/data/compute/* -text` and `analysis_out/*.jsonl -text`, with the same comment as the stage-1 block. Verified in a `core.autocrlf=true` clone: the dump and both ledgers check out `w/lf`, hash `d6597d97…` unchanged. Test: `test_from_file_prints_the_dump_hash_and_the_doc_pins_the_committed_one` now also asserts both patterns carry `-text`/`binary` in `.gitattributes` and that the dump on disk has no CRLF, so a missing pin fails on every platform.
- S3 — fixed (PR body edited via `gh pr edit --body-file`, no session link): 3,990 allocations; 95% PREEMPTED (96% with the 59 REQUEUED); the 21:17 sentence replaced by the 06:50 baseline snapshot from `gpu_hours_as_of.py`, with the earlier figure noted as retracted; the 158.0 GPU-h RUNNING caveat; 1,531 tests. No test.
- S4 — fixed in 3321dd2: "written (author time 07:07; the commit landed at 08:15, by which point the snapshot reads 503.9)" in `compute_cost.md` and the script docstring. No test.
- S5 — fixed in 3d10e8d: `report_usage` writes through `ledger.append_rows(usage_log_path, [rec])` inside the existing `try`/`except OSError`. Covered by `test_a_free_leg_still_gets_a_row`, `test_a_paid_leg_records_time_beside_the_tokens`.
- S6 — fixed in 410278e: `parse_sacct` prints `WARNING: N non-blank line(s) did not have the 13 pinned columns` with the first offending line; the committed dump reports none. Test: `test_lines_that_are_not_job_records_are_skipped_but_counted` (two junk lines counted, a `yolo|baseline` job name dropped and counted, a clean parse prints nothing).

`scripts/analysis/vertex_usage.py` untouched.

Full suite on the final tree: **1531 passed, 1 skipped**.

Issue #156, opened 2026-09-04
Score the frontier releases: Claude Fable 5/5.1 and GPT-6 Astra (#122 roster refresh)


jonfroehlich:
## Why

"Did you try the newest models?" is the first question a reviewer asks about a benchmark whose
headline is "RampNet beats every general-purpose model." Two frontier releases landed in the last
week, so the answer is currently "no", and the registry in #122 exists precisely so that answer is
cheap to refresh rather than a bespoke project each time.

This issue covers two candidates that are **not** the same amount of work. They are independent —
land either without the other.

## What we already know, so nobody expects a surprise

Three separate results in this benchmark say general capability does not transfer to this task:

- **Qwen3-VL 32B scores worse than Qwen3-VL 8B** — 0.018 F1 against 0.161 on `laurens_gsv`.
  Scaling flipped the failure mode instead of fixing it: the bigger model stops firing.
- **`claude-opus-5` at effort `high` loses 0.068 F1 to the same model at effort `low`** on
  annapolis (0.520 vs 0.588), burning 127,227 thinking tokens to do it. Effort is an
  operating-point dial, not a quality lever (#123).
- **Open-detector recall is mostly density.** OWLv2 and Grounding DINO post 0.85–0.88 recall on
  `laurens_gsv` at 0.028 precision — 6,000+ false positives. Recall without precision is coverage,
  not detection.

Finding small, distant curb ramps in a 4096x2048 equirectangular panorama is a localization
problem. Frontier reasoning gains have not touched it so far. The prior is that a new frontier
model lands near gemini-3.1-pro (0.28–0.57 F1 depending on split), not near RampNet (0.66–0.91).

**That prior is the reason to run it, not a reason to skip it.** A negative result here is
publishable content (see CLAUDE.md), and it is the cheapest possible answer to the reviewer
question. What we must not do is spend $150 discovering it.

## Protocol (preregistered, so the result is not a fishing expedition)

1. **One split first: `annapolis`.** It is the established beachhead for paid legs — four Claude
   legs already published there, with `tests/test_claude_annapolis_leg.py` recomputing every
   published number from committed detections with no API key, no GPU and no `.model_cache`.
   A new leg inherits that guarantee.
2. **Expand to the full 10 splits only if the model clears `gemini-3.1-pro-preview` (0.567 F1)
   on annapolis.** Below that it is a data point, not a challenger, and one split says it.
3. `standing=False` in the roster until it has been run on enough splits to join the tables —
[...]

PR #157, opened 2026-09-05
Claude Fable 5 and 5.1 on annapolis, via a second serving path (#156)

jonfroehlich:
Closes item A of #156. Adds a first-party Anthropic serving path to `ClaudeDetector`, runs both Fable ids on annapolis, and publishes/registers/documents the results.

## Why a second serving path

Vertex gates the whole Fable family behind a project-level publisher data-sharing setting we do not control. Probed 2026-09-04 and again 2026-09-05: both Fable ids return 403 there, `claude-opus-5` returns 200, and a deliberately misspelled `claude-fable-5.1` returns 404 — which is what proves the 403s resolved real ids behind a per-model gate rather than a blanket publisher block.

Anthropic's first-party API serves both ids today, so these legs ran there: a different account and a different rate card for the same weights.

## Results

Annapolis, `--claude-effort low`, 125/125 panos each.

| model | P | R | F1 | tp/fp/fn | boxes/pano | thinking tok | cost |
| :--- | ---: | ---: | ---: | :--- | ---: | ---: | ---: |
| `claude-fable-5` | 0.579 | 0.646 | **0.611** | 190/138/104 | 2.72 | 23,699 | $18.47 |
| `claude-fable-5-1` | 0.637 | 0.585 | **0.610** | 172/98/122 | 2.28 | 254 | $19.86 |
| *`claude-opus-5` (low)* | *0.572* | *0.605* | *0.588* | *178/133/116* | *2.56* | *523* | *$8.94* |

Both displace `claude-opus-5`, which had displaced `gemini-3.1-pro-preview` (0.567) — the first time a general-purpose model has beaten Opus on this split. RampNet still leads by 0.228.

**The two ids are the same detector at different operating points.** 0.610 against 0.611 is nothing; 5.1 trades 0.061 recall for 0.058 precision and emits 0.44 fewer boxes/pano. Same shape as the effort finding in #123 and the Qwen 8B→32B inversion — the knob moves along the P/R curve, the ceiling does not move. Under this project's recall-first framing that argues for `claude-fable-5`, the *older* id.

**#156's always-on-thinking cost premise is refuted.** It predicted a band "wider than a flat 2x" because Fable cannot disable thinking. It is a flat 2x: ~33 thinking tokens/call on 5, ~0.35 on 5.1, against Opus-low's ~0.7. A 5-pano calibration predicted full-leg cost to within 1.5% on both ids.

## The load-bearing design decision

**The serving path is deliberately NOT in the detection cache key.** It names which account answered, not what was asked. Putting it in `signature()` would fragment the cache along an axis that does not change the answer, and would orphan $28.82 of already-paid annapolis detections the moment any leg moved. `test_claude_cache_key_is_frozen` still passes on all four published Vertex legs, and a new test asserts the two paths hash identically.

Provenance goes to the usage log and the published file's new `pins` field instead — `model_versions` cannot stand in, because both paths report the bare model id.

This makes `claude_serving_path` the first roster pin that is not a signature field, which is documented where `pins` is defined. If the two paths ever do diverge in output, that is a finding to measure, not a cache key to add quietly.

## Guards that caught real problems

- The exporter wrote the **bare model id** — the exact failure `publication_name`'s docstring warns about. `leg_for()` could not match a leg pinned on a knob the exporter had no flag for. Added and wired through.
- `test_each_published_file_names_the_leg_it_says_it_is` assumed every pin is a signature field. The envelope now carries `pins` and the test looks there for pins the signature does not hold; legacy files never reach that branch.
- `test_every_registered_leg_has_published_detections` refused the roster rows when I added them *before* the run. They land here with measured density.
- `PUBLISHED_LEGS` keys grew a serving-path element: six rows come from Vertex and two do not, and a fixture that could not say which would hide the caveat these numbers most need to carry.

Also fixes a stale pointer — the docs told readers to run `tests/test_claude_annapolis_leg.py`, renamed to `test_claude_published_legs.py` some time ago.

## Caveats that travel with the numbers

- **Serving path**, recorded next to every number in `docs/model_comparison.md`. Whether the two paths return identical detections for one model id is untested; nothing here depends on it, since no model was run on both.
- **Reconciliation is weaker on this path.** `scripts/analysis/vertex_usage.py` recovers server-side spend for Vertex and has no first-party equivalent, so for these legs the committed ledger is the only record.
- **1 of 8 pooled splits**, `standing=False`, excluded from pooled tables.
- Prompt is fixed and Gemini-derived; JPEG q90 encoding and unpinned decoding carry over from the existing Claude legs.

## Not done, deliberately

**The expansion is not taken.** Both legs clear the pre-registered 0.567 gate, so it is now a live decision rather than a hypothetical — ~$155/id for `manual_gold` alone, ~$390/id for all remaining splits. Priced in the doc, awaiting a call. Item B (GPT-6 Astra) is untouched and independent.

## Verification

`pytest -q` — 1,495 passed, 1 skipped. Both new rows recompute from committed detections with no API key, no GPU and no `.model_cache`. `export_model_cache.py --verify` reports both legs score identically to the cache.

Total spend: $41.01, all recorded in `analysis_out/usage_log.jsonl`.


jonfroehlich:

0 major / 2 minor / 2 nit confirmed, plus 1 finding that is partly valid but cannot be settled from anything recorded in the repo. The PR's numbers reproduce: both published files carry `pins`, 125 panos, `n_uncached: 0`; every new ledger row recomputes to the cent from its own `pricing` block; thinking/call (32.9 and 0.35), boxes/pano, and the P/R deltas in the doc all check. The most important fix is the exporter: with the PR's own default of `claude_serving_path=vertex`, exporting either Fable leg without `--claude-serving-path anthropic` silently writes the bare model id with `pins: {}` — the exact failure `publication_name`'s docstring exists to prevent, and one this PR made easier to hit.

### Confirmed

#### S1 · minor · silent failure / replicability — `publication_name` falls back to the bare id when every candidate leg is pinned and none matches
- **Where:** `scripts/analysis/export_model_cache.py` line 126 — `return roster.published_name(leg) if leg is not None else spec_label(spec, cargs)`; `rampnet/roster.py` `leg_for` returns `None` when all candidates carry pins and none match `cargs`.
- **What:** Both Fable legs are pinned on `claude_serving_path=anthropic`, but `PROVIDER_DEFAULTS["claude_serving_path"]` is `vertex`. Forget the flag and the export succeeds (the path is not in the cache key, so the lookup hits), names the file `claude-fable-5__annapolis.json`, writes `pins: {}`, and collides with nothing, so the overwrite guard is silent. The effort pin never had this problem because its default (`low`) matches the published legs; the serving-path default matches neither.
- **Evidence:** run in the worktree with `PROVIDER_DEFAULTS` as `cargs`: `leg_for("claude:claude-fable-5", defaults) -> None`; `publication_name(...) -> 'claude-fable-5'`; with `claude_serving_path='anthropic'` -> `'claude-fable-5-effort-low-anthropic'`. Same pre-existing hole: `publication_name("claude:claude-sonnet-5", effort='medium') -> 'claude-sonnet-5'`. It is caught only later by `tests/test_roster.py::test_every_published_detections_file_belongs_to_a_registered_leg`, not at export time — which is why this is minor rather than major.
- **Fix:** in `publication_name` (or `leg_for`), when the spec has registered candidates that are all pinned and none matches, raise and list the pin sets the registry knows for that spec. Pin it with a test that `publication_name("claude:claude-fable-5", <PROVIDER_DEFAULTS>)` raises rather than returning `'claude-fable-5'`.
- **Found by:** Opus 5

#### S2 · minor · doc contradicted by its own table — `model_scoreboard.md` prose was not updated with the regenerated partial table
- **Where:** `docs/model_scoreboard.md` line 120 — `**Claude Opus 5 at low effort is the strongest challenger measured on annapolis** (F1 0.588, ...)`; line 96 `Six legs have run one split each`; lines 389–390 `**Six legs have one split each** ... the four Claude legs (annapolis)`.
- **What:** The regenerated table 14 lines above line 120 now shows Fable 5 at 0.611 and Fable 5.1 at 0.610, and `docs/model_comparison.md` says "Both displace `claude-opus-5` at the top of this split". The scoreboard — the summary page — says the opposite. The leg count is now eight (six Claude), not six (four). Note that "one split each" was already wrong on main (Opus-low has three rows: annapolis plus both Laurens arms); the PR adds the second contradiction.
- **Evidence:** `grep -n "Six legs\|strongest challenger" docs/model_scoreboard.md` -> 96, 120, 389. All three are outside the `<!-- BEGIN GENERATED -->` blocks, so `scoreboard.py` cannot fix them and no test covers this file's prose.
- **Fix:** rewrite the three sentences (eight legs, six Claude; Fable 5 is the strongest challenger on annapolis, with the serving-path caveat). While there: add `claude-fable-5-effort-low-anthropic` / `claude-fable-5-1-effort-low-anthropic` to `DISPLAY` in `scripts/analysis/scoreboard.py` (line 181) so the two rows render like their siblings instead of as raw slugs, and change `rampnet/roster.py` line 39 (`claude_serving_path (#156) will be the first pin`) to past tense now that it is.
- **Found by:** Opus 5

#### S3 · nit · doc numbers — cost column is the 120-pano row, and the `manual_gold` estimate does not follow from the rate it cites
- **Where:** `docs/model_comparison.md` Fable table (`$18.47` / `$19.86`) and the expansion sentence `at the measured $0.0272/call it is ~$155 per id for manual_gold alone`.
- **What:** (a) The table's cost is the 720-call full-leg ledger row, i.e. 120 panos; the other 5 per id were served from the calibration cache (commit `e3382ab` says so). Adding the first calibration pass gives ~$19.23 (5) / ~$20.67 (5.1), so the "2.1x / 2.2x the Opus leg" ratio is really 2.15x / 2.31x. (b) `manual_gold` is 1,000 panos × 6 views = 6,000 calls; at $0.0272/call that is $163, not $155. $155 reconciles with Fable 5's measured rate ($0.02566/call -> $154), not the $0.0272 the sentence quotes (which is 5.1's calibration rate).
- **Evidence:** `analysis_out/usage_log.jsonl` rows 2026-09-05 15:52 / 16:47: `calls: 720, panos_scored: 125`, per-call $0.02758 / $0.02566; calibration rows 14:21–14:23 total 54 + 48 calls, $2.68. `wc -l benchmark/manual_gold/records.jsonl` -> 1000; every signature has 6 views.
- **Fix:** label the column "cost, 120 uncached panos" (or fold the calibration in with a footnote), and write "~$154–165 per id" with the per-call rate the number came from.
- **Found by:** Opus 5

#### S4 · nit · convention — the probe script imports `anthropic` bare, unlike the detector it fronts
- **Where:** `scripts/model_comparison/probe_claude_models.py` — `from anthropic import AnthropicVertex` (and `from anthropic import Anthropic`, `import anthropic` in `probe`).
- **What:** Without the package the script dies with a raw `ModuleNotFoundError`. `ClaudeDetector._vertex_client` / `_anthropic_client` wrap the same import and name `requirements-vlm.txt`; the probe — the script the docs tell a new user to run first — does not. Downgraded from Gemini's "correctness": the traceback names the missing package, so nobody is misled, just under-helped.
- **Fix:** the same `try/except ImportError` with the `pip install -r requirements-vlm.txt` message. No test needed.

### Partly valid / unverified

#### P1 · minor · premise unmeasured — the forced-`tool_choice` guard refuses `claude-fable-5`, and nothing in the repo shows that id actually 400s
- **Where:** `scripts/model_comparison/detectors.py` line 944 — `return "fable" in (model_id or "").lower()` (`claude_forbids_forced_tools`); the constructor guard at line 1021; `tests/test_model_comparison.py::test_forced_tool_choice_is_refused_on_fable_before_the_run_starts[claude-fable-5]`.
- **Opus's claim:** only Fable 5.1 rejects `tool_choice` `tool`/`any`; Fable 5 accepts it, so the guard (and the test that pins it) is wrong for the id the doc recommends under the recall-first framing.
- **What I could verify:** the repo's own statement "Forced tool use is gone in that family" (commit `1213fbb`, the [#156](https://github.com/ProjectSidewalk/RampNet/issues/156) comment, the docstring, the ValueError text) has no recorded measurement behind it: `probe_claude_models.py` never sends `tool_choice`, every ledger row for both ids ran `auto`, and no usage-log row, doc, or issue comment records a forced-tool 400 on either id. The Anthropic API reference bundled with the Claude Code `claude-api` skill (cached 2026-06-24) lists the forced-`tool_choice` 400 as one of three breaking changes introduced in Fable 5.1 relative to Fable 5, and its per-model table names only 5.1 and Mythos 5.1 as rejecting it — so Opus quoted its source accurately. But that reference is not repo evidence, and it may itself be stale. Neither side has a live probe.
- **Failure scenario if Opus is right:** `compare.py ... --models claude:claude-fable-5 --claude-tool-choice forced` is refused before any call for a configuration the API accepts — and `forced` is the mode the detector's own comment calls "the better choice at effort=low". Published numbers are unaffected either way (both legs ran `auto`).
- **Fix:** settle it with a measurement, not a reference: one 16-token call with `tool_choice={"type":"tool", ...}` on each id (a fraction of a cent; add a `--tool-choice forced` option to `probe_claude_models.py` so the result is reproducible and record the status codes in [#156](https://github.com/ProjectSidewalk/RampNet/issues/156)). Then either narrow the guard to `claude-fable-5-1` and move `claude-fable-5` to `test_forced_tool_choice_still_works_on_the_models_that_published_with_it`, or keep it and cite the probe in the docstring. Independently: the substring `fable` does not cover `claude-mythos-5-1`, which the same reference says shares the restriction — harmless today (not in the roster), worth a comment.
- **Found by:** Opus 5

### Rejected or downgraded

- **Gemini 3.1 Pro: `location_warning` prints "WARNING: None is a REGIONAL Vertex endpoint" when `ClaudeDetector(location=None)`** — rejected. The constructor does `self.location = location or os.environ.get("GOOGLE_CLOUD_LOCATION") or "global"`, so `None` never reaches `location_warning`. Ran it: `ClaudeDetector(model_id='claude-fable-5', location=None).location -> 'global'`, `location_warning() -> None`; the warning fires only for a real region on the vertex path and is correctly suppressed on `anthropic`.
- **Gemini 3.1 Pro: probe script crashes without `anthropic`** — downgraded from correctness to nit (S4).
- **Opus 5: `claude-fable-5` accepts forced tool choice** — moved to partly valid / unverified (P1); the repo has no probe result either way.
- **Opus 5: cost column and $155** — kept as nit (S3), with the note that $155 does reconcile with Fable 5's rate, just not the rate the sentence cites.

### Coverage

- Gemini 3.1 Pro: exit 0, 2 findings, end marker verified, 425 s. Gemini 3.8 Flash: exit 0, 0 findings, end marker verified, 296 s. Files excluded from both Gemini reviews: none (19 files, 130,223 chars, 137,623 on the wire under the 180,000 guard). Flash's reasoning shows one drafted-then-dropped finding ("if `roster.leg_for` is unavailable the export will fail") — not a defect — and a fragment `"...may silently fail due to missing keys"` inside a section the transcript truncated, which cannot be recovered; its zero is therefore "found nothing it kept", not a clean bill.
- Tests run: full suite `PYTHONPATH=. pytest -q` in the worktree -> 1495 passed, 1 skipped (131 s), matching the PR body; the roster / published-legs / Fable / forced-tool subset re-run (61 passed). Reproduced the S1 fallback, the `location=None` path, the S3 arithmetic (ledger rows recompute to the cent from their own `pricing` blocks, cache tokens zero throughout), thinking/call and boxes/pano against the committed files, and the `Six legs` / `strongest challenger` grep.
- Not checked: the regenerated PNGs under `docs/figures/`; any live API call (no key spent, so P1 stays open); whether the vertex and first-party paths return identical detections for one id (the doc states this is untested, and it is); the disclosed calibration cache-write gap (42 of 60 re-called, in the ledger, nothing published depends on it).

---

jonfroehlich:

Commits `97c41aa..47fd930` on this branch, one per finding.

- S1 — fixed in [090f585](https://github.com/ProjectSidewalk/RampNet/commit/090f585916a897b075c55e079a71cc393eb65370): `publication_name` now raises when a spec is registered, every leg of it is pinned, and none matches the run's flags, naming the pin sets the registry knows and the values the run had. `--publish-as` still names a new leg; an unregistered spec still falls back to its plain label. The candidate lookup is factored out of `leg_for` as `roster.legs_of`. Tests: `test_a_registered_spec_whose_pins_all_miss_refuses_the_bare_label[claude:claude-fable-5 | claude:claude-fable-5-1]`, `test_the_same_hole_on_the_effort_pin_is_closed_too` (`tests/test_export_model_cache.py`).
- S2 — fixed in [1b89e66](https://github.com/ProjectSidewalk/RampNet/commit/1b89e669938c580cb99500142b3610524e071917): the three sentences in `docs/model_scoreboard.md` now say eight legs (six Claude; Opus-low has three splits) and name Fable 5 (0.611) as the strongest challenger on annapolis, with the serving-path caveat. The two Fable legs got `DISPLAY` names in `scoreboard.py`; doc, `scoreboard.json` and the by-split figure regenerated (`--check` passes). Roster docstring moved to past tense. Test: `test_the_prose_beside_the_partial_table_agrees_with_it` (`tests/test_scoreboard.py`) reads the leg count and the best annapolis challenger off the board and checks the prose against both; it fails on the pre-fix prose.
- S3 — fixed in [620f951](https://github.com/ProjectSidewalk/RampNet/commit/620f9517b96ded75aba7830f1acc451659c9b50d): the cost column is labelled "720 calls (120 of 125 panos)", a paragraph under the table gives the whole-split totals with the first calibration pass counted ($19.23 / $20.67) and notes the second pass belongs to neither number; the Opus ratio is now per call, 2.15x / 2.31x; the `manual_gold` estimate is ~$154 and ~$165 per id at each id's own full-leg rate (6,000 calls). No P/R/F1 changes. Test: `test_the_fable_cost_figures_are_what_the_ledger_says` (`tests/test_claude_published_legs.py`) recomputes every one of those figures from `analysis_out/usage_log.jsonl`.
- S4 — fixed in [cec219f](https://github.com/ProjectSidewalk/RampNet/commit/cec219f521d8d5dd9ea8f112b950868fc13dd08e): `probe_claude_models.py` imports `anthropic` through one guarded helper that raises the detector's "pip install -r requirements-vlm.txt" message. No test, as proposed; checked by hand with the package masked.
- P1 — addressed without a measurement in [47fd930](https://github.com/ProjectSidewalk/RampNet/commit/47fd930e9bcb8f39934014f1e8461eaedb843fb1): the guard matches exact ids instead of the substring `fable`. `CLAUDE_FORCED_TOOLS_REJECTED = {claude-fable-5-1, claude-mythos-5-1}` (Anthropic's API docs are the source, and the comment says so). `CLAUDE_FORCED_TOOLS_UNVERIFIED = {claude-fable-5}`: still refused, but the error now says the question is open and gives the command that settles it. `probe_claude_models.py --tool-choice forced` is that command (a one-field tool with `tool_choice` type `tool`). **No API call was made; the measurement is left to whoever holds the key.** A 200 moves `claude-fable-5` out of both sets, a 400 moves it to the rejected set; record the status code in [#156](https://github.com/ProjectSidewalk/RampNet/issues/156) either way. Behaviour for every published leg is unchanged. Tests: `test_forced_tool_choice_is_refused_before_the_run_starts[...]`, `test_the_forced_tool_guard_matches_exact_ids_not_a_family_substring`, `test_the_probe_can_send_the_forced_tool_choice_the_guard_is_waiting_on` (`tests/test_model_comparison.py`).

Full suite in the worktree, `PYTHONPATH=. pytest -q`: **1503 passed, 1 skipped** (106 s); 1495 passed, 1 skipped before these commits.

jonfroehlich:

0 major / 4 minor / 2 nit confirmed. Of the five round-1 items, four are properly fixed and pinned (S1, S3, S4, P1); the S2 fix is incomplete — the rewritten breakdown sentence miscounts, its test cannot tell Fable 5 from Fable 5.1, and the same stale "Opus is the strongest" claim survives in `docs/model_comparison.md`, which round 1 did not check. Everything else re-verifies: full suite in the worktree `PYTHONPATH=. pytest -q` → 1503 passed, 1 skipped; `scoreboard.py --check` current; every cost figure recomputes from the ledger; the S1 refusal and the S4 import guard reproduce by hand. The single most important fix is S3/S4 below: two sentences in `model_comparison.md` that state the pre-Fable result in the present tense, on the page where the Fable result is written up.

### Confirmed

#### S1 · minor · doc number — the S2 rewrite miscounts the one-split legs
- **Where:** `docs/model_scoreboard.md` line 96 — `Eight legs have not run the pooled splits — six of them one split each, and Opus 5 at low effort three`
- **What:** Six plus one is seven. Seven legs have one split each.
- **Evidence:** `analysis_out/scoreboard.json` `per_split`: Vistas curb cut [richmond], Vistas +curb [richmond], Fable 5, Fable 5.1, Opus high, Sonnet low, Sonnet high [annapolis each] = 7, plus Opus low [annapolis, laurens_gsv, laurens_mapillary] = 8. The generated table below the sentence and the bullet at lines 397–399 ("the two Vistas arms ... and the six Claude legs") both add up; only line 96 does not.
- **Fix:** "seven of them one split each". Extend `test_the_prose_beside_the_partial_table_agrees_with_it` to assert the one-split count from the board, not just the total — the total is the only thing it pins today.
- **Found by:** Opus 5

#### S2 · minor · test infrastructure — the S2 pin passes for either Fable id
- **Where:** `tests/test_scoreboard.py` line 518 — `assert best["display"].startswith(claim.group(1).split(" at ")[0])`
- **What:** The best-challenger check is a prefix match. `"Claude Fable 5.1 (low, anthropic)".startswith("Claude Fable 5")` is `True`, so if a re-score puts 5.1 on top (they are 0.001 F1 apart: 0.610932 vs 0.609929) the table changes, the prose keeps saying Fable 5, and the test stays green. Ran the test's own logic against the committed board: both Fable display names satisfy it.
- **Evidence:** it does fail on the pre-fix prose ("Claude Opus 5" is not a prefix of the Fable display), so the fix commit's claim holds as far as it goes; it just cannot catch the one flip that is plausible.
- **Fix:** compare the whole model name: `best["display"].split(" (")[0] == claim.group(1).split(" at ")[0]`.
- **Found by:** Opus 5

#### S3 · minor · doc contradicted by its own later section — `model_comparison.md` still calls Opus the strongest on annapolis
- **Where:** `docs/model_comparison.md` line 1623 — `**\`claude-opus-5\` at \`low\` is the strongest general model measured on annapolis**, at 0.588`; and lines 82–83 — `**Two more off-roster models, \`claude-sonnet-5\` and \`claude-opus-5\`, have been run on annapolis only** — four legs`.
- **What:** Both are present tense and both are false by line 1705 of the same file ("Both displace `claude-opus-5` at the top of this split"). The "What has been run where" paragraph exists to explain every "published, not in these tables" row in the matrix above it; that matrix now has two Fable rows (lines 62–63) the paragraph never mentions. `grep -in fable docs/model_comparison.md` finds nothing between line 63 and the Fable section. This is the identical drift round 1 caught in `model_scoreboard.md` (S2), in the other doc.
- **Fix:** line 1623 → "was the strongest ... until the Fable legs (below)"; lines 82–88 → one sentence naming the two Fable legs, annapolis, and the serving path. A test in the style of the S2 pin, reading the best annapolis leg off the published files and matching it against this claim, would stop a third recurrence.
- **Found by:** Opus 5

#### S4 · minor · instrument — the probe the guard points at does not send the detector's request
- **Where:** `scripts/model_comparison/probe_claude_models.py` lines 173–177 — `request = dict(model=model_id, max_tokens=max_tokens, messages=[...])` / `if tool_choice == "forced": request["tools"] = [PROBE_TOOL]; request["tool_choice"] = {"type": "tool", ...}`
- **What:** `ClaudeDetector._call` (`detectors.py` line 1274) always sends `output_config={"effort": self.effort}` next to `tool_choice`; the probe sends no `output_config`. The doc's own measurement (lines 1560–1566) is that forced tool choice "suppresses thinking entirely", and line 1720 says the Fable family cannot disable thinking — so effort/thinking is the one axis on which forcing is known to interact, and it is the axis the probe omits. A 200 from `--tool-choice forced` shows the bare shape is accepted; it does not show that `forced` + `output_config.effort`, the request the guard protects, is. The probe test's docstring says the flag "builds the request the detector would send", and it asserts `tools` and `tool_choice` only.
- **Evidence:** code paths above. Not measured — no API call was made this round — so this is a gap in the instrument, not a claim about the API.
- **Fix:** send `output_config={"effort": "low"}` from the probe (at least under `--tool-choice forced`), and extend `test_the_probe_can_send_the_forced_tool_choice_the_guard_is_waiting_on` to assert `seen["output_config"] == {"effort": "low"}`.
- **Found by:** Opus 5

#### S5 · nit · cost sentence not updated with the S3 fix
- **Where:** `docs/model_comparison.md` line 1756 — `costs ~$38:`
- **What:** $38 is the sum of the two 720-call rows ($18.47 + $19.86 = $38.33), i.e. 120 panos. The from-scratch command two lines below runs 125 panos / 750 calls per id, which the S3 paragraph 56 lines above now prices at $19.23 + $20.67 = $39.90.
- **Fix:** "~$40 (the whole-split totals above)".
- **Found by:** Opus 5

#### S6 · nit · test name claims a provenance no published leg has
- **Where:** `tests/test_model_comparison.py` line 1271 — `def test_forced_tool_choice_still_works_on_the_models_that_published_with_it():`
- **What:** No published Claude leg ran with forced tool choice: all six `benchmark/model_detections/claude-*__annapolis.json` signatures and all 12 Claude ledger rows carry `tool_choice: auto` (checked). The only forced measurement on record is the one-view Vertex check at doc lines 1560–1566. Introduced in this PR (`1213fbb`), touched again by the P1 fix.
- **Fix:** rename to `..._on_the_ids_it_was_measured_on` and cite the doc lines in the docstring.
- **Found by:** Opus 5

### Rejected or downgraded

- **Gemini 3.1 Pro: `from pricing import price_for` at `tests/test_model_comparison.py` line 1211 raises `ModuleNotFoundError` because the `sys.path.insert` lives inside a later test** — rejected. Line 15 of the same file inserts `scripts/model_comparison` at module import, and lines 39–40 already `import pricing` / `from pricing import estimate_cost, price_for` at module level; the function-local import at 1211 and the second `sys.path.insert` at 1284 are redundant, not broken. The test passes in the full run. (Gemini's own reasoning ends by concluding the path is set correctly, contradicting its filed finding.)
- **Gemini 3.8 Flash (drafted, not filed): a test uses `re.search` without importing `re`, a `NameError`** — rejected. `test_scoreboard.py` imports `re` at line 19; the other three touched test files import it function-locally before each use (`test_export_model_cache.py` 356, `test_model_comparison.py` 1511, `test_roster.py` 95 and 168). All pass.
- **Gemini 3.8 Flash (drafted, not filed): documented boxes/pano 2.72 for `claude-fable-5` disagrees with the derived 2.624** — rejected. 2.624 is (tp+fp)/125 = 328/125. The published file holds 340 boxes over 125 panos = 2.72 exactly; the 12 extra are detections inside an ignore radius, which `rampnet/detection_eval.py::score_pano` counts as neither TP nor FP (line 245). Same convention on every leg (Opus-low: 320 boxes, 178+133 = 311).
- **Gemini 3.8 Flash (drafted, not filed): `roster.leg_for` may raise `TypeError` on a leg whose `pins` is an empty tuple** — rejected. `if c.pins and all(...)` short-circuits on `()`, and the unpinned branch returns the leg; ran it for `rampnet` and `gemini:gemini-3.6-flash` with defaults and with `cargs=None`, both resolve.
- **Opus 5: all six findings** — confirmed as filed; none downgraded.

### Coverage

- Gemini 3.1 Pro: exit 0, 1 finding (rejected above), end marker verified, 238 s. Gemini 3.8 Flash: exit 0, 0 findings, end marker verified, 144 s. Files excluded from both Gemini reviews: none (21 files, 155,784 chars, 164,144 on the wire under the 180,000 guard). Flash's transcript is truncated in the middle (~31 KB dropped) and shows four candidates it examined and dropped, three of which are verified above; the fourth, "a potential incon[sistency]" in the compare arguments, is cut mid-word and cannot be recovered. Its zero is "nothing kept", not a clean bill.
- Round-1 fixes re-verified: S1 — `publication_name("claude:claude-fable-5", <defaults>)` raises naming `claude_effort=low, claude_serving_path=vertex` and the registered leg; with `claude_serving_path=anthropic` returns `claude-fable-5-effort-low-anthropic`; `--publish-as` short-circuits; `claude-sonnet-5` at `medium` raises; an unregistered spec still falls through to its label. S2 — `scoreboard.py --check --no-figures` reports doc and JSON current; the three sentences are rewritten (but see S1–S3 above). S3 — every figure recomputes from the six Fable ledger rows: first pass $0.76/$0.82, whole split $19.23/$20.67, $0.0257/$0.0276 per call, 2.15x/2.31x against $8.94/750, ~$154/~$165 for 6,000 calls, second pass 42 calls / $1.11; every row's `est_cost_usd` matches its own `pricing` block. S4 — with `anthropic` masked at import, `main(["--serving-path","anthropic","--no-control"])` raises the `pip install -r requirements-vlm.txt` message. P1 — exact-id sets, `claude-fable-5` still refused with the settle-it command in the error; all six published Claude files and all 12 ledger rows carry `tool_choice: auto`, so no published number depends on the guard.
- Tests run: full suite `PYTHONPATH=. pytest -q` in the worktree → 1503 passed, 1 skipped (114 s), matching the fixes comment. Merge base is `2916ba5` = current `origin/main`, so there is no merge of main to check.
- Not checked: the regenerated PNGs under `docs/figures/`; any live API call (none made, so P1 stays unmeasured and S4 is about the instrument); whether the vertex and first-party paths return identical detections for one id (still untested, as the doc says); `export_model_cache.py --canonicalize` (not re-run this round, since it writes).

**Merge readiness:** ready after the listed fixes — the code and the numbers hold; S1 and S3 are two sentences on the two summary pages that state the pre-Fable result as current, and S2 is the test that was supposed to stop exactly that.

---

jonfroehlich:

Commits `47fd930..dc4c5fb` on this branch.

- S1 — fixed in [9ca54c6](https://github.com/ProjectSidewalk/RampNet/commit/9ca54c68e6b9bca8710ac456badeafb16ae372e8): "seven of them one split each". Test: `test_the_prose_beside_the_partial_table_agrees_with_it` (`tests/test_scoreboard.py`) now checks the one-split count and the three-split leg against the board, not just the total.
- S2 — fixed in the same commit: the best-challenger pin compares the whole model name (`display.split(" (")[0] == claim`), not a prefix. Checked by hand that naming Fable 5.1 in the doc now fails it.
- S3 — fixed in [87ff1bd](https://github.com/ProjectSidewalk/RampNet/commit/87ff1bde17e14d37cd2601c1cbdb81d8c32f54e0): the coverage paragraph in `docs/model_comparison.md` now names all six off-roster Claude legs, their serving paths and why the Fable names carry `-anthropic`; the Vertex-section sentence reads "`claude-opus-5` at `low` was the strongest … until the two Fable legs (0.611 and 0.610, next section) displaced it in turn". Test: `test_the_strongest_annapolis_claim_tracks_the_published_legs` (`tests/test_claude_published_legs.py`) scores the published annapolis legs and checks the sentence's verb and model against the best of them; it fails on the doc as it was.
- S4 — fixed in [dc4c5fb](https://github.com/ProjectSidewalk/RampNet/commit/dc4c5fb5677517cab32f7c53cb61e3a2bd78a644): `probe_claude_models.py --tool-choice forced` now sends `output_config={"effort": "low"}` beside `tool_choice`, the shape `ClaudeDetector._call` sends; the reachability probe under `auto` is unchanged. Still no API call made. Test: `test_the_probe_can_send_the_forced_tool_choice_the_guard_is_waiting_on` asserts `output_config`, the full key set, and its absence under `auto`.
- S5 — fixed in [87ff1bd](https://github.com/ProjectSidewalk/RampNet/commit/87ff1bde17e14d37cd2601c1cbdb81d8c32f54e0): "costs ~$40 (the whole-split totals above)".
- S6 — fixed in [dc4c5fb](https://github.com/ProjectSidewalk/RampNet/commit/dc4c5fb5677517cab32f7c53cb61e3a2bd78a644): renamed to `test_forced_tool_choice_still_works_on_the_ids_it_was_measured_on`, docstring cites the one-view Vertex check in the doc; the `detectors.py` comment that names the tests updated to match.

Full suite in the worktree, `PYTHONPATH=. pytest -q`: **1504 passed, 1 skipped** (111 s); 1503 passed, 1 skipped before these commits. `scoreboard.py --check` current.

jonfroehlich:
## Merged main into this branch

[416640d](https://github.com/ProjectSidewalk/RampNet/commit/416640d01014ec87a8447c2ed2108a618edf323b) merges `origin/main` ([#146](https://github.com/ProjectSidewalk/RampNet/pull/146), [#147](https://github.com/ProjectSidewalk/RampNet/pull/147)) into this branch. A plain merge, no rebase. Five files conflicted, all docs and figures:

- `docs/model_comparison.md` — the off-roster Claude paragraph now says: **six Claude legs are off-roster** (four Vertex, two Fable on the first-party path); 16 published files, 14 verified against the cache that produced them (the two Laurens files are re-scored from the committed detections instead); `claude-opus-5` at `low` has run **eleven splits**, everything but `manual_gold`, with the measured $70.41 for eight splits / ~$8.80 per split from #139; the **other five legs are annapolis only**, and for the two Fable legs the expansion has not been taken (~$154–165 per id for `manual_gold` at their measured rates).
- `docs/model_scoreboard.md` — prose resolved by hand, tables/JSON/figures regenerated with `scoreboard.py` (`--check` current). With Opus-low complete, all seven partial legs are one-split, and the sentences say so; Fable 5 remains the strongest challenger on annapolis, beside main's note that Opus-low's annapolis lead over gemini-3.1-pro did not survive the pooled splits.
- `docs/replication.md` — `model_detections/` recomputed from the merged tree: 148 files, 25.3 MB, with a row for the two Fable files.
- the two scoreboard PNGs — regenerated.

Two tests adjusted to the merged board: the partial-table prose pin accepts an all-one-split board, and the `model_comparison.md` pin reads the merged sentence. Full suite: **1650 passed, 1 skipped**; CI green on 3.10 and 3.12.

PR #160, opened 2026-09-09
Show the invocation that actually runs: REPO and RAMPNET_ENV (#51, #135)

jonfroehlich:
Campaign B of the seed-variance campaign (#135, and the number #51 is blocked on) has now failed to train a single step **three times**, on three different one-line causes. It was dead again from 2026-09-06 until this morning, and nothing was queued in between.

| jobs | date | cause |
|---|---|---|
| 39515025/26/27 | 2026-09-03 | bash parse error in the `SEED` guard — fixed in 2b7471c |
| 39583887/90/91 | 2026-09-04 | `RAMPNET_ENV` unset → `source activate sidewalkcv2`, which a non-login sbatch environment cannot resolve |
| 39619547/48/49 | 2026-09-06 | `REPO` unset → `$HOME/RampNet`, a stale Aug-17 clone whose `train.py` predates `--seed`, so all 16 ranks died with `train.py: error: unrecognized arguments: --seed 1` |

The launcher's own USAGE header steered into two of those, because it showed the bare `SEED=1 sbatch stage_two/run_train_seed.slurm` form — which cannot work on klone as things stand. `docs/seed_variance_51_135.md`'s reproduce block showed the same form, plus `cd ~/RampNet`, which is precisely the stale clone. This was flagged on #135 at the time and not fixed.

Both now give the full invocation and say why each variable is load-bearing:

- **`REPO` also feeds `PYTHONPATH`**, so a wrong `REPO` hides `rampnet.seeding` as well as the wrong `train.py`. `$PWD` is the right value whenever you submit from the checkout you mean to run.
- **The argparse error is on line 1 of the `.err`**; the remaining ~30 KB is a torch `RendezvousConnectionError` traceback from the ranks tearing down. The header says to read the head, not the tail, because the tail is what you reach for first and it names the wrong thing.
- **The env path takes a literal `jonf`, not `$USER`.** `/gscratch/makelab` carries both a `jonf` and a `jfroehli` directory for the same person, and the env is under `jonf` while `$USER` is `jfroehli`. `$USER` is correct for `RUNDIR` and `DATA_ROOT` on `/gscratch/scrubbed`, which is what makes this easy to miss.
- **Verify with `sacct -D` about an hour after submit.** All three failures returned a job id and were indistinguishable from a healthy submission in `squeue`; two sat unnoticed for a day or more. Plain `sacct` additionally hides requeued incarnations.

Comments and documentation only — no behaviour change. `tests/test_seeding.py` passes (23 tests) and `bash -n stage_two/run_train_seed.slurm` is clean.

Campaign B was resubmitted with the corrected form this morning as **39880702 / 39880703 / 39880706**.


jonfroehlich:

0 major / 4 minor / 1 nit confirmed, from three reviews (Claude Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash) verified against the PR head (`84ec8f1`) and the merge-base (`2916ba5`). The PR does what it says: the USAGE header and the reproduce block in `docs/seed_variance_51_135.md` now show the invocation that actually ran Campaign B (`39880702/03/06`, all COMPLETED per [#135](https://github.com/ProjectSidewalk/RampNet/issues/135)), and each failure mode is named. The one thing to fix before merge: the file now documents that both defaults are wrong on the only cluster it targets, but leaves them in place, so the failure it describes is still one omitted variable away. Gemini 3.1 Pro's only finding (seeds overwrite each other) is rejected; the launcher derives `RUNDIR` from `SEED`.

### Confirmed

#### S1 · minor · correctness / maintainability — The defaults the header calls wrong are still the defaults
- **Where:** `stage_two/run_train_seed.slurm` line 103 — `REPO="${REPO:-$HOME/RampNet}"`; line 131 — `source activate sidewalkcv2`
- **What:** The new header (line 31) says "SET REPO AND RAMPNET_ENV EXPLICITLY -- THE DEFAULTS ARE BOTH WRONG ON KLONE", and lines 45–47 say the `else` branch at 129–133 cannot succeed on this cluster ("no *named* env … prefixes only"). Neither default was changed, and the guard the script already uses for `SEED` (line 101, `${SEED:?…}`) is not applied to `RAMPNET_ENV`.
- **Evidence:** Read lines 101–133 on the PR head. `bash -n` clean. Line 20 already requires submitting from the repo root, and Slurm sets `SLURM_SUBMIT_DIR` in the job environment (documented sbatch behaviour; not testable here), so the correct `REPO` is available without asking for it.
- **Fix:** `REPO="${REPO:-${SLURM_SUBMIT_DIR:-$PWD}}"`; `RAMPNET_ENV="${RAMPNET_ENV:?set RAMPNET_ENV to the conda prefix, e.g. /gscratch/makelab/jonf/envs/sidewalkcv2}"` and drop the `else` branch; a one-line preflight `grep -q -- '--seed' "$REPO/stage_two/train.py" || { echo "FATAL: $REPO/stage_two/train.py has no --seed (stale checkout?)" >&2; exit 1; }`. Pin it in `tests/test_seeding.py` next to `test_seed_launcher_requires_a_seed` (line 189) with an assertion that the launcher contains `RAMPNET_ENV:?`. If this PR is meant to stay comments-only, say so in the header; as written the file argues against itself.
- **Found by:** Opus 5; Gemini 3.8 Flash reached the same finding in its reasoning and then dropped it from its output.

#### S2 · minor · scientific record / replicability — The copy-out step uses `$USER` on the volume the PR says `$USER` is wrong for, and does not match where the artifacts went
- **Where:** `stage_two/run_train_seed.slurm` line 69 — `cp "$RUNDIR/best_model.pth" /gscratch/makelab/$USER/seedvar/rampnet_s${SEED}_best.pth`; `docs/seed_variance_51_135.md` lines 289–290 (same `cp`, `rampnet_s1`)
- **What:** Lines 49–52 of the same header say `$USER` (`jfroehli`) resolves to the wrong directory under `/gscratch/makelab`. The copy-out instruction then uses `$USER` under `/gscratch/makelab`. It resolves to `/gscratch/makelab/jfroehli/seedvar/`, while the as-run record in [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) (comment of 2026-09-14) is `/gscratch/makelab/jonf/seedvar/` with `rampnet_s{1,2,3}_best.pth`, `rampnet_s{1,2,3}_epoch_1_step_9378.pth`, the three `.out` logs and `SHA256SUMS`. The instruction has no `mkdir -p` (the `cp` fails on a fresh `seedvar/`), copies only `best_model.pth`, takes no hash, and line 64 ("best_model.pth -- the only artifact this campaign produces") is contradicted by what was kept.
- **Evidence:** Every other durable path in `docs/stage2_epoch_curve_84.md` (lines 168, 188) and `docs/stage2_cosine_rung_135.md` (lines 169, 172, 217, 265) is the literal `/gscratch/makelab/jonf/...`.
- **Fix:** Write the literal `jonf` path with the same caveat the env path already carries, add `mkdir -p`, list the files actually copied and the `sha256sum -c` step, and correct line 64. This block is now the as-run record of a finished experiment; it should say what ran. No test; doc change.
- **Found by:** Opus 5

#### S3 · minor · replicability — The reproduce block starts inside a checkout nothing creates, and the job never records its commit
- **Where:** `docs/seed_variance_51_135.md` line 275 — `cd /gscratch/scrubbed/$USER/RampNet_seedvar && mkdir -p logs`; `stage_two/run_train_seed.slurm` lines 145–154 (banner)
- **What:** No clone/rsync step precedes line 275 and the commit is unstated. The banner prints seed, dirs, nodes and restart count but not the repo commit, so the root cause of the 09-06 failure (a checkout at the wrong commit) is undetectable in a run that *succeeds*: a checkout stale in `rampnet/seeding.py` or the model head would train to completion and be scored in [#161](https://github.com/ProjectSidewalk/RampNet/pull/161) with nothing saying which code produced it. The checkout and its `logs/` are also on `/gscratch/scrubbed`, the volume the same block says purges at ~21 days.
- **Evidence:** `grep -n "rev-parse\|git " stage_two/*.slurm` → no matches on the PR head. Contrast `docs/stage2_epoch_curve_84.md` lines 138–141 (shows the rsync) and `docs/stage2_cosine_rung_135.md` lines 207–210 (records the sha256 of the `train.py` that ran).
- **Fix:** Add the sync line and the commit the three completed jobs ran at (or "commit not recorded" if the checkout is gone), and add to the banner `echo "Repo commit: $(git -C "$REPO" rev-parse --short HEAD 2>/dev/null || echo unknown)  dirty: $(git -C "$REPO" status --porcelain 2>/dev/null | wc -l)"`. Extend the `test_seed_launcher_*` group to assert the banner references `rev-parse`.
- **Found by:** Opus 5

#### S4 · minor · maintainability — Two env prefixes and two stories about the default in one file
- **Where:** `stage_two/run_train_seed.slurm` lines 112–120 (unchanged), e.g. line 116 — `# built at a PREFIX, /gscratch/scrubbed/$USER/envs/sidewalkcv2, which \`source activate`
- **What:** The new header (lines 27, 45–52) says use `/gscratch/makelab/jonf/envs/sidewalkcv2` and that the default cannot work; the older comment 80 lines down says "Default behaviour is unchanged" and points at `/gscratch/scrubbed/$USER/envs/sidewalkcv2`. A reader who reaches line 116 first submits with the scrubbed prefix, pends for an unbounded time on `ckpt-all`, then fails the line-124 `-x bin/torchrun` check at job start — the exact shape this PR exists to stop.
- **Evidence:** `docs/stage2_epoch_curve_84.md` line 256: the scrubbed prefix was built by job `38541308` on 2026-08-17 and is past the ~21-day idle window unless touched; `docs/stage2_cosine_rung_135.md` line 217: the durable copy under `makelab/jonf` (job `38566410`) is what the last two runs used.
- **Fix:** Collapse lines 112–120 to a pointer at USAGE and drop "Default behaviour is unchanged". Folds naturally into S1.
- **Found by:** Opus 5

#### S5 · nit · doc accuracy — The parse error fires at job start, not at `sbatch` time
- **Where:** `stage_two/run_train_seed.slurm` line 55 — `#     39515025/26/27 on 2026-09-03 at submit time, in about one second, printing`; also pre-existing lines 98–99, "dies at submit time"
- **What:** `sbatch` only scans `#SBATCH` directives; bash parses the script when the job starts on a node. The header itself says three lines later that these jobs "returned a job id and looked indistinguishable from a healthy submission in `squeue`", which cannot be true of a submit-time failure.
- **Evidence:** Commit `2b7471c`: the stderr was `slurm_script: line 46: unexpected EOF while looking for matching \`}'` — `slurm_script` is the name Slurm gives the batch script when it executes it on the node — and the jobs were "submitted … at 15:01 and were all dead one second later, FAILED with exit 2". The one second is job runtime with capacity free, not `sbatch` latency.
- **Fix:** "at job start" in both places.
- **Found by:** Opus 5

### Rejected or downgraded
- **Gemini 3.1 Pro: the submission loop omits the per-seed `RUNDIR` override, so all three seeds write to the same directory and overwrite each other's `best_model.pth`** (reported twice, once per file) — rejected. `RUNDIR` is derived from `SEED` at line 107: `RUNDIR="${RUNDIR:-/gscratch/scrubbed/$USER/seedvar/rampnet_s${SEED}}"`. Evaluated for `SEED=1,2,3` → `.../seedvar/rampnet_s1`, `_s2`, `_s3`. The deleted `SEED=2 RUNDIR=... sbatch` line in the old header set `RUNDIR` to exactly its default, so removing it changes nothing. `test_seed_launcher_isolates_each_replicate_in_its_own_directory` (line 196) pins this and passes. The reviewer's reasoning shows it never saw line 107 (it reviewed the patch only).
- **Gemini 3.8 Flash: no findings** — its reasoning found S1 ("contains the comment 'THE DEFAULTS ARE BOTH WRONG ON KLONE' but doesn't fix the defaults") and then emitted an empty list; credited under S1. It also mentioned "a trailing space" in the loop: rejected, neither the doc block (lines 264–292) nor the launcher header has trailing whitespace.

### Coverage
- Files excluded from the Gemini reviews: none (both logs: `excluded: (none)`, `untracked: 0 included`, `tree: clean`). Both runs exited 0 with the end marker verified; Gemini 3.1 Pro 218 s / 85,750 tokens, Gemini 3.8 Flash 166 s / 78,931 tokens.
- Tests run: `PYTHONPATH=<worktree> pytest -q tests/test_seeding.py tests/test_slurm_scripts.py` → 49 passed; `bash -n stage_two/run_train_seed.slurm` clean; `git merge-tree --write-tree HEAD origin/exp/seed-variance-read-51-135` clean, so this PR and [#161](https://github.com/ProjectSidewalk/RampNet/pull/161) do not conflict on the shared doc. Opus's full-suite run (1481 passed, 1 skipped) was not repeated.
- Not checked: anything on klone (no Duo) — the current state of `~/RampNet`, either env prefix, whether `/gscratch/makelab/jfroehli/` exists, and the "prefixes only, no named env" claim. S2's as-run path is taken from the [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) comment, not from the filesystem. `SLURM_SUBMIT_DIR` in S1 is documented Slurm behaviour, not tested here.

---

jonfroehlich:

Four commits on this branch, `84ec8f1..fcc4264`.

- S1 — fixed in 52ea7fb: `RAMPNET_ENV` is now required (`${RAMPNET_ENV:?...}`, the same guard as `SEED`) and the `source activate` fallback is gone; `REPO` defaults to `SLURM_SUBMIT_DIR` (then `$PWD`) instead of `$HOME/RampNet`, and a preflight aborts at job start if `$REPO/stage_two/train.py` does not take `--seed`. The doc's reproduce block says the same. Tests: `test_seed_launcher_requires_a_conda_prefix`, `test_seed_launcher_does_not_default_repo_to_home`, `test_seed_launcher_refuses_a_checkout_without_seed_support` (source-level), plus `test_the_env_guard_rejects_an_unset_prefix`, `test_repo_defaults_to_the_submit_directory`, `test_the_checkout_preflight_rejects_a_train_py_without_seed[current|stale]` (evaluated in bash). All seven fail on `84ec8f1`. `SLURM_SUBMIT_DIR` is set by hand in the test; Slurm exporting it is documented behaviour, not tested here.
- S2 — fixed in 54a5b27: the copy-out now targets the literal `/gscratch/makelab/jonf/seedvar/`, creates it, copies `best_model.pth`, `checkpoints/epoch_1_step_9378.pth` and the job log per seed, `cmp`s against the originals and writes and checks `SHA256SUMS`; that is the layout recorded in [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) on 2026-09-14. The header no longer calls `best_model.pth` the only artifact. Doc change; no test.
- S3 — fixed in a57e549: the banner prints the checkout path, its short commit and its dirty-file count (`unknown` when the directory is not a git repo, so a `.git`-less rsync still runs); the reproduce block gains the rsync step. The commit `39880702/03/06` ran at was not recorded and the doc says so; what the run pins down is stated beside it (at or after `455e1c4`, at or before `2916ba5`, and the one code difference inside that range trains the same recipe at its defaults with zero preemptions). Tests: `test_seed_launcher_logs_the_commit_it_ran`, `test_the_commit_line_survives_a_checkout_that_is_not_a_repo`.
- S4 — fixed in 52ea7fb, with S1: the older interpreter comment ("Default behaviour is unchanged", scrubbed prefix) is collapsed to the code it describes; the header is the one account of the defaults.
- S5 — fixed in fcc4264: "at job start" in the header, the `SEED` guard comment, the Amendment 1 paragraph of the doc, and the `tests/test_slurm_scripts.py` docstring, which repeated the same claim. Wording only; no test.

Not changed: the `REPO=$PWD` in the USAGE loop and the doc, now redundant, is kept because it is what the completed replicates were submitted with. Nothing on klone was checked (no Duo); S2's paths come from the [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) record.

Full suite: `1490 passed, 1 skipped` (was 1481 / 1 on `main`); `bash -n stage_two/run_train_seed.slurm` clean.

jonfroehlich:

0 major / 1 minor / 3 nit confirmed, from three reviews (Claude Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash) verified against the PR head (`fcc4264`, merge-base `2916ba5`). **All five round-1 items (S1–S5) are confirmed properly fixed**: the launcher now requires `RAMPNET_ENV`, defaults `REPO` to `SLURM_SUBMIT_DIR`, refuses a `train.py` without `--seed`, prints the checkout commit, and the doc's copy-out matches the [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) 2026-09-14 record; the seven new tests all fail against the `84ec8f1` launcher and pass on HEAD (details under Coverage). The one thing to fix: the new rsync line in the reproduce block expands `$USER` on the local machine, so the checkout lands somewhere other than where the block's own `cd` goes on klone — the same `$USER`-points-elsewhere defect this PR exists to document.

### Confirmed

#### S1 · minor · replicability — The rsync target's `$USER` is the local user, not the klone user
- **Where:** `docs/seed_variance_51_135.md` line 271 — `--exclude runs --exclude '*.pt' RampNet/ klone:/gscratch/scrubbed/$USER/RampNet_seedvar/`
- **What:** `rsync` runs locally; only the `klone:` prefix is remote, and the path after it is expanded by the local shell before rsync sees it. Line 295 (`cd /gscratch/scrubbed/$USER/RampNet_seedvar`) runs on klone, where `$USER` is `jfroehli` (launcher line 56). Anyone whose local username is not `jfroehli` syncs the tree to one directory and then `cd`s into another.
- **Evidence:** Local `$USER` measured on this desktop: WSL Ubuntu `jonf`; Git Bash **empty**, so from there the target becomes `klone:/gscratch/scrubbed//RampNet_seedvar/`, i.e. `/gscratch/scrubbed/RampNet_seedvar/`. Neither is where line 295 goes. The two docs this block descends from avoid the trap by targeting `klone:~/RampNet/` (`docs/stage2_epoch_curve_84.md` 139–141, `docs/stage2_cosine_rung_135.md` 163–165), where `~` is left to the remote. The PR moved the target to scratch, which is right (home is the stale clone), and picked up local expansion in the move. Whether the wrong target fails loudly or succeeds silently depends on whether that directory exists on klone, which was not checked (no Duo); either way the `cd` on line 295 then finds nothing or finds a stale `RampNet_seedvar`, and the preflight at launcher 121 catches that only if the stale `train.py` predates `--seed`.
- **Fix:** Write the klone username literally, `klone:/gscratch/scrubbed/jfroehli/RampNet_seedvar/`, with a one-line note that `$USER` on that line would be the *local* one (single-quoting does not help; rsync does not expand remote paths). Doc change, no test. If how the 09-09 checkout was actually synced is known, say so beside it, since the block claims "it is what the completed replicates used" and the [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) 09-09 comment records only `REPO=$PWD`.
- **Found by:** Opus 5

#### S2 · nit · test robustness — The preflight test pastes an unquoted temp path into a bash script
- **Where:** `tests/test_slurm_scripts.py` line 138 — `[BASH, "-c", "REPO=" + tmp_path.as_posix() + "\n" + m.group(0)],`
- **What:** If `tmp_path` ever contains a space or a shell metacharacter, bash parses `REPO=/a b` as a prefix assignment to the command `b`; `REPO` is then unset for the preflight and the `current` case fails spuriously.
- **Evidence:** Reproduced with `tmp_path=/tmp/p 160`: unquoted → `bash: line 1: 160: command not found`, then `FATAL: /stage_two/train.py is missing ...`, exit 1; quoted → exit 0. pytest's `tmp_path` on Linux CI (`/tmp/pytest-of-runner/pytest-N/<sanitized name>0`) has no such character, which is why this cannot fire on the workflow as configured; on Windows it would for a profile path containing a space. Cheap and certain, so a nit, not a defect in the launcher.
- **Fix:** `"REPO=" + shlex.quote(tmp_path.as_posix())`. The existing parametrized test covers it once the path is quoted.

#### S3 · nit · doc accuracy — Header and doc disagree on where the archived job logs live
- **Where:** `stage_two/run_train_seed.slurm` line 77 — `rampnet_s<SEED>_epoch_1_step_9378.pth, logs/seedvar_<jobid>.out and a SHA256SUMS`
- **What:** The header puts the logs under `seedvar/logs/`; the doc's copy-out (lines 321–323) copies them to the top of `$DST` and hashes `seedvar_*.out` there. The [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) 09-14 record says only "the three `.out` logs", so which layout is on disk cannot be told from the record.
- **Evidence:** Read both on HEAD.
- **Fix:** Drop `logs/` from line 77, or add it to the doc's `cp` and `sha256sum` lines — whichever matches `/gscratch/makelab/jonf/seedvar/`.
- **Found by:** Opus 5

#### S4 · nit · log accuracy — `dirty files: 0` when the commit is `unknown`
- **Where:** `stage_two/run_train_seed.slurm` line 165 — `echo "Repo commit: $(git -C "$REPO" rev-parse --short HEAD 2>/dev/null || echo unknown)   dirty files: $(git -C "$REPO" status --porcelain 2>/dev/null | wc -l)"`
- **What:** With no `.git` (or no `git` on the node) the line prints `Repo commit: unknown   dirty files: 0`, which reads as a clean checkout when nothing was measured.
- **Evidence:** `bash -c 'set -euo pipefail; REPO=/nonexistent; <line 165>'` → `Repo commit: unknown   dirty files: 0`, exit 0. `wc -l` on an empty pipeline is `0` regardless of why. `test_the_commit_line_survives_a_checkout_that_is_not_a_repo` asserts only the `Repo commit: unknown` prefix.
- **Fix:** Compute one status: `if commit=$(git -C "$REPO" rev-parse --short HEAD 2>/dev/null); then dirty=$(git -C "$REPO" status --porcelain | wc -l); else commit=unknown; dirty=unknown; fi`, echo both, and extend that test to assert `dirty files: unknown`.
- **Found by:** Opus 5

### Rejected or downgraded
- **Gemini 3.1 Pro: "vulnerable to word splitting and evaluation" / "command injection" in the test** — partly valid, kept as S2 at nit. The mechanism is real (reproduced above), but the input is pytest's own `tmp_path`, not attacker-controlled, and it cannot contain a space on the CI runner. Its reasoning also says it "verified the fix by adding quotes"; no such change exists in the diff.
- **Gemini 3.8 Flash: no findings.** Its reasoning ends on "the script creates the logs directory inside the repo, which causes a dirty file" and then emits an empty list. Rejected: root `.gitignore` ignores `*.out`/`*.err`, and `git status --porcelain` lists neither an empty directory nor one holding only ignored files (tested in a scratch repo: empty `logs/` → 0, `logs/` with only `.out`/`.err` → 0, with one non-ignored file → 1). The new `dirty files` count is not inflated by the launcher's own `mkdir -p "$REPO/logs"` or by Slurm's log files. Flash also noted the rsync does not exclude `.git`; that is correct and required, since the commit banner needs it.
- **Opus 5, considered and not reported (agreed):** the `RAMPNET_ENV` guard (139) runs after `mkdir -p` (132), so an unset env leaves an empty rundir — harmless; the preflight `grep` would match `--seed` in a comment — against the stale `train.py` that actually exists (`3a95305`, 0 matches) it is not vacuous; `rampnet/roster.py` also changed between `455e1c4` and `538a3bb` — not on the training path, the doc's "same recipe" conclusion stands; the [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) 09-14 comment says "the 09-10 resubmission" where the doc and the 09-09 comment say 09-09 — the doc is right, issue-side slip.

### Coverage
- Files excluded from the Gemini reviews: none (both logs: `excluded: (none)`, `untracked: 0 included`, `tree: clean`, 4 files / 408 lines). Both exited 0 with the end marker verified; Gemini 3.1 Pro 361 s / 76,459 tokens / 1 finding, Gemini 3.8 Flash 191 s / 85,238 tokens / 0 findings.
- Round-1 fixes verified: S1 — launcher 116, 121–125, 139 read; no `source activate` on a code line; the 7 new tests (9 ids, parametrized) copied into a scratch tree with the `84ec8f1` launcher all fail, all pass on HEAD; the Aug-17 `train.py` (`3a95305`) has 0 `--seed`, HEAD has 4. S2 — doc 313–323 literal `/gscratch/makelab/jonf/seedvar`, `mkdir -p`, three files per seed, `cmp`, `SHA256SUMS -c`; header 69–82 no longer calls `best_model.pth` the only artifact (S3 above is the one residual). S3 — banner line 165 present and pinned; the doc's commit-range argument checks out: `455e1c4` (2026-09-04) is the first commit with `RAMPNET_ENV` in the launcher, `git diff --stat 538a3bb 2916ba5 -- rampnet stage_two` is empty, `455e1c4..538a3bb` touches `stage_two/train.py` only. S4 — the old "Default behaviour is unchanged" block is gone; remaining `$HOME/RampNet` / `source activate` mentions are historical (lines 36, 49, 115, 136; doc 288–290). S5 — "job start" at launcher 53, 60, 108, doc 131, test docstring 6; no "submit time" left in the four files.
- Tests run: `PYTHONPATH=<worktree> pytest -q tests/test_seeding.py tests/test_slurm_scripts.py` → 58 passed; `bash -n stage_two/run_train_seed.slurm` clean; worktree clean at `fcc4264`. Opus's full-suite run (1490 passed, 1 skipped) was not repeated.
- Not checked: anything on klone (no Duo) — whether `/gscratch/scrubbed/jonf` or `/gscratch/scrubbed/RampNet_seedvar` exists (decides whether S1 fails loudly or silently), the on-disk layout under `/gscratch/makelab/jonf/seedvar/` (S3), how the 09-09 checkout was synced, and Slurm's `SLURM_SUBMIT_DIR` export (documented, set by hand in the test).

**Merge readiness:** ready after the listed fixes — S1 is a one-line doc edit and the only item that changes what a reader following the block would do; S2–S4 are cheap and can ride the same commit.

---

jonfroehlich:

Two commits on this branch, `fcc4264..0929070`.

- S1 — fixed in 1611042: the rsync target is the literal `klone:/gscratch/scrubbed/jfroehli/RampNet_seedvar/`, with a note that the path after `klone:` is expanded on the local machine (`$USER` there is `jonf` in WSL and empty in Git Bash). The block now says that directory is where the completed replicates were submitted from on 2026-09-09 and that how it was synced that day was not recorded, instead of claiming the rsync is what ran. Doc change; no test.
- S2 — fixed in 0929070: `tmp_path` is `shlex.quote`d before being pasted into the preflight test's bash script.
- S3 — fixed in 0929070: the header now lists the archived logs as `seedvar_<jobid>.out` beside the checkpoints, matching the doc's copy-out and the [#135](https://github.com/ProjectSidewalk/RampNet/issues/135) record.
- S4 — fixed in 0929070: one `if commit=$(git rev-parse ...)` sets both fields, so a checkout without `.git` prints `Repo commit: unknown   dirty files: unknown`. Test: `test_the_commit_line_survives_a_checkout_that_is_not_a_repo` now evaluates the whole block and pins both fields; it fails on `fcc4264`.

Full suite: `1490 passed, 1 skipped`; `bash -n stage_two/run_train_seed.slurm` clean. Nothing on klone was checked (no Duo).

